PR #294, opened 2024-09-29
[To split] Refactor BaseFold and add some features.

yczhangsjtu:
Massively refactored the code. Eliminate the repetitive codes.

Add features:
- [X] Add new API that supports opening multiple groups of polynomials at one point.
- [x] Configurable hash function
- [x] Update the query complexity according to the new paper
- [ ] ~~Skipping rounds optimization (consider canceling this because it's complex to implement, but the benefit is small since proving time is only 200ms now.)~~

API change:
- (Breaking change) To differentiate the three versions of batching protocols, rename the original `batch_open/verify` into `batch_vlmp_*` where `vlmp` stands for "polys of Variable Length opened at Multiple Points".
- The new API is named `batch_vlop_open/verify` where `vlop` stands for "polys of Variable Length opened at One Points".
- (Breaking change) Remove `Rng` from the template parameters for `Basefold`.


yczhangsjtu:
Too big PR. Splitting gradually (but not top priority for now), close first.

PR #357, opened 2024-10-11
[WIP] auto generate public values structure and index

hero78119:
This PR resolve possible tech debt in the future which defines public value index against the struct in different place, which  might lead to potential mismatch.

A getter/setter/index are all generate via rust macro

### DIFF
--- a/ceno_zkvm/examples/riscv_opcodes.rs
+++ b/ceno_zkvm/examples/riscv_opcodes.rs
@@ -2,7 +2,7 @@ use std::{iter, panic, time::Instant};
 
 use ceno_zkvm::{
     instructions::riscv::{arith::AddInstruction, branch::BltuInstruction},
-    scheme::prover::ZKVMProver,
+    scheme::{prover::ZKVMProver, PublicValues},
     tables::ProgramTableCircuit,
 };
 use clap::Parser;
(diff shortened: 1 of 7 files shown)


lispc:
i feel this is outdated and can be closed?

lispc:
> repr(C)

i usually do not do this in our use cases.   This is like mixing hardware encoding and business encoding. 

repr(C) like idea relates to programming language, compiler, cpu, os. (for example, i remember C or C++ support non-8-bit-byte?).  Fully writing your encoding inside codes means fully portable readable stable.

hero78119:
I will make it as draft first, for this PR still matter to address possible future bug, just not very urgent :)

matthiasgoergens:
> > repr(C)
> 
> i usually do not do this in our use cases.   This is like mixing hardware encoding and business encoding.  It is like we don't use protobuf for eth protocol wire encoding.
> 
> repr(C) like idea relates to programming language, compiler, cpu, os. (for example, i remember C or C++ support non-8-bit-byte?).  Fully writing your encoding inside codes explicitly means fully portable readable stable. 

In principle you would be right, but in Rust `repr(C)` is used a bit more generically than just for C compatibility.

Issue #359, opened 2024-10-12
Dummy Instruction Circuit


naure:
Create placeholder circuits that can handle any instructions.

The approach is to write one Dummy circuit for each instruction type. These produce the correct side-effects of each type: PC, register ops, memory ops. The circuit structures for each type already exist as gadgets (r_insn.rs, ...). The values are all available in the trace already so no calculation is needed. The dummy circuits just don't check the value correctness.

PR #369, opened 2024-10-12
Dummy Circuit - Basic ECALL

naure:
_Issue #359 and #567_

* The `DummyInstruction` implements all the communications of a step: state, fetch, registers, memory. But it does not verify calculations: any value can be written out.

* Placeholder circuits for missing implementations, including unknown ecalls.

* More precise register op assignment, see #570.

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
+++ b/ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
@@ -67,13 +67,16 @@ impl<E: ExtensionField> DummyConfig<E> {
         circuit_builder: &mut CircuitBuilder<E>,
         codes: InsnCodes,
     ) -> Result<Self, ZKVMError> {
-        let (with_rs1, with_rs2, with_rd) = match codes.format {
-            InsnFormat::R => (true, true, true),
-            InsnFormat::I => (true, false, true),
-            InsnFormat::S => (true, true, false),
-            InsnFormat::B => (true, true, false),
-            InsnFormat::U => (false, false, true),
-            InsnFormat::J => (false, false, true),
+        let (with_rs1, with_rs2, with_rd) = match (codes.format, codes.kind) {
+            // ECALL reads its syscall_id, then do nothing.
+            (_, InsnKind::EANY) => (true, false, false),
+            // Regular instructions do what is implied by their format.
+            (InsnFormat::R, _) => (true, true, true),
+            (InsnFormat::I, _) => (true, false, true),
+            (InsnFormat::S, _) => (true, true, false),
+            (InsnFormat::B, _) => (true, true, false),
+            (InsnFormat::U, _) => (false, false, true),
+            (InsnFormat::J, _) => (false, false, true),
         };
         let with_mem_write = matches!(codes.category, InsnCategory::Store);
         let with_mem_read = matches!(codes.category, InsnCategory::Load);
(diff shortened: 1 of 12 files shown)


naure:
Updated + now with ecall support.

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
```diff
@@ -0,0 +1,238 @@
...
+    imm: WitIn,
+}
+
+impl<E: ExtensionField> DummyConfig<E> {
+    fn construct_circuit(
+        circuit_builder: &mut CircuitBuilder<E>,
+        codes: InsnCodes,
+    ) -> Result<Self, ZKVMError> {
+        let (with_rs1, with_rs2, with_rd) = match (codes.format, codes.kind) {
+            (_, InsnKind::EANY) => (true, true, false),
```
kunxian-xia, line 71: For `ecall` instructions, I think they need to read the `rd` register to get `syscall_id`.

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
```diff
@@ -0,0 +1,238 @@
...
+            (InsnFormat::R, _) => (true, true, true),
+            (InsnFormat::I, _) => (true, false, true),
+            (InsnFormat::S, _) => (true, true, false),
+            (InsnFormat::B, _) => (true, true, false),
+            (InsnFormat::U, _) => (false, false, true),
+            (InsnFormat::J, _) => (false, false, true),
```
kunxian-xia, line 77: Maybe we can get rid of them as opcodes that are not supported (i.e. `DIV / REM / REMU`) right now belong to the `R` category.

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
naure: It has to match `VMState::ecall`. That’s not true in this PR, I fixed it in another branch. I can backport it here to be consistent.

naure: Better idea: just treat it as an I instruction. I’ll make it so in the guest-example. Done.

hero78119:
DummyCircuit implementation LGTM!
Do you think we better make a feature toggle e.g. "dummy" and make it as default feature now?
Via a feature toggling means, at some point, we can assure there is no dummy opcode by toggle off, so in case there are some leak but we dont find it.

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
naure: After reworking this a bit: the ignored ecalls just read the syscall_id and do nothing else. Added comments to explain.

naure:
@hero78119 This is a gadget that an example or a benchmark can use. If you don’t use it, it’s not there.

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
naure: I also turned it into a parameter: `Platform.unsafe_ecall_nop`

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
naure: Yes we have implemented almost everything in the mean time 😄 . Maybe there will be other use-cases: test, benchmarks, demos ?

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
```diff
@@ -0,0 +1,236 @@
...
+        };
+        let with_mem_write = matches!(codes.category, InsnCategory::Store);
+        let with_mem_read = matches!(codes.category, InsnCategory::Load);
+        let branching = matches!(codes.category, InsnCategory::Branch)
```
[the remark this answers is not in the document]
naure: > let the caller decide

Done.

> several memory words

(discussed offline) This is the next step to do in follow-up PRs.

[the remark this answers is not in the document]
naure: Tracked in #567.

kunxian-xia (approved):
Looks good to me! Let's merge it to unblock the integration test #368. Any remaining issue can be addressed in follow-up PRs.

PR #435, opened 2024-10-21
Simplify debug naming of variables in constraints

matthiasgoergens:
Instead of passing an `FnOnce` closure with no arguments, we just pass the name itself.

It looks like we copied the old convention from other R1CS proving systems like [bellman](https://github.com/zkcrypto/bellman/blob/3a1c43b01a89d426842df39b432de979917951e6/src/lib.rs#L374).  But we always execute our closure, rendering the whole exercise pointless, while they only do execute theirs in debug contexts.  (Thanks to Ming for the historic context!)

(And even for bellman, it's unlikely they actually get anything out of the extra complexity: making and passing closures around is a lot of work at runtime.  And an optimising compiler that can remove an unused closure can also much easier remove an unused string.)

---

I ran some benchmarks.  The change ever so slightly speeds up compilation and runtime, and decreases the size of our binaries.

On `master`:
```console
$ du -h target/*/libceno_zkvm.rlib
4.8M    target/debug/libceno_zkvm.rlib
4.4M    target/release/libceno_zkvm.rlib
```

On this branch:
```console
$ du -h target/*/libceno_zkvm.rlib
4.7M    target/debug/libceno_zkvm.rlib
4.3M    target/release/libceno_zkvm.rlib
```

This benchmark is just for fun, the PR is for readability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -241,6 +241,7 @@ dependencies = [
  "anyhow",
  "ceno-examples",
  "elf",
+ "itertools 0.13.0",
  "num-derive",
  "num-traits",
  "strum",
--- a/ceno_emul/Cargo.toml
+++ b/ceno_emul/Cargo.toml
@@ -7,6 +7,7 @@ version.workspace = true
 [dependencies]
 anyhow = { version = "1.0", default-features = false }
 elf = "0.7"
+itertools.workspace = true
 num-derive.workspace = true
 num-traits.workspace = true
 strum.workspace = true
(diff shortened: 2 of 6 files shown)


matthiasgoergens:
@lispc If you want to quickly get PRs into master, here's one that's been sitting for a while.  It simplifies our code, without any compromise in performance.

hero78119:
> @lispc If you want to quickly get PRs into master, here's one that's been sitting for a while. It simplifies our code, without any compromise in performance.

I dont think this fit the long sitting for a while PR, especially from my comment above. The closure style didn't bring up readibility issue to me, while with that in key setup/release build, we can eliminate those debugging message.

matthiasgoergens:
Yes, you can get rid of these messages in either style. And the optimiser has a slightly easier job with the plain string than when unnecessarily wrapping it in a closure.

PR #454, opened 2024-10-23
(WIP) Prototype for clone-free `Expression`

matthiasgoergens:
This should help with readability.  We want to cut down on visual noise, so that it'll be easier for us and any auditor (and enthusiasts, once this is open sourced) to spot problems with the constraints.  Or otherwise convince themselves that there are no problem with the constraints.

In theory, the new approach can be ever so slightly faster, but I don't think it makes much of a difference performance-wise.  The technique is borrowed from what the Rust compiler uses internally.


matthiasgoergens:
Closing this as obsolete for now.  Might resurrect, if there's demand.

Issue #378, opened 2024-10-14
investigate efficiency of memory finalized circuit


hero78119:
Per discussion here https://github.com/scroll-tech/ceno/pull/251#discussion_r1797717452

Currently in memory consistency check it's relied on `max_memory` as circuit constant parameter, which obviously bring up a possible `gap` between real in-used memory since max_memory need to cover the `worst case` memory consumption.

One solution, as non-uniform circuit and sumcheck based design, it's naturally possible treating memory table as special case of dynamic table, such that "memory size" can be auxiliary input from prover so the size can be determined for different proof and prover just need to "pay-as-you-go". 
Extra cost of this design is we need to prove `address` column are well-formed: uniquely and sorted

hero78119:
A simple non-uniform memory design mentioned here https://github.com/scroll-tech/ceno/issues/378#issuecomment-2436580306 has been addressed in #457 . The left issue is heap/stack address space is hardcode. We can raise another new issue to track this.

PR #457, opened 2024-10-24
implement (non-)volatile table modular circuits and e2e public io

hero78119:
### Working items & Scope
- [x] add module table circuit, e.g. `DynVolatileRamTable` for memory, `NonVolatileRamCircuit` for Register/Public IO/ProgramDataCircuit.
- [x] separate program data into dedicate circuit, and keep memory volatile.
- [x] prover/verifier protocol to evaluate public io
- [x] evaluate `DynVolatileRamTable` table address on verifier succinctly
- [x] register public_io and test e2e

succinctly address vector argument refer here: https://github.com/scroll-tech/ceno/issues/378#issuecomment-2436580306

Public IO Auxiliary will cover in next PR

### DIFF
--- a/ceno_zkvm/examples/riscv_opcodes.rs
+++ b/ceno_zkvm/examples/riscv_opcodes.rs
@@ -145,7 +145,10 @@ fn main() {
     let verifier = ZKVMVerifier::new(vk);
 
     for instance_num_vars in args.start..args.end {
-        let step_loop = 1 << (instance_num_vars - 1); // 1 step in loop contribute to 2 add instance
+        // The performance benchmark is hook on number of "add" opcode instances.
+        // Each iteration in the loop contributes 2 add instances,
+        // so we divide by 2 here to ensure "instance_num_vars" aligns with the actual number of add instances.
+        let step_loop = 1 << (instance_num_vars - 1);
 
         // init vm.x1 = 1, vm.x2 = -1, vm.x3 = step_loop
         let public_io_init = init_public_io(&[1, u32::MAX, step_loop]);


matthiasgoergens:
> Public IO Auxiliary will cover in next PR

@hero78119 Do you have a link to that PR, please?

ceno_zkvm/examples/riscv_opcodes.rs
```diff
@@ -112,38 +114,49 @@ fn main() {
...
 
-        let pk = zkvm_cs
-            .clone()
-            .key_gen::<Pcs>(pp.clone(), vp.clone(), zkvm_fixed_traces)
-            .expect("keygen failed");
-        let vk = pk.get_vk();
+    for instance_num_vars in args.start..args.end {
+        // The performance benchmark is hook on number of "add" opcode instances.
+        // Each iteration in the loop contributes 2 add instances,
+        // so we divide by 2 here to ensure "instance_num_vars" aligns with the actual number of add instances.
```
matthiasgoergens, line 150: @hero78119 

What does this mean?

What does `instance_num_vars` mean?

kunxian-xia: It means the number of variables in the multilinear polynomial which encodes witnesses of the add opcode circuit. For example, if we have 2^10 add execution traces, then we can use ml polynomials of nv = 10 to encode the witnesses (`rs1.value: UInt, rs2.value: UInt, rd.value: UInt`).

ceno_zkvm/src/tables/ram/ram_impl.rs
```diff
@@ -58,39 +65,67 @@ impl<RAM: RamTable + Send + Sync + Clone> RamTableConfig<RAM> {
...
             [
                 // a v t
-                vec![(RAM::RAM_TYPE as usize).into()],
+                vec![(NVRAM::RAM_TYPE as usize).into()],
                 vec![Expression::Fixed(addr)],
-                final_v.iter().map(|v| v.expr()).collect_vec(),
+                final_v
+                    .as_ref()
+                    .map(|v_limb| v_limb.iter().map(|v| v.expr()).collect_vec())
+                    .unwrap_or_else(|| init_v.iter().map(|v| v.expr()).collect_vec()),
```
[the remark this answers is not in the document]
matthiasgoergens: > If we wanna strongly enforce then we need to do it in lw/sw, but it will be very costly.

It's actually pretty cheap to enforce that.

PR #466, opened 2024-10-27
Remove redundant `Result` handling

matthiasgoergens:
We use a lot of `Result` return types where we don't need to.  Almost all functions we use it for can't actually fail, and for the rest the only 'error-handling' we do in the end it call `.unwrap()`.

So we might as well get rid of these instances of `Result` and make the code base easier to read.  (And slightly faster and smaller.)

```1620 insertions(+), 2314 deletions(-)```

Ie that's about 700 lines gone, 700 lines guaranteed not to introduce any bugs, exploits or backdoors.


lispc:
what is the future of this PR? i noticed several smaller related prs merged?  Are we going to split this into smaller prs and handle them one by one?  Seems for some we have consensus the simplification is good, while for others no consensus?

matthiasgoergens:
> what is the future of this PR? i noticed several smaller related prs merged? Are we going to split this into smaller prs and handle them one by one? Seems for some we have consensus the simplification is good, while for others no consensus?

Thanks for checking in on this.  So the future is https://github.com/scroll-tech/ceno/pull/494 and https://github.com/scroll-tech/ceno/pull/435

https://github.com/scroll-tech/ceno/pull/435 is held up by status quo bias and an idiosyncratic expectation of how optimising compilers work.  (Somehow recognising and removing an unused closure is supposed to be easier for the compiler than recognising and removing an unused pointer to a string literal.)  In any case, the performance doesn't matter at all either way.

For https://github.com/scroll-tech/ceno/pull/494 there was some promise of a a refactoring of how namespaces work, and thus a request to hold back on the PR.  But that promised refactoring has been outstanding since October, so I wouldn't hold my breadth.  (I'm happy to help resolve any merge conflicts of any outstanding refactoring work, if any, if https://github.com/scroll-tech/ceno/pull/494 goes into `master` first.)

I can check what else was in this PR that's not in the two that are still outstanding that I mentioned, but I don't think it was much.

PR #468, opened 2024-10-27
Implement `IntoIterator` and use it for an example

matthiasgoergens:
There's lots more in the code where we can use this to simplify things later.  But I want to keep the PR simple.


matthiasgoergens:
@lispc Here's another simple one that we can get into `master`.

lispc:
> @lispc Here's another simple one that we can get into `master`.

I am neutral for this change.   iter() does not seems so bad for me.  Wait the original struct author to review.

matthiasgoergens:
> I am neutral for this change. iter() does not seems so bad for me. Wait the original struct author to review.

I guess I need to make good on the promise in the PR description to show why this is a useful change:

> There's lots more in the code where we can use this to simplify things later.

`.iter()` by itself isn't too bad, but implementing standard interfaces (that do the same thing) makes our tools more generally useful, because they work better with other functionality that works with standard interfaces.

lispc (approved):
oh i realized before it was a self-named-iter method.

PR #473, opened 2024-10-28
Extract copy-and-pasted function

matthiasgoergens:
We'll likely deduplicate more functionality as we go along.


yczhangsjtu (approved):

matthiasgoergens:
We deleted Singer, and that removed the duplication.

PR #494, opened 2024-10-29
Recognise that `CircuitBuilder::namespace` can't fail

matthiasgoergens:
At least `CircuitBuilder::namespace` won't fail on its own.  But it might fail, if the circus construction function (or so) you are passing to fails.

The new type reflects that.

Extracted from https://github.com/scroll-tech/ceno/pull/466 and addressing https://github.com/scroll-tech/ceno/pull/455/files#r1818907561

This will lead to a lot of clean up down the line, but I want to keep this PR small.


ceno_zkvm/src/chip_handler/general.rs
```diff
@@ -321,8 +321,8 @@ impl<'a, E: ExtensionField> CircuitBuilder<'a, E> {
     pub fn namespace<NR: Into<String>, N: FnOnce() -> NR, T>(
         &mut self,
         name_fn: N,
-        cb: impl FnOnce(&mut CircuitBuilder<E>) -> Result<T, ZKVMError>,
-    ) -> Result<T, ZKVMError> {
+        cb: impl FnOnce(&mut CircuitBuilder<E>) -> T,
```
matthiasgoergens, line 324: @lispc Here's a simple realisation: the `namespace` function as written actually works for any type `T`, not just for `Result`.

Realising that, lets us remove plenty of unnecessary `OK(..)` and `.unwrap()` later down.

It's useful to do this, because `Result` types should be used when we actually have something that can go wrong and should be handled; instead of being dealt with via `.unwrap()` thoughout.  (If all the 'error handling' you ever gonna do is via `.unwrap()`, you might as well `panic!()` instead of creating an `Err` in the first place, at least that gives more useful stack traces.)

matthiasgoergens:
> I have some plan to formalized our namespace to make it more organized. Which means namespace string should meet with some format, or it will return failed With that, please onhold for attempt to change on namespace api (and all the namespace/debug map related) first.

Looks like the namespace reorganisation plan isn't happening.  At least not anytime soon.  Can we please merge this?

matthiasgoergens:
@hero78119 Any news on the promised refactoring?

PR #496, opened 2024-10-30
Instruction category is a function of opcode

matthiasgoergens:
So we don't need to cache it.


matthiasgoergens:
> This implementation is based on a lookup table, why rewrite it with a different technique?

We are storing redundant information in `InsnCodes`.

PR #521, opened 2024-11-01
fix/0-instances

naure:
_Issue #433_ 
_Extracted from #368_


naure:
> equivalent in functionality to claim a list of opcode absence

Indeed, no issues with the constraints. It’s about the IOP:

> …degree of freedom of num-instance seem only categories into num = 0, or num > 0 case…
> …the free-input candidate is the witness commitments…

Right! Or in other words:

1. The prover can choose the set of circuits while knowing the challenge.
2. The witness columns can be interpreted as belonging to different circuits. The prover aligns his _free_ sequence of proofs (`opcode_proofs`) with the _committed_ sequence of polynomials (transcript).

I’m tracking this question in #542. Let’s go this PR to unblock e2e testing, and continue discussion there?

matthiasgoergens:
> @hero78119 I'm not totally convinced by this solution because since the commitments are optional, that gives degrees of freedom for the prover and that's no good with Fiat-Shamir.

In my previous VM we solved a similar problem by just always requiring people to commit to some positive number of instances, and didn't allow zero.  (In our case, all our tables had to be padded to a power of two and for further technical reasons we needed at least four.  So that's what we padded everything to.)

I'm not sure if any of this applies here.

In any case, I'm very in favour of making it possible to prove the execution of a program that has 0 occurrences of specific features or instructions.

kunxian-xia:
> In any case, I'm very in favour of making it possible to prove the execution of a program that has 0 occurrences of specific features or instructions.

If there is no instances of any particular instruction / opcode, why do we bother to prove them as the accumulated `r_eval = 1, w_eval = 1, log_sum = 0`?

hero78119 (approved):
LGTM!

Checked with @dreamATD and also suggested add num_instance to transcript as well for more safety.
Thanks for raising separete issue https://github.com/scroll-tech/ceno/issues/542 and we can move discussion there!

PR #523, opened 2024-11-01
fix/fixed_commit_unwrap

naure:
Integration test currently broken with `RUST_LOG=debug`:

```
thread 'main' panicked at /home/a/INV/scroll/ceno/ceno_zkvm/src/scheme/verifier.rs:761:46:
called `Option::unwrap()` on a `None` value
```

Fixed here.


matthiasgoergens (approved):
Nice find!

PR #527, opened 2024-11-01
emul-cache: Cache instruction decoding

naure:
_Issue #524_

This is an approach to accelerate decoding. The original idea was to pre-decode all instructions of the program.

This PR does it by caching instead. It caches from instruction word to decoded structure. This may make the emulator marginally faster; did not benchmark.

In principle it could cache from PC to decoded, saving one hash table lookup. However, that’s a little more complicated because you’d need to change the tracer too.

### DIFF
--- a/ceno_emul/src/rv32im.rs
+++ b/ceno_emul/src/rv32im.rs
@@ -15,7 +15,7 @@
 // limitations under the License.
 
 use anyhow::{Result, anyhow};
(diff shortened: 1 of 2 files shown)


matthiasgoergens (changes requested):
I don't think this is useful.  It mostly just adds more complexity, and doesn't actually simplify anything.

PR #531, opened 2024-11-01
Feat/MULHSU opcode

bgillesp:
Implements the MULHSU opcode using a circuit design similar to that used by MULH:
1. Map signed and unsigned inputs as corresponding field elements
2. Compute the product of the inputs, and verify that `rd` can be used as the upper limb for a 64-bit 2s complement signed value equal to this product.

A small complication comes up in this design because of the ambiguity of the representation of a small number of 64-bit signed values as Goldilocks field elements.  While this ambiguity did not have an impact on the signed/signed products for the MULH opcode, here the interval of signed/unsigned product values does overlap the ambiguous interval of 64-bit 2s complement values.  This is mitigated by explicitly constraining the value of the `rd` register to be not equal to its maximal signed value (`2^31 - 1`) by witnessing the inverse of `(2^31 - 1) - rd`.  This makes all remaining valid 64-bit values unambiguously represented, and also still permits representation of all valid signed/unsigned product values.  See the module doc comment for `mulh.rs` for a more complete soundness analysis.  (Some extra eyes on this analysis during the review would probably be a good idea, as the details are quite sensitive to the ranges of values and the size of the Goldilocks field.)

This PR additionally refactors the existing MULHU opcode to use an analogous circuit design to this, also requiring a one-element restriction on the value of `rd` to ensure soundness.  This implementation avoids the overhead of using `UInt::mul` to compute the product explicitly in `u16` limbs, which requires additional carry witnesses with range constraints, and should reduce the total number of `WitIn`s needed for the circuit fairly substantially.

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/mulh.rs
+++ b/ceno_zkvm/src/instructions/riscv/mulh.rs
@@ -154,6 +154,7 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for MulhInstructionBas
     fn name() -> String {
         format!("{:?}", I::INST_KIND)
     }
+
     fn construct_circuit(
         circuit_builder: &mut CircuitBuilder<E>,
     ) -> Result<MulhConfig<E>, ZKVMError> {
--- a/ceno_zkvm/src/instructions/riscv/mulh.rs
+++ b/ceno_zkvm/src/instructions/riscv/mulh.rs
@@ -82,6 +82,7 @@ use std::{fmt::Display, marker::PhantomData};
 
 use ceno_emul::{InsnKind, StepRecord};
 use ff_ext::ExtensionField;
+use goldilocks::SmallField;
 
 use crate::{
     circuit_builder::CircuitBuilder,
@@ -156,6 +157,14 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for MulhInstructionBas
     fn construct_circuit(
         circuit_builder: &mut CircuitBuilder<E>,
     ) -> Result<MulhConfig<E>, ZKVMError> {
+        // The soundness analysis for these constraints is only valid for
+        // 32-bit registers represented over the Goldilocks field, so verify
+        // these parameters
+        assert_eq!(UInt::<E>::TOTAL_BITS, u32::BITS as usize);
+        assert_eq!(E::BaseField::MODULUS_U64, goldilocks::MODULUS);
+
+        // 0. Registers and instruction lookup
+
         let rs1_read = UInt::new_unchecked(|| "rs1_read", circuit_builder)?;
         let rs2_read = UInt::new_unchecked(|| "rs2_read", circuit_builder)?;
         let rd_written = UInt::new(|| "rd_written", circuit_builder)?;
--- a/ceno_zkvm/src/instructions/riscv/mulh.rs
+++ b/ceno_zkvm/src/instructions/riscv/mulh.rs
@@ -107,6 +107,12 @@ impl RIVInstruction for MulhuOp {
 }
 pub type MulhuInstruction<E> = MulhInstructionBase<E, MulhuOp>;
 
+pub struct MulhsuOp;
+impl RIVInstruction for MulhsuOp {
+    const INST_KIND: InsnKind = InsnKind::MULHSU;
+}
+pub type MulhsuInstruction<E> = MulhInstructionBase<E, MulhsuOp>;
+
 pub struct MulhConfig<E: ExtensionField> {
     rs1_read: UInt<E>,
     rs2_read: UInt<E>,
@@ -120,7 +126,6 @@ enum MulhSignDependencies<E: ExtensionField> {
     UU {
         constrain_rd_inv: WitIn,
     },
-    #[allow(dead_code)]
     SU {
         rs1_signed: Signed<E>,
         rd_signed: Signed<E>,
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -7,195 +87,170 @@ use crate::{
...
-                    rs1_read,
-                    rs2_read,
-                    rd_written,
-                    rd_written_hi.register_expr(),
-                )
-            }
-
-            _ => unreachable!("Unsupported instruction kind"),
-        };
+    ) -> Result<MulhConfig<E>, ZKVMError> {
```
hero78119, line 158: Thanks for the documentation! it's great help to grasp the high level idea!

Giving currently implementation are strong coupling with field size, due to 64 bit representation got some overlap with Goldilock max field.
How about we also add some assertion, e.g. 

```
assert_eq!(UInt::<E>::TOTAL_BITS, u32::BITS as usize);
assert_eq!(E::BaseField::MODULUS_U64, goldilocks::MODULUS);
```
so once in the future we are having the different field implementation, then it can help to hint for implementation/constrain need to tweak a bit

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -7,195 +87,170 @@ use crate::{
...
+pub struct MulhConfig<E: ExtensionField> {
+    rs1_read: UInt<E>,
+    rs2_read: UInt<E>,
+    rd_written: UInt<E>,
+    sign_deps: MulhSignDependencies<E>,
+    prod_low: UInt<E>,
+    r_insn: RInstructionConfig<E>,
+}
+
+enum MulhSignDependencies<E: ExtensionField> {
```
KimiWu123, line 134: This is a good idea to make the declaration of `_Config` more succinct. I believe there are many opcodes can apply this pattern. What do you think, @hero78119 ?

hero78119: definitely it can clean up on opcode which expose `Option<>` as internal field, so we can control them clearly in enum
@KimiWu123 if you interesting in the cleanup it's good to have an issue to track it :)

ceno_zkvm/src/instructions/riscv/mulh.rs
bgillesp: Yes, definitely agree!  Just added these to the start of `construct_circuit`.

hero78119 (approved):
Epic works 👍

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -1,201 +1,266 @@
...
+                )
+            }
+
+            InsnKind::MULHSU => {
+                let rs1_signed = Signed::construct_circuit(circuit_builder, || "rs1", &rs1_read)?;
+                let rd_signed = Signed::construct_circuit(circuit_builder, || "rd", &rd_written)?;
 
-                let rd_written = rs1_read.mul_hi(&rs2_read, lk_multiplicity, true);
+                // constrain that (signed) rd does not represent 2^31 - 1
+                let rd_avoid = Expression::<E>::from((1u64 << (BIT_WIDTH - 1)) - 1);
```
matthiasgoergens, line 224: `i32::MAX` would have been a more readable way to write this?

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -1,201 +1,266 @@
...
-        // field with order `p = 2^64 - 2^32 + 1`.
-
-        let r_insn = RInstructionConfig::<E>::construct_circuit(
-            circuit_builder,
-            InsnKind::MULH,
-            rs1_read.register_expr(),
-            rs2_read.register_expr(),
-            rd_written.register_expr(),
+            rs1_val * rs2_val,
+            rd_val * (1u64 << 32) + prod_low.value(),
```
matthiasgoergens, line 255: `(rd_val << 32) + prod_low.value()` should just work?

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -207,47 +272,79 @@ impl<E: ExtensionField> Instruction<E> for MulhInstruction<E> {
...
+                let rs2_s = rs2_signed.assign_instance(instance, lk_multiplicity, &rs2_val)?;
+                rd_signed.assign_instance(instance, lk_multiplicity, &rd_val)?;
+
+                let prod = (rs1_s as i64) * (rs2_s as i64);
+                ((prod as u64) % (1u64 << BIT_WIDTH)) as u32
+            }
+            MulhSignDependencies::UU { constrain_rd } => {
+                // assign nonzero value (u32::MAX - rd)
+                let rd_f = E::BaseField::from(rd as u64);
+                let avoid_f = E::BaseField::from((1u64 << BIT_WIDTH) - 1);
```
matthiasgoergens, line 316: `u32::MAX` would probably be easier to read?

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -207,47 +272,79 @@ impl<E: ExtensionField> Instruction<E> for MulhInstruction<E> {
...
+                constrain_rd.assign_instance(instance, rd_f, avoid_f)?;
+
+                let prod = rs1_val.as_u64() * rs2_val.as_u64();
+                (prod % (1u64 << BIT_WIDTH)) as u32
```
matthiasgoergens, line 320: Why not just `prod as u32`?

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -207,47 +272,79 @@ impl<E: ExtensionField> Instruction<E> for MulhInstruction<E> {
...
+            } => {
+                // Signed register values
+                let rs1_s = rs1_signed.assign_instance(instance, lk_multiplicity, &rs1_val)?;
+                let rd_s = rd_signed.assign_instance(instance, lk_multiplicity, &rd_val)?;
+
+                // assign nonzero value (i32::MAX - rd)
+                let rd_f = i64_to_base(rd_s as i64);
+                let avoid_f = i64_to_base((1i64 << (BIT_WIDTH - 1)) - 1);
+                constrain_rd.assign_instance(instance, rd_f, avoid_f)?;
+                // let nonzero_val = ((1i64 << (BIT_WIDTH - 1)) - 1) - (rd_s as i64);
```
matthiasgoergens, line 335: Please remove commented out code.

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -207,47 +272,79 @@ impl<E: ExtensionField> Instruction<E> for MulhInstruction<E> {
...
+                // constrain_rd.assign_instance(instance, i64_to_base(nonzero_val))?;
+
+                let prod = (rs1_s as i64).wrapping_mul(rs2 as i64);
+                ((prod as u64) % (1u64 << BIT_WIDTH)) as u32
```
matthiasgoergens, line 339: `prod as u32` should do the same thing and be easier to understand?

matthiasgoergens (commented):
A post-mortem review.

PR #536, opened 2024-11-02
refactor/move-circuit-imm: Move circuit-specific logic out of the emulator

naure:
- Move any circuit-specific logic of the immediate column of the program table, into the program table module.
- Simplify InsnRecord back to a simple list.
- Minimize API into a single function `imm_internal`. It uses an intermediate i64 instead of previous u32 and bool.

This was an abstraction leakage across crates. This PR finally moves it at the right place.

### DIFF
--- a/ceno_emul/src/rv32im.rs
+++ b/ceno_emul/src/rv32im.rs
@@ -270,42 +270,23 @@ impl DecodedInstruction {
         }
     }
 
-    /// The internal view of the immediate, for use in circuits.
-    pub fn imm_internal(&self) -> u32 {
+    pub fn codes(&self) -> InsnCodes {
+        FastDecodeTable::get().lookup(self)
+    }
+
+    pub fn immediate_msb(&self) -> u32 {
+        self.top_bit
+    }
+
+    pub fn immediate(&self) -> u32 {
         match self.codes().format {
-            R => self.func7,
-            I => match self.codes().kind {
-                // decode the shift as a multiplication/division by 1 << immediate
-                SLLI | SRLI | SRAI => 1 << self.imm_shamt(),
-                _ => self.imm_i(),
-            },
+            R => 0,
+            I => self.imm_i(),
             S => self.imm_s(),
             B => self.imm_b(),
             U => self.imm_u(),
             J => self.imm_j(),
         }
-    }
-
-    /// The internal interpretation of the immediate sign, for use in circuits.
-    /// Indicates if the immediate value, when signed, needs to be encoded as field negative.
-    /// example:
-    /// imm = ux::MAX - 1 implies
-    /// imm_field = FIELD_MODULUS - 1 if imm_field_is_negative
-    /// imm_field = ux::MAX - 1 otherwise
-    /// see InsnRecord::imm_internal_field
-    pub fn imm_field_is_negative(&self) -> bool {
-        match self.codes() {
-            InsnCodes { format: R | U, .. } => false,
-            InsnCodes {
-                kind: SLLI | SRLI | SRAI | ADDI | SLTIU | ANDI | XORI | ORI,
-                ..
-            } => false,
-            _ => self.top_bit != 0,
-        }
-    }
-
-    pub fn codes(&self) -> InsnCodes {
-        FastDecodeTable::get().lookup(self)
     }
 
     fn imm_b(&self) -> u32 {
@@ -317,11 +298,6 @@ impl DecodedInstruction {
 
     fn imm_i(&self) -> u32 {
         (self.top_bit * 0xffff_f000) | (self.func7 << 5) | self.rs2
-    }
-
-    /// Shift amount field of SLLI, SRLI, SRAI.
-    fn imm_shamt(&self) -> u32 {
-        self.rs2
     }
 
     fn imm_s(&self) -> u32 {
@@ -339,33 +315,6 @@ impl DecodedInstruction {
 
     fn imm_u(&self) -> u32 {
         self.insn & 0xfffff000
-    }
-}
-
-#[cfg(test)]
-#[test]
-#[allow(clippy::identity_op)]
-fn test_decode_imm() {
-    for (i, expected) in [
-        // Example of I-type: ADDI.
-        // imm    | rs1     | funct3      | rd     | opcode
-        (89 << 20 | 1 << 15 | 0b000 << 12 | 1 << 7 | 0x13, 89),
-        // Shifts get a precomputed power of 2: SLLI, SRLI, SRAI.
-        (31 << 20 | 1 << 15 | 0b001 << 12 | 1 << 7 | 0x13, 1 << 31),
-        (31 << 20 | 1 << 15 | 0b101 << 12 | 1 << 7 | 0x13, 1 << 31),
-        (
-            1 << 30 | 31 << 20 | 1 << 15 | 0b101 << 12 | 1 << 7 | 0x13,
-            1 << 31,
-        ),
-        // Example of R-type with funct7: SUB.
-        // funct7     | rs2    | rs1     | funct3      | rd     | opcode
-        (
-            0x20 << 25 | 1 << 20 | 1 << 15 | 0 << 12 | 1 << 7 | 0x33,
-            0x20,
-        ),
-    ] {
-        let imm = DecodedInstruction::new(i).imm_internal();
-        assert_eq!(imm, expected);
     }
 }
 
--- a/ceno_emul/src/rv32im.rs
+++ b/ceno_emul/src/rv32im.rs
@@ -262,10 +262,6 @@ impl DecodedInstruction {
         }
     }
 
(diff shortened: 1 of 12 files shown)


ceno_zkvm/src/instructions/riscv/shift_imm.rs
```diff
@@ -181,6 +181,12 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for ShiftImmInstructio
     }
 }
 
+/// Decode the immediate into the format of this circuit.
+/// The shift is implemented as a multiplication/division by 1 << immediate.
+pub fn process_imm(insn: &DecodedInstruction) -> u32 {
```
KimiWu123, line 186: I would like to propose to move this function to `program.rs`. It would be easier for circuit writers, they don't need to care the bits shifting. What they need to do is just calling `process_imm` and `process_imm` will return what they expected.

ceno_zkvm/src/instructions/riscv/slti.rs
```diff
@@ -105,8 +105,8 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for SetLessThanImmInst
             .rs1_read
             .assign_value(instance, Value::new_unchecked(rs1));
 
-        let imm = step.insn().imm_internal();
-        let imm_field = InsnRecord::imm_internal_field::<E::BaseField>(&step.insn());
+        let imm = step.insn().immediate();
```
KimiWu123, line 108: I think we can replace it with `InsnRecord::process_imm` here and as well as in all the code above, to make it more unified.`InsnRecord::process_imm` seems like a place for special cases, similar to original `imm_internal()`. By unifying calls to `process_imm`, circuit writers won’t have to decide whether to use `step.insn().immediate()` or `InsnRecord::process_imm()`.

ceno_emul/src/rv32im.rs
```diff
@@ -270,44 +270,25 @@ impl DecodedInstruction {
         }
     }
 
-    /// The internal view of the immediate, for use in circuits.
-    pub fn imm_internal(&self) -> u32 {
+    pub fn codes(&self) -> InsnCodes {
+        FastDecodeTable::get().lookup(self)
+    }
+
+    pub fn immediate_msb(&self) -> u32 {
```
matthiasgoergens, line 277: One reason to have getters instead of direct access to public struct member is to do error handling.

This getter is only valid for instructions that have an immediate field.  (And even for those, it only makes sense for immediate shifts, I think?)  But the getter here is happy to silently hand you garbage data.

As a minimal change, you could at least make it return a consistent 0 for instructions that don't have an immediate field in the spec.

(But more holistically, we should probably move to `rrs-lib` for decoding like SP1 does.)

ceno_zkvm/src/tables/program.rs
```diff
@@ -46,38 +49,56 @@ impl<T> InsnRecord<T> {
...
     }
 
     /// Interpret the immediate or funct7 as unsigned or signed depending on the instruction.
     /// Convert negative values from two's complement to field.
-    pub fn imm_internal_field<F: SmallField>(insn: &DecodedInstruction) -> F {
-        if insn.imm_field_is_negative() {
-            -F::from(-(insn.imm_internal() as i32) as u64)
+    pub fn imm_internal_field(insn: &DecodedInstruction) -> F {
+        let imm = Self::process_imm(insn);
+        if Self::imm_field_is_negative(insn) {
```
matthiasgoergens, line 71: Since you are touching this anyway:

We can just represent the immediate field as an i64, and then use `i64_to_base` (or one her slightly renamed copy-and-paste sisters) to convert to a field element.  Instead of re-implementing that logic yet another time.

ceno_zkvm/src/tables/program.rs
```diff
@@ -46,38 +49,56 @@ impl<T> InsnRecord<T> {
...
-            F::from(insn.imm_internal() as u64)
+            F::from(imm as u64)
+        }
+    }
+
+    /// The internal view of the immediate, for use in circuits.
+    fn process_imm(insn: &DecodedInstruction) -> u32 {
+        let codes = insn.codes();
+        match codes.format {
+            R => codes.func7,
```
matthiasgoergens, line 82: R-types don't have an immediate.  Why does it show up here?

The doc here says 'for use in circuits.'  Why do circuits care about funct7 at all?  That's only useful for determining the kind of instruction, and we can already read that directly from the struct.  I'm a bit confused.

ceno_zkvm/src/tables/program.rs
```diff
@@ -46,38 +49,56 @@ impl<T> InsnRecord<T> {
...
+            I if matches!(codes.kind, SLLI | SRLI | SRAI) => shift_imm::process_imm(insn),
+            _ => insn.immediate(),
+        }
+    }
+
+    /// The internal interpretation of the immediate sign, for use in circuits.
```
matthiasgoergens, line 88: Just report the immediate as a `i64` for use by circuits.  They can then directly encode that as a field element.

No need to make them (re-) do a bunch of logic.

matthiasgoergens (commented):
I'm a bit confused by what we are trying to do here and why.

ceno_zkvm/src/tables/program.rs
naure: Need to merge #538 into this PR.

naure: done

ceno_zkvm/src/tables/program.rs
naure: Can do it that way now yes. This PR is a code move, I’d rather do implementation refactoring separately.

ceno_zkvm/src/instructions/riscv/shift_imm.rs
matthiasgoergens: We should just record the proper shift amount in a field named 'immediate' in the Decoded instruction.

naure: Done.

ceno_zkvm/src/instructions/riscv/slti.rs
naure: Right! I will put it back that way.

ceno_zkvm/src/tables/program.rs
naure: Done in #546

ceno_emul/src/rv32im.rs
naure: Removed in #546.

ceno_zkvm/src/instructions/riscv/jump/auipc.rs
```diff
@@ -74,10 +76,10 @@ impl<E: ExtensionField> Instruction<E> for AuipcInstruction<E> {
         step: &ceno_emul::StepRecord,
     ) -> Result<(), ZKVMError> {
         let pc: u32 = step.pc().before.0;
-        let imm: u32 = step.insn().imm_internal();
-        let (sum, overflow) = pc.overflowing_add(imm);
+        let imm = InsnRecord::imm_internal(&step.insn());
+        let (sum, overflow) = pc.overflowing_add(imm as u32);
 
-        set_val!(instance, config.imm, imm as u64);
+        set_val!(instance, config.imm, i64_to_base::<E::BaseField>(imm));
```
matthiasgoergens, line 82: It's a bit of a shame that we need to add `<E::BaseField>` everywhere now.  I wonder if we can avoid that.

ceno_zkvm/src/tables/program.rs
```diff
@@ -145,21 +159,10 @@ impl<E: ExtensionField, const PROGRAM_SIZE: usize> TableCircuit<E>
                 let insn = DecodedInstruction::new(program.instructions[i]);
                 let values = InsnRecord::from_decoded(pc, &insn);
 
-                // Copy all the fields except immediate.
-                for (col, val) in config
-                    .record
-                    .without_imm()
```
matthiasgoergens, line 151: Nice that we can remove that.

matthiasgoergens (approved):

ceno_zkvm/src/instructions/riscv/jump/auipc.rs
naure: What about a From<i64>?

ceno_zkvm/src/tables/program.rs
naure: Yes that was itching.

ceno_zkvm/src/instructions/riscv/jump/auipc.rs
matthiasgoergens: That would be best, but to make that work, you need to make a change to the crate that implements Goldilocks. (Or you need to introduce an extra wrapper.)

Rust doesn't allow 'orphans'.

Issue #533, opened 2024-11-02
Switch to `InsnKind` in the program table


naure:
Idea extracted from discussions with @matthiasgoergens ([thread](https://github.com/scroll-tech/ceno/pull/519#discussion_r1826041801)).

All circuits support a single constant opcode/funct3/funct7, so they can be merged into a single constant with `enum InsnKind`.

This saves 1 fixed column, 1 fixed record field, and cleans up a bit.

Note: One day we might merge some opcode circuits together or split circuits for special cases, to improve the verifier or prover cost respectively. But that will probably not match the original opcode/etc classification. So there’s no drawback in removing it.

- [x] Double-check that `opcode/funct3/funct7` are constants in all circuits.
- [x] Change the layout of the program table.
- [x] Change the calls to `circuit_builder.lk_fetch`.
- [x] Remove `func7` from `imm_internal`.

PR #538, opened 2024-11-02
Switch to `InsnKind` in the program table

naure:
_Matthias:_

Risc-V instruction are encoded in multiple different 'types'.  These types have various fields.  One particularly common field is `opcode`, which typically (but not always) occupies the 7 least significant bits.

As the name suggests, opcode helps code _which_ operation is intended.  I say 'helps' because some operations share the same opcode.

Our code before this PR assumed that bits `12..=14` in the instruction word are exactly enough extra information to determine the operation.  (Confusingly enough, our code follows Risc0 convention, and calls that range of bits `funct3`.  Only some operations have a field named `funct3` in the spec, and not all of them have it in the same place.)  That assumption is wrong, and our circuits are unsound.  Specifically, the distinction between 'signed' (ie 'arithmetic') and 'unsigned' (ie 'logical') immediate right shifts is encoded in bit 30, as laid out in section 2.4.1 of the [spec](https://github.com/riscv/riscv-isa-manual/releases/download/20240411/unpriv-isa-asciidoc.pdf).

We could introduce a new column called 'bit 30' in the program table to fix the symptom.  But instead of such a piecemeal approach, we attack the underlying problem: we switch the program table to record `InsnKind`.  `InsnKind` directly reports the kind of instruction we are dealing with.  We already have that information available in the decoder, and just re-use it here.

Fixes https://github.com/scroll-tech/ceno/issues/539 and https://github.com/scroll-tech/ceno/issues/533

### DIFF
--- a/ceno_emul/src/rv32im_encode.rs
+++ b/ceno_emul/src/rv32im_encode.rs
@@ -49,7 +49,8 @@ const fn encode_i(kind: InsnKind, rs1: u32, rd: u32, imm: u32) -> u32 {
     let rd = rd & MASK_5_BITS;
     let func3 = kind.codes().func3;
     let opcode = kind.codes().opcode;
-    let imm = imm & MASK_12_BITS;
+    let shift_func = (matches!(kind, InsnKind::SRAI) as u32) << 10; // SRAI kind is encoded in imm.
+    let imm = imm & MASK_12_BITS | shift_func;
     imm << 20 | rs1 << 15 | func3 << 12 | rd << 7 | opcode
 }
 
--- a/ceno_emul/src/rv32im_encode.rs
+++ b/ceno_emul/src/rv32im_encode.rs
@@ -49,8 +49,9 @@ const fn encode_i(kind: InsnKind, rs1: u32, rd: u32, imm: u32) -> u32 {
     let rd = rd & MASK_5_BITS;
     let func3 = kind.codes().func3;
     let opcode = kind.codes().opcode;
-    let shift_func = (matches!(kind, InsnKind::SRAI) as u32) << 10; // SRAI kind is encoded in imm.
-    let imm = imm & MASK_12_BITS | shift_func;
+    // SRLI/SRAI use a specialization of the I-type format with the shift type in imm[10].
+    let is_arithmetic_right_shift = (matches!(kind, InsnKind::SRAI) as u32) << 10;
+    let imm = imm & MASK_12_BITS | is_arithmetic_right_shift;
     imm << 20 | rs1 << 15 | func3 << 12 | rd << 7 | opcode
 }
 
--- a/ceno_zkvm/src/instructions/riscv/ecall_insn.rs
+++ b/ceno_zkvm/src/instructions/riscv/ecall_insn.rs
@@ -39,9 +39,8 @@ impl EcallInstructionConfig {
 
         cb.lk_fetch(&InsnRecord::new(
             pc.expr(),
-            (EANY.codes().opcode as usize).into(),
+            EANY.into(),
             None,
-            (EANY.codes().func3 as usize).into(),
             0.into(),
             0.into(),
             0.into(), // imm = 0
--- a/ceno_emul/src/rv32im.rs
+++ b/ceno_emul/src/rv32im.rs
@@ -197,9 +197,9 @@ pub struct InsnCodes {
     pub format: InsnFormat,
     pub kind: InsnKind,
     category: InsnCategory,
-    pub opcode: u32,
-    pub func3: u32,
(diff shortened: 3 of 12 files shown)


ceno_emul/src/rv32im_encode.rs
```diff
@@ -49,7 +49,8 @@ const fn encode_i(kind: InsnKind, rs1: u32, rd: u32, imm: u32) -> u32 {
     let rd = rd & MASK_5_BITS;
     let func3 = kind.codes().func3;
     let opcode = kind.codes().opcode;
-    let imm = imm & MASK_12_BITS;
+    let shift_func = (matches!(kind, InsnKind::SRAI) as u32) << 10; // SRAI kind is encoded in imm.
```
hero78119, line 52: Could we elaborate more on this operation? It looks to me like an special encoding outside of riscv spec

hero78119 (approved):
Change looks pretty neat to me! Just left a question on of `encode_i` handling

KimiWu123 (approved):

hero78119:
Btw, as https://github.com/scroll-tech/ceno/issues/539 point out the bug.
But the fixed in PR is not that obvious, could you explain a bit on PR description :)
Thanks!

ceno_emul/src/rv32im_encode.rs
matthiasgoergens: Adding a comment might be useful, but as far as I can tell, it's pretty much in line with the spec.

See section 2.4.1 of the [spec](https://github.com/riscv/riscv-isa-manual/releases/download/20240411/unpriv-isa-asciidoc.pdf).

ceno_zkvm/src/instructions/riscv/ecall_insn.rs
```diff
@@ -39,9 +39,8 @@ impl EcallInstructionConfig {
 
         cb.lk_fetch(&InsnRecord::new(
             pc.expr(),
-            (EANY.codes().opcode as usize).into(),
```
matthiasgoergens, line 42: `EANY` should disappear as a kind, and be replaced with `ECALL` and `EBREAK`.

`EANY` ain't even an instruction that appears in Risc-V.  It's just an artefact of Risc0's design decisions.

(But we can do that in a subsequent PR.)

matthiasgoergens (approved):
Thanks!

matthiasgoergens:
> But the fixed in PR is not that obvious, could you explain a bit on PR description :)

Good call!  Please check out the shiny new PR description I wrote. @hero78119 

Feel free to either complain about why wording here (and I'll do something about it) or just directly fix it yourself.

ceno_emul/src/rv32im_encode.rs
hero78119: I see, thanks for the reference

Then it might make thing clear
```diff
diff --git a/ceno_emul/src/rv32im_encode.rs b/ceno_emul/src/rv32im_encode.rs
index 03128c27..61eed9fe 100644
--- a/ceno_emul/src/rv32im_encode.rs
+++ b/ceno_emul/src/rv32im_encode.rs
@@ -49,8 +49,8 @@ const fn encode_i(kind: InsnKind, rs1: u32, rd: u32, imm: u32) -> u32 {
     let rd = rd & MASK_5_BITS;
     let func3 = kind.codes().func3;
     let opcode = kind.codes().opcode;
-    let shift_func = (matches!(kind, InsnKind::SRAI) as u32) << 10; // SRAI kind is encoded in imm.
-    let imm = imm & MASK_12_BITS | shift_func;
+    let is_arithmetic_right_shift = (matches!(kind, InsnKind::SRAI) as u32) << 10; // SRLI/SRAI kind is encoded in imm[10].
+    let imm = imm & MASK_12_BITS | is_arithmetic_right_shift;
     imm << 20 | rs1 << 15 | func3 << 12 | rd << 7 | opcode
 }
```

matthiasgoergens: The rename looks good to me.  One cosmetic suggestion: put `// SRLI/SRAI kind is encoded in imm[10].` before the line, not behind it.

naure:
@hero78119 The bug is fixed because now the encoding of the program table (`kind`) matches exactly what the circuits do (implement one kind).

ceno_emul/src/rv32im_encode.rs
naure: done

ceno_zkvm/src/instructions/riscv/ecall_insn.rs
naure: That makes sense.

PR #543, opened 2024-11-04
Update BaseFold number of queries to 200 according to the recent papers.

yczhangsjtu:
Extracting small PRs from #294

### DIFF
--- a/mpcs/src/basefold/encoding/rs.rs
+++ b/mpcs/src/basefold/encoding/rs.rs
@@ -193,6 +193,15 @@ pub fn coset_fft<E: ExtensionField>(
 pub struct RSCodeDefaultSpec {}
 
 impl RSCodeSpec for RSCodeDefaultSpec {
+    // According to Theorem 1 of paper <BaseFold in the List Decoding Regime>
+    // (https://eprint.iacr.org/2024/1571), the soundness error is bounded by
+    // $O(1/|F|) + (\sqrt{\rho}+\epsilon)^s$
+    // where $s$ is the query complexity and $\epsilon$ is a small value
+    // that can be ignored. So the number of queries can be estimated by
+    // $$
+    // \frac{2\lambda}{-\log\rho}
+    // $$
+    // If we take $\lambda=100$ and $\rho=1/2$, then the number of queries is $200$.
     fn get_number_queries() -> usize {
         200
     }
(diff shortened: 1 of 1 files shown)


mpcs/src/basefold/encoding/rs.rs
```diff
@@ -194,7 +194,7 @@ pub struct RSCodeDefaultSpec {}
 
 impl RSCodeSpec for RSCodeDefaultSpec {
     fn get_number_queries() -> usize {
-        972
+        200
```
kunxian-xia, line 197: Can you put some links or comments here to justify the reduction?

yczhangsjtu: Sure. I'll copy from the original PR.

yczhangsjtu: According to Theorem 1 of paper <BaseFold in the List Decoding Regime> (https://eprint.iacr.org/2024/1571), the soundness error is bounded by $O(1/|F|) + (\sqrt{\rho}+\epsilon)^s$ where $s$ is the query complexity and $\epsilon$ is a small value that can be ignored. So the number of queries can be estimated by

$$
\frac{2\lambda}{-\log\rho}
$$

If we take $\lambda=100$ and $\rho=1/2$, then the number of queries is $200$.

hero78119: Thanks for the change, I will try to verify with quick benchmark and report e2e data here

matthiasgoergens: @yczhangsjtu Could you make that explanation part of the code base, perhaps as a comment?

kunxian-xia (approved):

mpcs/src/basefold/encoding/rs.rs
hero78119: Hi @yczhangsjtu I try run benchmark on 2^20 which involve basefold commit/opening
On ceno server, benchmark result shows no significant change before/after. 
does it fit the expectation, or do you have any benchmark result for that change?
Just double check in case the change actually not take effect

kunxian-xia: We should expect a big reduction of the proof size. @yczhangsjtu Can you give us how much bytes do we saved after applying this PR?

PR #546, opened 2024-11-04
Convert internal immediate through i64

naure:
_Follow-up to #536 _

Consolidate the processing of immediate values into a single `i64` instead of previous `u32` and `bool`.


naure:
@matthiasgoergens The end result is nice, but please double-check all the conversions.

PR #547, opened 2024-11-04
feat : fibonacci with risc-v in zkvm implemented

### LINKED ISSUE #353: Fibonacci program in riscv and zkVM

mahmudsudo:
closes #353

### DIFF
--- a/ceno_zkvm/examples/fibonacci.rs
+++ b/ceno_zkvm/examples/fibonacci.rs
@@ -1,0 +1,139 @@
+use std::{panic, time::Instant};
+
+use ceno_zkvm::{
+    declare_program,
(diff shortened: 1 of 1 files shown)


kunxian-xia:
Unfortunately this is not what we have in mind. We plan to write the fibonacci in Rust and then use ceno's rust toolchain to compile the source to ELF. This will then be proved/verified by ceno prover/verifier.

matthiasgoergens:
> Unfortunately this is not what we have in mind. We want the fibonacci guest program to be written in Rust and then use ceno's rust toolchain to compile the source to ELF. This will then be proved/verified by ceno prover/verifier.

Yes, exactly.  We don't particularly care about Fibonacci, it's just an example of a simple ELF.

lispc:
closed since it is not n

Issue #316, opened 2024-10-04
[placeholder issue] for recording old crate removal commit hash


matthiasgoergens:
Commit 0a407211dcbb5eeb9e0b02438c92ed46b3197c5f from https://github.com/scroll-tech/ceno/pull/89 broke `singer-pro`.  In https://github.com/scroll-tech/ceno/pull/314 we removed it temporarily.

This issue is here as a place-holder to remind us that we want to eventually restore the package.

TODO(Matthias): post instruction for how to do that, once the PR is merged, and we thus have a commit hash I can use in the instructions here.

_Originally posted by @hero78119 in https://github.com/scroll-tech/ceno/issues/314#issuecomment-2392919107_

PR #548, opened 2024-11-05
Remove unused sub-crates from `master`

matthiasgoergens:
We still have them in the git history, if/when we want them back.

@hero78119 Thanks for the suggestion.


hero78119 (approved):
👍 thanks for the effort!
We have an issue to track "singer-pro" after its removed, probably this PR can also point to same issue for reference.

hero78119:
related to #316

Cargo.lock
```diff
@@ -223,9 +223,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.1.31"
+version = "1.1.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2e7962b54006dcfcc61cb72735f4d89bb97061dd6a7ed882ec6b8ee53714c6f"
+checksum = "67b9470d453346108f93a59222a9a1a5724db32d0a4727b7ab7ace4b4d822dc9"
```
kunxian-xia, line 228: Is it possible to revert the upgrade of these crates.

matthiasgoergens: Can.  I just recreated the lock file.

matthiasgoergens: Done.

PR #549, opened 2024-11-05
Remove `Rng` from PCS type paramter.

yczhangsjtu:
Extracting a small PR from #294.

The random generator is used only when the encoding scheme is Basecode, and not used for RS code. So remove it from the type parameter for Basefold, i.e., `Basefold<E, Spec, Rng>` is replaced with `Basefold<E, Spec>` to make the API simpler. Fix it to `ChaCha8Rng` in the implementation of Basecode.


matthiasgoergens (approved):
Thanks for breaking this off into a simple and easily reviewable PR.

kunxian-xia (approved):

PR #552, opened 2024-11-05
Refactor BaseFold test code

yczhangsjtu:
Extracting small PR from #294 

Refactor the test code in `mpcs/src/lib.rs`. Define some utility functions to simplify the tests.

Also merged the test functions in `mpcs/src/basefold.rs`.

PR #554, opened 2024-11-05
Merge two similar benchmarks in BaseFold into one

yczhangsjtu:
Extracting small PR from #294.

The two benchmarks `benches_commit_verify_rs` and `benches_commit_verify_basecode` are similar, differing only in the choice of parameter. Merge them into one benchmark file.

Also simplify the codes using the test utility functions introduced in #552 

Waiting for #552 to merge.


mpcs/src/lib.rs
```diff
@@ -357,14 +357,27 @@ fn err_too_many_variates(function: &str, upto: usize, got: usize) -> Error {
     })
 }
 
-#[cfg(test)]
+// TODO: Need to use some functions here in the integration benchmarks. But
+// unfortunately integration benchmarks do not compile the #[cfg(test)]
+// code. So remove the gate for the entire module, only gate the test
+// functions.
+// TODO(Matthias): use cfg(or(test, bench)) or something like that.
```
matthiasgoergens, line 364: We might want to actually do this. :)  Let me see what I can do for you.

mpcs/src/lib.rs
```diff
@@ -377,29 +388,26 @@ pub mod test_util {
...
-            DenseMultilinearExtension::from_evaluations_ext_vec(
-                num_vars,
-                (0..(1 << num_vars))
-                    .map(|_| E::random(&mut OsRng))
-                    .collect_vec(),
-            )
-        }
+    pub fn gen_rand_poly_base<E: ExtensionField>(num_vars: usize) -> DenseMultilinearExtension<E> {
+        DenseMultilinearExtension::random(num_vars, &mut OsRng)
+    }
```
lispc, line 393: in lib code (instead of testing code), we usually don't hard code RNG directly.   (besides, OsRng is slow actually, most of the time we need some kind of lightweighted rng like CHACHA RNG, that would be enough. 
  So i suggest give user  of this function the freedom to choose rng , and it is ok to use osrng in testing codes)

mpcs/src/lib.rs
```diff
@@ -357,14 +357,25 @@ fn err_too_many_variates(function: &str, upto: usize, got: usize) -> Error {
...
 }
 
-#[cfg(test)]
+// TODO: Need to use some functions here in the integration benchmarks. But
+// unfortunately integration benchmarks do not compile the #[cfg(test)]
+// code. So remove the gate for the entire module, only gate the test
+// functions.
+// This is not the best way: the test utility functions should not be
+// compiled in the release build. Need a better solution.
+#[doc(hidden)]
```
lispc, line 366: in zkevm-circuits, we have a "test" feature flag. and we write "#[cfg(any(feature = "test", test))]" a lot of places.

lispc (approved):

mpcs/src/lib.rs
yczhangsjtu: That means we run benchmarks using `cargo bench --features "test"`?

PR #556, opened 2024-11-05
Refactor BaseFold hashing benchmark.

yczhangsjtu:
Extract a small independent change from #294


mpcs/benches/hashing.rs
```diff
@@ -5,28 +5,19 @@ use goldilocks::Goldilocks;
 use mpcs::util::hash::{Digest, hash_two_digests};
 use poseidon::poseidon_hash::PoseidonHash;
 
+fn random_ceno_goldy() -> Goldilocks {
+    Goldilocks::random(&mut test_rng())
```
hero78119, line 9: A better style probably been changed to call `test_rng()` once and reused it instead of instantiate & drop immediately for every field element, which is very costly.

The good thing is it's put in bench "setup", at least not distorted benchmark results.

Not directly related to this PR, as it already there for long time. overall LGTM!

hero78119 (approved):

PR #560, opened 2024-11-06
BaseFold: Extract Merkle tree inner into a new struct.

yczhangsjtu:
Extract small PR from #294

Currently the inner (without the leafs) of the Merkle tree is represented by a `Vec<Vec<Digest>>`. Sometimes (for optimization reasons, e.g., when the leaf values should stay in the original variable and we don't want to clone it into the Merkle tree) this inner part needs to be accessed and stored separately from the leaves. So we encapsulate it in a `struct MerkleTreeDigests` for more readability, implementing common methods on them, and allow enforcing particular constraints over the `Vec<Vec<Digest>>`.

This refactor provides new APIs for `MerkleTreeDigests` and make the corresponding `MerkleTree` APIs a thin wrapper around the APIs for `MerkleTreeDigests`. There are some new APIs and the behavior of existing APIs stay the same.


mpcs/src/util/merkle_tree.rs
```diff
@@ -23,36 +23,89 @@ use ark_std::{end_timer, start_timer};
...
+        leaf_group_index: usize, // Two leaves make a group.
+    ) -> MerklePathWithoutLeafOrRoot<E> {
+        assert!(leaf_group_index < self.bottom_size());
+        MerklePathWithoutLeafOrRoot::<E>::new(
+            self.inner
+                .iter()
+                .take(self.height() - 1)
+                .enumerate()
+                .map(|(index, layer)| {
+                    Digest::<E::BaseField>(layer[(leaf_group_index >> index) ^ 1].clone().0)
```
matthiasgoergens, line 87: This bit manipulation would deserve an explanatory comment somewhere.

(But it looks like you are only moving that stuff around in this PR.)

mpcs/src/util/merkle_tree.rs
```diff
@@ -23,36 +23,89 @@ use ark_std::{end_timer, start_timer};
...
-    pub fn compute_inner_ext(leaves: &[E]) -> Vec<Vec<Digest<E::BaseField>>> {
-        merkelize_ext::<E>(&[leaves])
+    pub fn root(&self) -> Digest<E::BaseField> {
+        self.inner.last().unwrap()[0].clone()
     }
 
-    pub fn root_from_inner(inner: &[Vec<Digest<E::BaseField>>]) -> Digest<E::BaseField> {
-        inner.last().unwrap()[0].clone()
+    pub fn root_ref(&self) -> &Digest<E::BaseField> {
+        &self.inner.last().unwrap()[0]
```
matthiasgoergens, line 65: It looks like we require that `inner` not be empty.  I suggest [we encode that in the type](https://docs.rs/nonempty/latest/nonempty/), instead of going with `Vec` which is happy being empty.

mpcs/src/util/merkle_tree.rs
```diff
@@ -23,36 +23,89 @@ use ark_std::{end_timer, start_timer};
...
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
-#[serde(bound(serialize = "E: Serialize", deserialize = "E: DeserializeOwned"))]
-pub struct MerkleTree<E: ExtensionField>
+#[serde(bound(deserialize = "E: DeserializeOwned"))]
+pub struct MerkleTreeDigests<E: ExtensionField>
 where
     E::BaseField: Serialize + DeserializeOwned,
 {
     inner: Vec<Vec<Digest<E::BaseField>>>,
```
matthiasgoergens, line 31: You might want to leave a doc comment on the struct explaining how this vector encodes a tree?

mpcs/src/util/merkle_tree.rs
```diff
@@ -23,36 +23,89 @@ use ark_std::{end_timer, start_timer};
...
 }
 
-impl<E: ExtensionField> MerkleTree<E>
+impl<E: ExtensionField> MerkleTreeDigests<E>
 where
     E::BaseField: Serialize + DeserializeOwned,
 {
-    pub fn compute_inner(leaves: &FieldType<E>) -> Vec<Vec<Digest<E::BaseField>>> {
-        merkelize::<E>(&[leaves])
+    pub fn from_leaves(leaves: &FieldType<E>) -> Self {
```
matthiasgoergens, line 38: Consider implementing `From` instances instead?  (Not completely sure.)

mpcs/src/util/merkle_tree.rs
```diff
@@ -23,36 +23,89 @@ use ark_std::{end_timer, start_timer};
...
+        self.inner.len()
+    }
+
+    pub fn bottom_size(&self) -> usize {
+        self.inner.first().unwrap().len()
+    }
+
+    pub fn merkle_path_without_leaf_sibling_or_root(
+        &self,
+        leaf_group_index: usize, // Two leaves make a group.
```
matthiasgoergens, line 78: Perhaps stick that in a doc comment instead, so it shows up when people try to use this function?

matthiasgoergens (commented):
Please see comments.

Btw, you don't need to give the types everywhere, Rust can figure much of that out on its own.  Here's an example of two places, but there are more:

```diff
diff --git a/mpcs/src/basefold/commit_phase.rs b/mpcs/src/basefold/commit_phase.rs
index 992745bb..5df3aa19 100644
--- a/mpcs/src/basefold/commit_phase.rs
+++ b/mpcs/src/basefold/commit_phase.rs
@@ -98,8 +98,7 @@ where
         );
 
         if i > 0 {
-            let running_tree =
-                MerkleTree::<E>::new(running_tree_inner, FieldType::Ext(running_oracle));
+            let running_tree = MerkleTree::new(running_tree_inner, FieldType::Ext(running_oracle));
             trees.push(running_tree);
         }
 
@@ -114,7 +113,7 @@ where
             // Then the oracle will be used to fold to the next oracle in the next
             // round. After that, this oracle is free to be moved to build the
             // complete Merkle tree.
-            running_tree_inner = MerkleTreeDigests::<E>::from_leaves_ext(&new_running_oracle);
+            running_tree_inner = MerkleTreeDigests::from_leaves_ext(&new_running_oracle);
             let running_root = running_tree_inner.root();
             write_digest_to_transcript(&running_root, transcript);
             roots.push(running_root.clone());
diff --git a/mpcs/src/util/merkle_tree.rs b/mpcs/src/util/merkle_tree.rs
index 53568af8..81af42bd 100644
--- a/mpcs/src/util/merkle_tree.rs
+++ b/mpcs/src/util/merkle_tree.rs
@@ -83,9 +83,7 @@ where
                 .iter()
                 .take(self.height() - 1)
                 .enumerate()
-                .map(|(index, layer)| {
-                    Digest::<E::BaseField>(layer[(leaf_group_index >> index) ^ 1].clone().0)
-                })
+                .map(|(index, layer)| Digest(layer[(leaf_group_index >> index) ^ 1].clone().0))
                 .collect(),
         )
     }
```

(Sometimes adding extra redundant type information helps readability, but here it just seems to add verbosity.)

matthiasgoergens:
Could you please add in the PR description _why_ we are doing this?

mpcs/src/util/merkle_tree.rs
yczhangsjtu: Yes, that would be better. But there are more to constrain than this, e.g., the size of the vectors should be 1, 2, 4, 8, and so on. I don't know how to encode all these in the type. Currently, it's just relying on `inner` being private, and is never changed once the Merkle tree is built.

mpcs/src/util/merkle_tree.rs
yczhangsjtu: I'm not sure, either. Computing the Merkle tree is an expensive procedure, so we probably want it to be always invoked explicitly.

mpcs/src/util/merkle_tree.rs
yczhangsjtu: Added comments for this.

mpcs/src/util/merkle_tree.rs
matthiasgoergens: Rust never invokes `from` implicitly.

yczhangsjtu: No, not in the language. I just have the impression (which may be wrong) that `.into()` gives people the hint that this is a cheap type conversion, invoking some wrapper functions, without any expensive stuff.

hero78119:
Will close it as now all switch to p3 mmcs impl :)

Issue #537, opened 2024-11-02
"Some Improvements for the PIOP for ZeroCheck" paper follow up


hero78119:
Quick hacky try on  [Some Improvements for the PIOP for ZeroCheck](https://eprint.iacr.org/2024/108.pdf) with 2 tricks
- 3.1 Sending less data => skip send univariate u(1) as it can be derived from verifier
- 3.2 Slight modification of the Protocol => exploit "eq" for its special form. In general, it can be derived as "common factor" into the protocol, so the overall degree will be minus one

for above 2 tricks can be easily verified as "low-hanging fruit"

This is a workaround commit to evaluate potiential gain on Ceno zkVM end-to-end test https://github.com/scroll-tech/ceno/commit/02983a3d57b5e7e85acdae6e18ba7feba02117d5, idea of these changes are to simulate
- exclude one poly from sumcheck to simulate `eq` exploited
> eq still need to perform `fix_variable` alike operation to derive "common factor"

- send 1 less point u(1)

### Preliminary Benchmark result
- 2^20 add instances ~= -7.07% throughput improvement
- fabonacci assemply toy example < 10% throughput improvement
- fabonacci e2e => TBD

### (Preliminary) Conclusion
Overall improvement percentage depends on how much proportion sumcheck protocol involve into critical path. We can expect it benefit on tower sumcheck more, as it can reduce logup GKR sumcheck degree from 3 ((eq * f1 * g2)) -> 2 (f1 * g2). Tricks 3.1, 3.2 looks like probably improve around 5~10%, which is much less than my expectation (like > 30% ) It will be worth to dive in more, and probably further breakdown by flamegraph and see what could be the bottleneck.

hero78119:
design addressed in #799

PR #568, opened 2024-11-06
[experiment] more tracing span to identify mpcs latency breakdown on flamegraph

hero78119:
This PR just add tracing span to measure the mpcs break down and visualized in flamegraph

Also attachmented framegraph (its svg file and please download it local then open in browser which allow iteractive navigating)
flamegraph

The bottleneck part probably not on sumcheck, and instead it's from `compute_inner_ext` and `root_from_inner` operation
https://github.com/scroll-tech/ceno/blob/a756f466f7086bf5b204bff7afbf175c648050da/mpcs/src/basefold/commit_phase.rs#L458-L461


hero78119:
related to https://github.com/scroll-tech/ceno/pull/568

hero78119:
More added on: the latency breakdown
- 75% time on mpcs
- 25% time on opcode/table sumckeck proof

So optimize mpcs at current stage will be more huge gain.


In another follow up https://github.com/scroll-tech/ceno/issues/537#issuecomment-2453364815 the improvement only <15% overall is because it only affect opcode proof sumcheck part. mpcs got another sumcheck implementation and we plan to merge it in issue.

In overall, as mpcs occupied more time (til 7 Nov '24), any improvement on mpcs will have better impact on overall throuput.

hero78119:
Closed in favor of new mpcs implementation.

PR #526, opened 2024-11-01
Feat: mock prover for zkvm

kunxian-xia:
This PR aims to improve current mock prover to find out constraint errors in the zkvm setting.

- [x] lookup errors.
- [x] read / write not equal for global state 
- [x] read / write not equal for registers
- [x] read / write not equal for memory

Issue #485, opened 2024-10-28
Add comment about what's supposed to be not serialised


matthiasgoergens:
> In almost all the case, unifying source of truth and less error prune make much sense!
> But in this case I will say not.
> 
> We expect `ConstraintSystem`
> https://github.com/scroll-tech/ceno/blob/79bb73bb89c7a1f503b0838711acaa4d7e7679f2/ceno_zkvm/src/circuit_builder.rs#L82
> to be serialized into proving/verifying key. So, we expect some fields scope should be move to `cfg[test]` or just under some feature flag toggle on
> All the "xxxxx_map" fields is one of the example. Those debug info are 
> 1. from prover perspective, only push those name strings in non-release build. Those variable debug name shouldn't be in release build nor in binary.
> 2. from verifier perspective, neither debug nor release need this information, so they should have them "serde skip".  
> 
> With above 2 points, I will say this PR change is unnessesary, because source of truth shouldn't reverse release info to depends on debug info

@hero78119 Could you please add that reasoning as a comment in the code?  And perhaps make it so that `witin_namespace_map` actually only show up in debug mode?  (Or otherwise leave a TODO-comment to that effect?)

Btw, I don't think we should spend effort removing variable names from a release build.  It's better to leave them around for debugging, because things can go wrong with the release build, too.

(That's all assuming that constructing the circuit is far from the most expensive thing we are doing.  I expect (recursive) proving to take the majority of the runtime.)

_Originally posted by @matthiasgoergens in https://github.com/scroll-tech/ceno/issues/465#issuecomment-2441251845_

PR #569, opened 2024-11-07
Feat: allow constraint_system to take record as input instead of record's rlc

kunxian-xia:
This PR is extracted from #526. 

In practice, when we use mock prover to find out constraint errors like r/w mismatch, we want to know the exact value of these records. However, currently we only store the RLC of record in the `ConstraintSystem`. Therefore in this PR we change the interface of `ConstraintSystem`.

### DIFF
--- a/ceno_zkvm/Cargo.toml
+++ b/ceno_zkvm/Cargo.toml
@@ -26,19 +26,19 @@ strum_macros.workspace = true
 tracing.workspace = true
 tracing-flame.workspace = true
 tracing-subscriber.workspace = true
-serde_json.workspace = true
 
 clap = { version = "4.5", features = ["derive"] }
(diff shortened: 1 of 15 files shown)


hero78119 (approved):
LGTM!

A side effect is it will increase verifier key size as we put more debug info, where is a general question we already have and probably address together in https://github.com/scroll-tech/ceno/issues/485

kunxian-xia:
> A side effect

We can skip these debugging info when serialize.

hero78119:
> > A side effect
> 
> We can skip these debugging info when serialize.

its an option. Another is we should isolate them with specific feature toggle, so debug purpose expressed clearly
https://github.com/scroll-tech/ceno/pull/571#discussion_r1832990748

PR #572, opened 2024-11-07
[experiment] Extend profiling using `tracing`

mcalancea:
### Initial comment (Nov 7th)
This PR aims to improve profiling using spans from the `tracing` library. Some observations up to this point:
- This is how a flamegraph looks for me on this branch (open in browser for interaction): 
tracing-flamechart

- When spawning threads, we need to be explicit about recording the work performed by the spawn thread within the current span in the parent thread. See first commit, where before the change the `sumcheck::prove_rounds` span was a sibling of the `TOPLEVEL` span in the flamechart, instead of being contained by it.
- Span names must be constants. To distinguish between different spans in the same loop we need to use fields: 
           ` let span = entered_span!("commit to iteration", circuit_name=circuit_name);`
Unfortunately the flamegraph doesn't include fields in labels. Currently, in this situation it's better to filter after the parent span (in this case `commit_to_traces` and look at the spans textually. It looks like this:
Screenshot from 2024-11-07 16-32-52 Filtering by _span name_ for manual inspections is simple as:
`.with(EnvFilter::new("[commit_to_traces]=info"))`
I can make it an option.
- Filtering spans over fields and field values would be quite powerful (for example for easy granularity flags).. if it worked like the docs said, but I am currently of the opinion that it does not :). 

I can apply these improvements in more places if they sound good. 

### Later edit (Nov 8th)
I've applied the changes in several places. I've added comments about pitfalls I'm aware of.

I believe the `invisible gap` symptom (as pointed out by @hero78119 in the comments) is (at least in some cases) a bug in the `tracing-flame` crate. 
tracing-flamechart
As can be seen in the code, the `wrap_batch` span is a direct wrapper of its child span, but it introduces a huge `invisible gap`. I believe it's a bug in how `tracing-flame` generates the timings for in the `.folded` file because:
- Inspecting the text logs, the duration of `wrap_batch` is only fractionally longer than the child, as one would expect.
- The flamegraph suggests a total runtime which is over twice the real time.

If I get the time I will try to minimize and post an issue upstream. 

I would've liked to include an easy way to filter away small spans for easier inspection of the logs, but it's a bit complicated. Maybe let's merge this first and let me know about your experience with it.

### DIFF
--- a/ceno_zkvm/examples/riscv_opcodes.rs
+++ b/ceno_zkvm/examples/riscv_opcodes.rs
@@ -25,8 +25,10 @@ use ff_ext::ff::Field;
 use goldilocks::GoldilocksExt2;
 use itertools::Itertools;
 use mpcs::{Basefold, BasefoldRSParams, PolynomialCommitmentScheme};
+use sumcheck::{entered_span, exit_span};
 use tracing_flame::FlameLayer;
 use tracing_subscriber::{EnvFilter, Registry, fmt, layer::SubscriberExt};
+use tracing_subscriber::fmt::format::FmtSpan;
 use transcript::Transcript;
 
 const PROGRAM_SIZE: usize = 16;
@@ -96,6 +98,7 @@ fn main() {
         .with(
             fmt::layer()
                 .compact()
+                .with_span_events(FmtSpan::CLOSE)
                 .with_thread_ids(false)
                 .with_thread_names(false),
         )
@@ -103,6 +106,9 @@ fn main() {
         .with(flame_layer.with_threads_collapsed(true));
     tracing::subscriber::set_global_default(subscriber).unwrap();
 
+    let top_level = entered_span!("TOPLEVEL");
+
+    let keygen = entered_span!("KEYGEN");
     // keygen
     let pcs_param = Pcs::setup(1 << MAX_NUM_VARIABLES).expect("Basefold PCS setup");
     let (pp, vp) = Pcs::trim(&pcs_param, 1 << MAX_NUM_VARIABLES).expect("Basefold trim");
@@ -138,6 +144,7 @@ fn main() {
         .expect("keygen failed");
     let vk = pk.get_vk();
 
+    exit_span!(keygen);
     // proving
     let prover = ZKVMProver::new(pk);
     let verifier = ZKVMVerifier::new(vk);
@@ -277,9 +284,12 @@ fn main() {
         let timer = Instant::now();
 
         let transcript = Transcript::new(b"riscv");
+
+        let span = entered_span!("CREATE_PROOF");
         let mut zkvm_proof = prover
             .create_proof(zkvm_witness, pi, transcript)
             .expect("create_proof failed");
+        exit_span!(span);
 
         println!(
             "riscv_opcodes::create_proof, instance_num_vars = {}, time = {}",
@@ -329,4 +339,5 @@ fn main() {
             }
         };
     }
+    exit_span!(top_level);
 }
--- a/ceno_zkvm/examples/riscv_opcodes.rs
+++ b/ceno_zkvm/examples/riscv_opcodes.rs
@@ -94,15 +94,17 @@ fn main() {
             .collect(),
     );
     let (flame_layer, _guard) = FlameLayer::with_file("./tracing.folded").unwrap();
+    let mut fmt_layer =             fmt::layer()
+    .compact()
+    .with_span_events(FmtSpan::CLOSE)
+    .with_thread_ids(false)
+    .with_thread_names(false);
+    fmt_layer.set_ansi(false);
+
     let subscriber = Registry::default()
         .with(
-            fmt::layer()
-                .compact()
-                .with_span_events(FmtSpan::CLOSE)
-                .with_thread_ids(false)
-                .with_thread_names(false),
+            fmt_layer
         )
-        //.with(EnvFilter::from_default_env())
         .with(EnvFilter::new("info"))
         .with(flame_layer.with_threads_collapsed(true));
     tracing::subscriber::set_global_default(subscriber).unwrap();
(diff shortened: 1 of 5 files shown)


hero78119:
I was bothered it for a while especially `recording the work performed by the spawn thread` so it marked and stacktrace collapsed nice, thanks for clarify it, it save my ass!

hero78119:
Another thing quite mystery in flamegraph for me is about `invisible gap`, for example 
image
point to this code
https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/prover_v2.rs#L359-L362

As I can tell this gap should not exist, because there should be neglible workload during this time.
So I am guessing few probablity
- this gap imply some context switch (?)
- this gap is a false signal and it's due to collapse stacktrace from multi-thread so it's not precisely
- there are really some hidden workload (I think less possible...)

But still no clear idea as haven't look into the problem detail.

matthiasgoergens:
Thanks for the nice write-up in the PR description!  We should do more of that kind of thing to give context for reviews.

ceno_zkvm/examples/riscv_opcodes.rs
```diff
@@ -277,9 +284,12 @@ fn main() {
         let timer = Instant::now();
 
         let transcript = Transcript::new(b"riscv");
+
+        let span = entered_span!("CREATE_PROOF");
```
matthiasgoergens, line 288: Perhaps something more descriptive?  (Just like you did for the outer spans?)

```suggestion
        let create_proof = entered_span!("CREATE_PROOF");
```

ceno_zkvm/examples/riscv_opcodes.rs
```diff
@@ -329,4 +339,5 @@ fn main() {
             }
         };
     }
+    exit_span!(top_level);
```
matthiasgoergens, line 342: Btw, what happens if you forget to exit a span, or exit them in the wrong order?

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -100,16 +102,18 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
                         PCS::batch_commit_and_write(&self.pk.pp, &witness, &mut transcript)
                             .map_err(ZKVMError::PCSError)?,
                     );
-                    tracing::info!(
-                        "commit to {} traces took {:?}",
-                        circuit_name,
-                        commit_dur.elapsed()
-                    );
+                    // tracing::info!(
```
matthiasgoergens, line 105: When the PR is ready, please remove commented out code, we can always get it back from the git history.

(Commented out code is fine for a draft.)

matthiasgoergens (commented):
Looking forward to this arriving in master, when it's ready!

ceno_zkvm/examples/riscv_opcodes.rs
mcalancea: > Btw, what happens if you forget to exit a span, or exit them in the wrong order?

Right now it seems to misrecord the offending spans silently :). I think in general it depends on the implementation of the subscriber/layer that processes them. A custom one could panic instead. 

> Might be useful to have a macro that works like this:

I understand the benefit, but for me the extra nesting makes the code less readable. Also the `enter/exit` pattern existed already in the codebase.

ceno_zkvm/examples/riscv_opcodes.rs
mcalancea: Renamed in most cases :+1

mcalancea:
> Another thing quite mystery in flamegraph for me is about invisible gap, for example

@hero78119 Yeah so sometimes I think it's due to having code not covered by spans, but other times it's just wrong. I updated the PR description with a case which I think points to a bug in the flame layer.

ceno_zkvm/examples/riscv_opcodes.rs
naure: Or they could be a check a the end of the program that all spans were closed. If that's possible to know.

naure: If you forget the `exit_span` you get a word from the compiler (should be refused by CI):

`warning: unused variable: my_span`

and if you forget the `let my_span`:

`warning: unused `EnteredSpan` that must be used`

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -1025,7 +1036,7 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
             .collect_vec();
         // TODO implement mechanism to skip commitment
 
-        let span = entered_span!("pcs_opening");
+        let pcs_opening = entered_span!("pcs_opening");
```
naure, line 1039: Maybe keep the word `span` in names to distinguish them from the logic.

naure (approved):

hero78119 (commented):
All changes LGTM! Its greatly improve the way we using transcripts.


Btw one thing just to double check :) may I know what the raw command you have to generate tracing logs? I noticed "commit_to_traces", "pcs_opening" should occupied quite of amount in end-to-end latency, but in flamegraph it's just be a small fraction.

mcalancea:
Running
`RUST_LOG="info" cargo run  --release --example riscv_opcodes -- --start 2 --end 5 > logs.txt`
to see all spans (unfortunately kind of verbose).

To me it looks like `commit_to_traces` really takes a small fraction of time in this case: 90ms vs. 3.15s. 
```
2024-11-12T07:14:14.041104Z  INFO TOPLEVEL:ZKVM_create_proof:commit_to_traces: ceno_zkvm::scheme::prover: close time.busy=90.1ms time.idle=1.85µs
2024-11-12T07:14:16.608592Z  INFO TOPLEVEL: riscv_opcodes: close time.busy=3.15s time.idle=2.77µs
```
Note that:
- The span data in the logs should not be interpreted as "uniform samples" as in other profiling methods, but intervals with different lengths (`time.busy` gives the length). So, for example, a span name can appear many times in the file, but it just means many instances of it were executed; the total time may still be small.
- The time is given varyingly in seconds, microseconds or even nanoseconds according to its magnitude.

hero78119:
> > Another thing quite mystery in flamegraph for me is about invisible gap, for example
> 
> @hero78119 Yeah so sometimes I think it's due to having code not covered by spans, but other times it's just wrong. I updated the PR description with a case which I think points to a bug in the flame layer.

Hi @mcalancea when deal with other issue in codebase, I also observed the similar gap even flamegraph already removed , I think the answer for invisible gap is probably due to "Drop" behaviours after function finish, which indicates we have some heavy objects. So probably its not the bug of flamegraph upstream

PR #576, opened 2024-11-08
Remove unimplmented and unused functions

matthiasgoergens:


hero78119:
LGTM!
Just a suggestion: why clippy not capture this :) ?  is it because there lack of some setting?

matthiasgoergens:
> LGTM! Just a suggestion: why clippy not capture this :) ? is it because there lack of some setting?

It's because clippy can't know whether some other project uses our `pub` functions.  So all `pub` functions are considered used by default.

(I wonder if there's a setting we can use to tell clippy that not all `pub` functions should automatically be counted as used.)

hero78119 (approved):

PR #577, opened 2024-11-08
dummy-ecall-power: Let ecall do anything

naure:
A variant of #369 with all-powerful ecall instead of NOP.

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
+++ b/ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
@@ -1,6 +1,6 @@
 use std::marker::PhantomData;
 
-use ceno_emul::{InsnCategory, InsnCodes, InsnFormat, InsnKind, StepRecord};
+use ceno_emul::{InsnCategory, InsnFormat, InsnKind, StepRecord};
 use ff_ext::ExtensionField;
 
 use super::super::{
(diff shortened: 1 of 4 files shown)


ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
```diff
@@ -67,21 +67,23 @@ impl<E: ExtensionField> DummyConfig<E> {
         circuit_builder: &mut CircuitBuilder<E>,
         codes: InsnCodes,
     ) -> Result<Self, ZKVMError> {
-        let (with_rs1, with_rs2, with_rd) = match (codes.format, codes.kind) {
-            // ECALL reads its syscall_id, then do nothing.
-            (_, InsnKind::EANY) => (true, false, false),
+        // ECALL can do everything.
+        let is_ecall = matches!(codes.kind, InsnKind::EANY);
+        let (with_rs1, with_rs2, with_rd) = match codes.format {
+            _ if is_ecall => (true, true, true),
```
kunxian-xia, line 73: I felt like we should let the caller to decide how many registers they want to access.

kunxian-xia: For example, `ecall/commit` only needs to access X5 and X10 (set `with_rs1 = true, with_rs2 = true`). While complex ecall may use X5 / X10 / X11 by setting `with_rs1 = true, with_rs2 = true, with_rd = true`.

naure: I made it configurable so it is easy to write restricted ecall circuit. The default option is everything at once.

naure:
Clearly this is what we want so I include it in #369 .

ceno_zkvm/src/instructions/riscv/dummy/dummy_circuit.rs
```diff
@@ -34,7 +34,37 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for DummyInstruction<E
...
+            InsnFormat::S => (true, true, false),
+            InsnFormat::B => (true, true, false),
+            InsnFormat::U => (false, false, true),
+            InsnFormat::J => (false, false, true),
+        };
+        let with_mem_write = matches!(codes.category, InsnCategory::Store) || is_ecall;
+        let with_mem_read = matches!(codes.category, InsnCategory::Load);
+        let branching = matches!(codes.category, InsnCategory::Branch)
+            || matches!(codes.kind, InsnKind::JAL | InsnKind::JALR)
+            || is_ecall;
```
kunxian-xia, line 56: Most of `Ecall` instructions follow the default branching rule `pc' = pc + 4` except `Ecall/Halt`.

naure: That’s ok, who can do more, can do less. I do imagine that is how I will do the complex ecalls: jump to handler code.

PR #583, opened 2024-11-11
Fix: assignment of Memory Finalize

kunxian-xia:
This is just a temporary fix of the memory errors caused by wrong assignment to memory init / finalize table.


ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -1011,7 +1011,7 @@ Hints:
                         let mut records = vec![];
                         for (row, record_rlc) in write_rlc_records.into_iter().enumerate() {
                             // TODO: report error
-                            assert_eq!(writes.insert(record_rlc), true);
+                            writes.insert(record_rlc);
```
naure, line 1014: This is supposed to later collect and return the errors (not assert), right?

ceno_zkvm/src/tables/ram/ram_impl.rs
```diff
@@ -396,46 +396,41 @@ impl<DVRAM: DynVolatileRamTable + Send + Sync + Clone> DynVolatileRamTableConfig
...
             .with_min_len(MIN_PAR_SIZE)
-            .zip(final_mem.into_par_iter())
-            .for_each(|(row, rec)| {
-                set_val!(row, self.addr, rec.addr as u64);
-                if self.final_v.len() == 1 {
-                    // Assign value directly.
-                    set_val!(row, self.final_v[0], rec.value as u64);
+            .enumerate()
+            .for_each(|(i, row)| {
+                let addr = DVRAM::addr(i);
```
naure, line 412: I think this should work the other way around -- the tables should take the set of addresses as input rather than impose it and try to reconcile it all that way.

But this PR is progress.

naure (approved):

ceno_zkvm/src/scheme/mock_prover.rs
kunxian-xia: This assertion should not be removed. I've restored it.

Issue #566, opened 2024-11-06
Make the platform dynamic


naure:
Currently, platform details (e.g. special memory addresses) are constants, and guest programs must target it.

It’s possible to upgrade these constants to parameters chosen by guest programs. These parameters can be given in ELF files and/or CLI arguments.

To do it, replace all usages of the `CENO_PLATFORM` constant into a `platform: Platform` argument, like for example `VMState.platform` does. It might be harder than it sounds; did not try yet.

_Refs: [platform.rs](https://github.com/scroll-tech/ceno/blob/master/ceno_emul/src/platform.rs), [linker script](https://github.com/scroll-tech/ceno/blob/master/ceno_rt/ceno_link.x), etc._

_Related: #215, #378, #476._

naure:
A number of these dependencies to Platform are being removed in #368.

A notable exception is `StaticMemTable::len()` and `PubIOTable::len()` which are currently hard-coded.

mcalancea:
Yep.

PR #587, opened 2024-11-13
[experiment] Remove PROGRAM_SIZE from type

mcalancea:
_naure:_
- Issue #566.
- Start a move towards circuit **parameters as values**, instead of as type.


ceno_zkvm/src/structs.rs
```diff
@@ -152,10 +152,10 @@ impl<E: ExtensionField> ZKVMConstraintSystem<E> {
         config
     }
 
-    pub fn register_table_circuit<TC: TableCircuit<E>>(&mut self) -> TC::TableConfig {
+    pub fn register_table_circuit<TC: TableCircuit<E>>(&mut self, tc: TC) -> TC::TableConfig {
```
naure, line 155: There could be a convenience method (original name) that uses Default (if implemented) and a new one with the parameter.

```rust
impl<TC> …
where TC::TableConfig: Default {
    pub fn register_table_circuit(…) {
        let tc = Default::default();
…
}

impl<TC> … {
    pub fn register_table_circuit_param(…, tc: TC) {
…
}
```

naure: This way there is no need to change every call.

mcalancea: :+1:  agreed, got a bit carried away there

ceno_zkvm/src/tables/program.rs
```diff
@@ -104,11 +104,22 @@ pub struct ProgramTableConfig {
     mlt: WitIn,
 }
 
-pub struct ProgramTableCircuit<E, const PROGRAM_SIZE: usize>(PhantomData<E>);
+#[derive(Clone, Default)]
```
naure, line 107: Default here is 0 size which is probably not what you want. Just leave it without default to force the use of the `…_param` function?

ceno_zkvm/src/tables/mod.rs
```diff
@@ -18,18 +18,23 @@ pub use program::{InsnRecord, ProgramTableCircuit};
 mod ram;
 pub use ram::*;
 
-pub trait TableCircuit<E: ExtensionField> {
+pub trait TableCircuit<E: ExtensionField>
+where
+    Self: Default,
```
naure, line 23: This Default requirement could move as far out as possible towards the function that actually needs it.

naure (approved):

ceno_zkvm/examples/fibonacci_elf.rs
```diff
@@ -79,15 +78,18 @@ fn main() {
...
     zkvm_cs.register_global_state::<GlobalState>();
 
     let mut zkvm_fixed_traces = ZKVMFixedTraces::default();
 
-    zkvm_fixed_traces.register_table_circuit::<ExampleProgramTableCircuit<E>>(
+    zkvm_fixed_traces.register_table_circuit_param::<ProgramTableCircuit<E>>(
         &zkvm_cs,
         &prog_config,
         vm.program(),
+        ptc_with_size,
```
naure, line 92: I applied the same change to this new file.

PR #596, opened 2024-11-18
Feat/Opcode Circuits for DIV, REM, and REMU opcodes

bgillesp:
This PR extends the existing `DIVU` opcode implementation to expose the remainder as an output, and to handle the (more complicated) signed case.  See the comments in `div.rs` for details, but in summary:
- The basic design of signed and unsigned `DIV` and `REM` opcode circuits is to represent the division algorithm relation `dividend = quotient * divisor + remainder` using native field elements, and constraining so that either `0 <= remainder < divisor` or `divisor == 0` holds.
- Some nontrivial steps are needed in the signed case to handle the additional complexity of different signs for the divisor and remainder values, as well as to handle signed division overflow where `-2^31 / -1 = 2^31` is out of the representable range for `i32` values.
- In either case, either the `quotient` or the `remainder` value is exposed as the output register `rd`.
- In theory with the current implementation, consecutive calls to `DIV` and `REM` or `DIVU` and `REMU` could be batched into a single circuit since both of the outputs are represented; however, this is not yet implemented.

Currently the circuit implementation and documentation are finished and ready for review.  Unit tests still need to be finished, but it would be great to get some eyes on the circuit design now.


kunxian-xia:
@bgillesp when will this pr be ready for review?

bgillesp:
Unit tests still need to be added for the new opcode circuits, but everything else is in shape for an initial review -- will change the PR out of Draft now.

ceno_zkvm/src/gadgets/is_lt.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
 
 use super::SignedExtendConfig;
 
+// TODO rename to AssertLtConfig (LT -> Lt) to fit naming conventions
```
matthiasgoergens, line 21: See https://github.com/scroll-tech/ceno/pull/731

matthiasgoergens (commented):
A few comments and suggestions so far.

ceno_zkvm/src/instructions/riscv/div.rs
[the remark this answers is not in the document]
kunxian-xia: #733 is merged.

matthiasgoergens:
@bgillesp I made a version of this PR that has an easier to read diff and left some comments there.  They apply here as well.  See https://github.com/scroll-tech/ceno/pull/734

hero78119:
Attach DIVU circuit statistics,

First row master branch, second row this PR
```
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| opcode_name   | num_instances | lookups | reads | witnesses | writes | 0_expr_deg | 0_expr_sumcheck_deg |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| DIVU          | 0             | 21      | 4     | 37        | 4      | [1: 7]     | [2: 8]              | 
| DIVU          | 0             | 13      | 4     | 29        | 4      | [1: 5]     | [2: 5]              | 
```

matthiasgoergens:
> Attach DIVU circuit statistics 🔥
> 
> First row master branch, second row this PR
> 
> ```
> +---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
> | opcode_name   | num_instances | lookups | reads | witnesses | writes | 0_expr_deg | 0_expr_sumcheck_deg |
> +---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
> | DIVU          | 0             | 21      | 4     | 37        | 4      | [1: 7]     | [2: 8]              | 
> | DIVU          | 0             | 13      | 4     | 29        | 4      | [1: 5]     | [2: 5]              | 
> ```

Keep in mind that the PR has at least one bug (and that might affect the stats)?

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -0,0 +1,599 @@
...
+
+            InsnKind::DIV | InsnKind::REM => {
+                let dividend_signed: Signed<E> =
+                    Signed::construct_circuit(cb, || "dividend_signed", &dividend)?;
+                let divisor_signed: Signed<E> =
+                    Signed::construct_circuit(cb, || "divisor_signed", &divisor)?;
+                let quotient_signed: Signed<E> =
+                    Signed::construct_circuit(cb, || "quotient_signed", &quotient)?;
+                let remainder_signed: Signed<E> =
+                    Signed::construct_circuit(cb, || "remainder_signed", &quotient)?;
```
KimiWu123, line 192: This should be `remainder` as Matt mentioned.

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -0,0 +1,599 @@
...
+        // dividend = quotient * divisor + remainder
+        let dividend = step.rs1().unwrap().value;
+        let divisor = step.rs2().unwrap().value;
+
+        let dividend_v = Value::new_unchecked(dividend);
+        let divisor_v = Value::new_unchecked(divisor);
+
+        config
+            .dividend
+            .assign_limbs(instance, dividend_v.as_u16_limbs());
```
KimiWu123, line 364: You can use `assign_value` instead.

ceno_zkvm/src/gadgets/signed.rs
```diff
@@ -0,0 +1,57 @@
...
+
+            Ok(Self { is_negative, val })
+        })
+    }
+
+    pub fn assign_instance(
+        &self,
+        instance: &mut [MaybeUninit<E::BaseField>],
+        lkm: &mut LkMultiplicity,
+        val: &Value<u32>,
```
KimiWu123, line 42: Will it be better to use `i32`? Callers could assign `i32::MAX + 1`, and this could cause unexpeced result If the type is`u32`.

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -0,0 +1,599 @@
...
+        let is_divisor_zero =
+            IsZeroConfig::construct_circuit(cb, || "is_divisor_zero", divisor.value())?;
+
+        // For zero division, quotient must be the "all ones" register for both
+        // unsigned and signed cases, representing 2^32-1 and -1 respectively.
+        cb.condition_require_equal(
+            || "quotient_zero_division",
+            is_divisor_zero.expr(),
+            quotient.value(),
+            ((1u64 << UInt::<E>::TOTAL_BITS) - 1).into(),
```
KimiWu123, line 296: The same here, can use `u32::MAX` directly.

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -0,0 +1,599 @@
...
+                    cb,
+                    || "is_dividend_max_negative",
+                    dividend.value(),
+                    (1u64 << (BIT_WIDTH - 1)).into(),
+                )?;
+                let is_divisor_minus_one = IsEqualConfig::construct_circuit(
+                    cb,
+                    || "is_divisor_minus_one",
+                    divisor.value(),
+                    ((1u64 << BIT_WIDTH) - 1).into(),
```
KimiWu123, line 215: This curcuit is limited to 32-bits registers and Goldiliock field so I think we can just use `u32::MAX` here.

ceno_zkvm/src/gadgets/signed.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl<E: ExtensionField> Signed<E> {
+    pub fn construct_circuit<NR: Into<String> + Display + Clone, N: FnOnce() -> NR>(
+        cb: &mut CircuitBuilder<E>,
+        name_fn: N,
+        unsigned_val: &UInt<E>,
+    ) -> Result<Self, ZKVMError> {
+        cb.namespace(name_fn, |cb| {
+            let is_negative = unsigned_val.is_negative(cb)?;
+            let val = unsigned_val.value() - (1u64 << BIT_WIDTH) * is_negative.expr();
```
hero78119, line 32: could reused util function defined on `uint` to deal with the conversion `to_field_expr(is_negative)`

ceno_zkvm/src/instructions/riscv/div.rs
hero78119, line 192: `&remainder` This could be captured once unittest added

hero78119 (commented):
Epic job! thanks for the implementation and clear documentation.
Just do a quick overview and will proceed arithmetics constrain review 2nd round

ceno_zkvm/src/gadgets/signed_ext.rs
```diff
@@ -9,11 +9,16 @@ use crate::{
 use ff_ext::ExtensionField;
 use std::{marker::PhantomData, mem::MaybeUninit};
 
+/// Extract the most significant bit from an expression previously constrained
+/// to an 8- or 16-bit length.
```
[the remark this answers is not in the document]
bgillesp: Sounds good, modified!

ceno_zkvm/src/instructions/riscv/div.rs
bgillesp: Yep, that's fair, the limitation is pretty hard-coded throughout, and this is definitely more readable.  Will make the change.

ceno_zkvm/src/instructions/riscv/div.rs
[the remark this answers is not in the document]
bgillesp: Finished merging in master, hopefully the diff is a little nicer now.

ceno_zkvm/src/gadgets/signed_ext.rs
```diff
@@ -9,11 +9,16 @@ use crate::{
 use ff_ext::ExtensionField;
```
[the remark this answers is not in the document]
bgillesp: Yep, makes sense -- updated!

bgillesp:
Thank you for the generous feedback everyone!  :pray:  I think I've addressed your comments as much as I can currently -- will keep an eye on this space to help move things forward as the review progresses.

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -1,124 +1,348 @@
...
+                    Signed::construct_circuit(cb, || "quotient_signed", &quotient)?;
+                let remainder_signed =
+                    Signed::construct_circuit(cb, || "remainder_signed", &remainder)?;
+
+                // The quotient and remainder can be interpreted as non-positive
+                // values when exactly one of dividend and divisor is negative
+                let neg_div_expr = {
+                    let a_neg = dividend_signed.is_negative.expr();
+                    let b_neg = divisor_signed.is_negative.expr();
+                    &a_neg * (1 - &b_neg) + (1 - &a_neg) * &b_neg
```
KimiWu123, line 202: Really minor one, I like the original style more because It's clear and slightly efficient. It reduces the number of recursion [here](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/expression/monomial.rs#L13). But it depends on you :)
```rust                    
// a_neg * (1 - b_neg) + (1 - a_neg) * b_neg
(a_neg.clone() + b_neg.clone()) - (Expression::<E>::from(2) * a_neg * b_neg)
```

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -128,43 +352,122 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for ArithInstruction<E
...
-        config
-            .remainder_lt
-            .assign_instance(instance, lkm, r.as_u64(), divisor.as_u64())?;
+        config.r_insn.assign_instance(instance, lkm, step)?;
 
         Ok(())
     }
 }
 
+// TODO Tests
```
KimiWu123, line 469: I guess you forgot this `TODO`. I didn't see any testing added.

bgillesp: Not forgotten!  I mentioned that this is still TODO in the PR description -- another member of our team is working on it, but paused yesterday due to the rapid changes from the review.  Should be finished relatively soon, but probably an initial review of the circuit design would still be valuable at this point before we validate with unit tests.

ceno_zkvm/src/instructions/riscv/div.rs
bgillesp: Yeah, I think the fact that it was simplified to monomial form was why I went with the original version, but then the current revision is closer to the underlying logic of the expression, which gives an easily readable meaning for binary inputs.  I don't think I have a strong opinion about it -- any thoughts @matthiasgoergens since the change was your suggestion?

ceno_zkvm/src/gadgets/signed.rs
```diff
@@ -0,0 +1,57 @@
...
+        instance: &mut [MaybeUninit<E::BaseField>],
+        lkm: &mut LkMultiplicity,
+        val: &Value<u32>,
+    ) -> Result<i32, ZKVMError> {
+        self.is_negative.assign_instance(
+            instance,
+            lkm,
+            *val.as_u16_limbs().last().unwrap() as u64,
+        )?;
+        let signed_val = val.as_u32() as i32;
```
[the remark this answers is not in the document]
bgillesp: Just merged this in from master and ported the `i32::from(val)` syntax from the `Signed` implementation in `mul.rs`, so should be all set here.

ceno_zkvm/src/instructions/riscv/div.rs
matthiasgoergens: There are two parts to my suggestion.  One is to remove the manual clones, and the other is to use a simpler algebraic expression.

They are independent of each other.

I don't really have much of an opinion on whether `(a + b - 2 * a * b)` or `a * (1 - b) + (1 - a) * b` is easier to read.  I went with the latter, because someone had made a comment `// a_neg * (1 - b_neg) + (1 - a_neg) * b_neg` to that end; so I assume someone thought that form is easier to read.

Here's my suggestion, if you want to use the original expression:

```rust
// a_neg ^ b_neg
&a_neg + &b_neg - 2 * a_neg * b_neg
```

Btw, when you say 'efficient', do you mean that the distribution function will be called slightly less often; or do you mean that the final resulting expression will be different?

bgillesp: Sure, this seems also like a good way to write it, with the simple comment -- updated to this version in the latest commit.

bgillesp:
Note that while commit `e719763` is showing failed tests, the `cargo make tests` command runs successfully on my local build, so I'm guessing it might be something stalled on the CI end.

matthiasgoergens:
> Note that while commit `e719763` is showing failed tests, the `cargo make tests` command runs successfully on my local build, so I'm guessing it might be something stalled on the CI end.

I clicked the button to kick off a re-run of the tests.  (Is that button available for you?)

ceno_zkvm/src/gadgets/is_lt.rs
matthiasgoergens, line 21:
```suggestion
```

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -1,123 +1,348 @@
...
+                    || "is_divisor_minus_one",
+                    divisor.value(),
+                    (-1i32).into(),
+                )?;
+                let is_signed_overflow = cb.flatten_expr(
+                    || "signed_division_overflow",
+                    is_dividend_max_negative.expr() * is_divisor_minus_one.expr(),
+                )?;
+
+                // For signed division overflow, dividend = -2^31 and divisor
```
matthiasgoergens, line 224: I suspect we actually might be able to get away without any special handling of signed division overflow (and just get the right answer out of the general constraints).

But that's something for me to think about later, nothing that should hold up this PR.

> My suspicion is because we have more range available in Goldilocks, and (i32::MIN as Goldilocks) / (-1) == 1<<31 as Goldilocks, which is the answer we want at the end.

matthiasgoergens (approved):
Thanks for your patience with the review process.  Looking good now.

Please remove the now obsolete TODO about the LT -> Lt rename before merging.

matthiasgoergens:
Is there anything left to do here, or can we merge? @bgillesp Do you have enough rights to the repository to hit the 'Merge when ready' button yourself, or do you need someone else to do it for you?

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -1,123 +1,348 @@
...
+//! `remainder` are non-positive values, and we instead require
+//!
+//! `-|divisor| < remainder <= 0` (5)
+//!
+//! To handle these variations of the remainder inequalities in a uniform
+//! manner, we derive expressions representing the "positively oriented" values
+//! with signs set so that the inequalities are always of the form (3).  Note
+//! that it is not enough to just take absolute values, as this would allow
+//! values with an incorrect sign, e.g. for 10 divided by -6, one could witness
+//! `10 = -6 * 2 + 2` instead of the correct expression `10 = -6 * 1 - 4`.
```
hero78119, line 60: This example need fix

hero78119 (commented):
some review feedback of docs :)

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -1,123 +1,348 @@
...
+                    let b_neg = divisor_signed.is_negative.expr();
+                    // a_neg xor b_neg
+                    &a_neg + &b_neg - 2 * a_neg * b_neg
+                };
+                let negative_division = cb.flatten_expr(|| "neg_division", neg_div_expr)?;
+
+                // Check for signed division overflow: i32::MIN / -1
+                let is_dividend_max_negative = IsEqualConfig::construct_circuit(
+                    cb,
+                    || "is_dividend_max_negative",
```
hero78119, line 209: is_dividend_`min`_negative to match with i32::MIN syntatically

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -1,123 +1,348 @@
...
+                    dividend.value(),
+                    i32::MIN.into(),
+                )?;
+                let is_divisor_minus_one = IsEqualConfig::construct_circuit(
+                    cb,
+                    || "is_divisor_minus_one",
```
hero78119, line 215: not sure syntax, but seems `is_divisor_neg_one` fit better

ceno_zkvm/src/instructions/riscv/div.rs
hero78119: I think the key point here is not Goldilocks field range, it's riscv signed 32 bit registet can not represent overflow 2^31. So riscv define a special case for it.

hero78119 (commented):
reviewed on the arithmetics & constrains design. Docs explanations fit well with constrains design. 👍 
Stay tune for for unittest added then it's good to go!

ceno_zkvm/src/instructions/riscv/div.rs
matthiasgoergens: I know.  I am saying that the special case disappears in our setting!

ceno_zkvm/src/instructions/riscv/div.rs
matthiasgoergens: I tried to write a simple test, but it kept failing.  (Not sure that's because I made some mistakes, or because the code here is wrong.)

So I would suggest that we only merge this PR once we have some passing tests.

ceno_zkvm/src/instructions/riscv/div.rs
```diff
@@ -179,7 +483,7 @@ mod test {
             circuit_builder::{CircuitBuilder, ConstraintSystem},
             instructions::{
                 Instruction,
-                riscv::{constants::UInt, div::DivUInstruction},
```
matthiasgoergens, line 182: We can do that rename in its own PR, to make this one easier to review:

https://github.com/scroll-tech/ceno/pull/766

ceno_zkvm/src/instructions/riscv/div.rs
bgillesp: Agreed the "max negative" phrasing is a bit awkward -- is `is_dividend_signed_min` okay?

bgillesp:
Just incorporated the updated interface for signed instance assignment to the `AssertLtConfig` gadget used by the `remainder_nonnegative` field, fixed up the documentation to address the bug we found in normalizing the signed remainder, and addressed some comments from @hero78119.  The new tests are passing with the updated assignment, and I think this should be good to go.  Give it one last look over @mcalancea?

mcalancea (commented):
Unit tests added, releasing my requested changes.

hero78119 (approved):
👍 LGTM!

mcalancea (approved):

PR #597, opened 2024-11-18
test/canary: troubleshoot corrupt witness memory

naure:
This is the canary method to find missing assignments in `RowMajorMatrix`.

If interested, I can turn this into a testing feature.


naure:
Better improve the API instead of this hack. Issue https://github.com/scroll-tech/ceno/issues/600

PR #599, opened 2024-11-19
feat: official runner

lightsing:


scripts/ci/Dockerfile
```diff
@@ -1,31 +1,7 @@
...
-    && curl -o actions-runner-linux-x64-$RUNNER_VERSION.tar.gz -L https://github.com/actions/runner/releases/download/v$RUNNER_VERSION/actions-runner-linux-x64-$RUNNER_VERSION.tar.gz \
-    && echo "$CHECKSUM  actions-runner-linux-x64-$RUNNER_VERSION.tar.gz" | shasum -a 256 -c \
-    && tar xzf ./actions-runner-linux-x64-$RUNNER_VERSION.tar.gz
-
-RUN chown -R docker ~docker && cd /home/docker/actions-runner/bin/ \
-    && chmod +x installdependencies.sh \
-    && ./installdependencies.sh
+FROM ghcr.io/actions/actions-runner:latest
 
+RUN sudo ./bin/installdependencies.sh
```
matthiasgoergens, line 3: Oh, we still need to run this, it's not already run on the base image?  Interesting.

scripts/ci/compose.yml
```diff
@@ -0,0 +1,8 @@
+name: ceno
+services:
+  runner:
+    restart: always
+    build:
+      context: .
+      dockerfile: Dockerfile
+    env_file: .env
```
matthiasgoergens, line 8: We should probably put the `.env` in the `.gitignore`, so people don't commit one by accident?

lightsing: yeah you are right

.gitignore
```diff
@@ -7,3 +7,4 @@ report.txt
 report.json
 table_cache_dev_*
 .DS_Store
+.env
```
matthiasgoergens, line 10: Please set up your editor to all end lines (especially the last line) with a newline.  It's just less hassle with git that way.

I can fix it for this particular instance, but it would be useful to fix your config in general.  Thanks!

scripts/ci/Dockerfile
```diff
@@ -1,31 +1,7 @@
...
 COPY start.sh start.sh
-RUN chmod +x start.sh
-
-# GitHub action runner scripts (config.sh and run.sh) are not allowed to be run as root
-USER docker
+RUN sudo chown runner:runner start.sh && chmod +x start.sh
```
matthiasgoergens, line 5: Why does `start.sh` need to be owned by the `runner` user?

matthiasgoergens: OK?  Is that relevant?

scripts/ci/README.md
matthiasgoergens, line 1: Thanks for the write-up.  Could you also please mention _where_ you need to run these commands?  Presumably not on your laptop?

You can add some words to explain that this is only for people at Scroll, and that if you work on an open source fork, this doesn't apply to you.  Thanks.

icemelon: @lightsing could you address this comments?

lispc (approved):

icemelon (approved):

PR #601, opened 2024-11-19
Revert dyn ram

naure:
- Revert #368 for `DynVolatileRamTableConfig`.
- Add padding of `final_cycle`.

### DIFF
--- a/ceno_zkvm/src/tables/ram/ram_impl.rs
+++ b/ceno_zkvm/src/tables/ram/ram_impl.rs
@@ -391,6 +391,7 @@ impl<DVRAM: DynVolatileRamTable + Send + Sync + Clone> DynVolatileRamTableConfig
                         set_val!(row, limb, 0u64);
                     });
                     set_val!(row, self.addr, DVRAM::addr(paddin_entry_start + i) as u64);
+                    set_val!(row, self.final_cycle, 0_u64);
                 });
         }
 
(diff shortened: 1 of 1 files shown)


hero78119 (approved):
The revert back version look correct to me!

Btw @naure I am curious, does it make sense to switch fabonacci static_mem to dyn ram ?

naure:
@hero78119 The layout of even that simple program is like:

```
ELF segment 0: R-- vaddr=0x00010000 file_size=276 mem_size=276
ELF segment 1: R-X vaddr=0x00200800 file_size=35088 mem_size=35088
ELF segment 2: R-- vaddr=0x0020a110 file_size=5608 mem_size=5608
ELF segment 3: RW- vaddr=0x0020c6f8 file_size=0 mem_size=220
Stack:             vaddr=0x0020_0400 size=1024
```

Which works really well with a single fixed set of addresses.

ceno_zkvm/src/tables/ram/ram_impl.rs
```diff
@@ -353,41 +353,47 @@ impl<DVRAM: DynVolatileRamTable + Send + Sync + Clone> DynVolatileRamTableConfig
...
+                        let val = (rec.value >> (l * LIMB_BITS)) & LIMB_MASK;
+                        set_val!(row, limb, val as u64);
+                    });
+                }
+                set_val!(row, self.final_cycle, rec.cycle);
+            });
+
+        // set padding with well-form address
+        if final_mem.len().next_power_of_two() - final_mem.len() > 0 {
+            let paddin_entry_start = final_mem.len();
```
kunxian-xia, line 382: I think this has implicit assumption that 
1. addr in `final_mem` is mononically increasing and is divisible by 4;
2. `final_mem[0].addr == DVRAM::addr(0)`;
3. `final_mem.last().unwrap().addr == DVRAM::addr(final_mem.len()-1)`;

Therefore we should add assertion to avoid this issue.

PR #603, opened 2024-11-20
Clean up after 'WIP: integration test (#368)'

matthiasgoergens:
While reviewing commit 63ad6f9c35e4d270d8334c71468fcdd21ab454b7 / PR https://github.com/scroll-tech/ceno/pull/368 I came up with a few clean ups and improvements.  Nothing earth shattering, but might as well fix it.

Sorry, I didn't have time to review it, before it went into `master`.


ceno_zkvm/examples/fibonacci_elf.rs
```diff
@@ -35,6 +35,24 @@ struct Args {
     max_steps: Option<usize>,
 }
 
+fn with_panic_hook<F, R>(hook: Box<dyn Fn(&PanicHookInfo<'_>) + Sync + Send + 'static>, f: F) -> R
```
matthiasgoergens, line 38: I feel this is a bit cleaner for the reader to extract, then inlining it into the code further down.

lispc:
seems ok.  How did you find these?  reading codes? cargo clippy?

matthiasgoergens:
> seems ok. How did you find these? reading codes? cargo clippy?

The main goal for me was reading through the code of https://github.com/scroll-tech/ceno/pull/368 to understand what's going on.  These small improvements are just a by-product of that active reading.

ceno_zkvm/examples/fibonacci_elf.rs
```diff
@@ -269,18 +295,16 @@ fn main() {
     }
 
     let transcript = Transcript::new(b"riscv");
-    // change public input maliciously should cause verifier to reject proof
+    // Maliciously changing the public input should cause the verifier to reject the proof.
     zkvm_proof.raw_pi[0] = vec![<GoldilocksExt2 as ff_ext::ExtensionField>::BaseField::ONE];
     zkvm_proof.raw_pi[1] = vec![<GoldilocksExt2 as ff_ext::ExtensionField>::BaseField::ONE];
 
-    // capture panic message, if have
-    let default_hook = panic::take_hook();
```
matthiasgoergens, line 277: @lispc Here's an example where I had some (minor) trouble understanding what the original code tried to do.  So I cleaned it up into its own helper function, in order to make the job of the next reader down the line a bit easier.

matthiasgoergens:
Ha, interesting we get some test failures!  Seems like I didn't understand the original code well enough.  Time to learn something.

ceno_zkvm/examples/fibonacci_elf.rs
naure: Might as well go to the root cause and change the assert to a returned error.

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
```diff
@@ -142,14 +138,9 @@ impl MemPadder {
...
-            self.padded(new_len, records)
-        } else {
-            self.padded(new_len, records)
-                .into_iter()
-                .sorted_by_key(|record| record.addr)
-                .collect()
-        }
+        let mut padded = self.padded(new_len, records);
+        padded.sort_by_key(|record| record.addr);
+        padded
```
naure, line 143: This does extra work without being clearer when we just want the default content. If you really want to remove that `if` then it's more meaningful to split it into two functions.

ceno_zkvm/examples/fibonacci_elf.rs
```diff
@@ -209,7 +230,7 @@ fn main() {
             MemFinalRecord {
                 addr: rec.addr,
                 value: vm.peek_memory(vma),
-                cycle: *final_access.get(&vma).unwrap_or(&0),
+                cycle: final_access.get(&vma).copied().unwrap_or_default(),
```
naure, line 233: 0 is the intended value. No reason to hide it.

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
matthiasgoergens: Oh, I didn't actually care that much about the `if`.  It was more about converting from vector to iterator (then internally in `sorted_by_key` convert back to vector and sort then convert back to iterator) and then back to vector again.  I'm not at all against letting the computer do some extra work, but here it didn't even make the code easier to read.

But while I was at it, I also simplified the structured by removing the `if`.  The fewer branching code paths you have, the easier it is to achieve good test coverage of all the code paths.  Rust's sorting is very efficient for already sorted data: it finishes in O(n).  And we need O(n) anyway just to produce the data in the first place.

ceno_zkvm/examples/fibonacci_elf.rs
matthiasgoergens: I'm fine either way here.

ceno_zkvm/examples/fibonacci_elf.rs
matthiasgoergens: Interesting suggestion.  Can you make a PR for that, please?

I didn't want to change the behaviour here, just conservatively make the code that we have easier to understand.

ceno_emul/src/vm_state.rs
```diff
@@ -79,13 +79,7 @@ impl VMState {
...
     pub fn iter_until_halt(&mut self) -> impl Iterator<Item = Result<StepRecord>> + '_ {
         let emu = Emulator::new();
-        from_fn(move || {
-            if self.halted() {
-                None
-            } else {
-                Some(self.step(&emu))
-            }
-        })
+        from_fn(move || self.halted().not().then(|| self.step(&emu)))
```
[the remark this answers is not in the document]
matthiasgoergens: That's exactly what I thought as well!  But both `halted` and `step` want to borrow `self`, so the borrow checker didn't like that.

I was very tempted to take that as a suggestion that we should refactor.  But that would have been way beyond the narrow scope of what I was trying to do here.

If you make a PR to that effect, I'm happy to review.

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
naure: See https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.sorted_by_key

ceno_emul/src/vm_state.rs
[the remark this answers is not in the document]
naure: There would be a lot of comments if we’d start documenting hypothetical things that don’t work.

PR #608, opened 2024-11-20
[exp] Dynamic platform attempt

mcalancea:
Makes the platform dynamic:
- Extracts some of the more fixed behaviors of `Platform` into static methods, which replace some calls to `CENO_PLATFORM`.
- Makes the `NVRAM/DVRAM` trait methods take a `Platform` argument
- Config types now store a `Platform` argument to supply `NVRAM/DVRAM` calls.
- Config types get their platform from `ConstraintSystem`, which in turn gets it from `ZKVMConstraintSystem`.
- Makes the constraint system take an explicit platform (currently the old constructors default to `CENO_PLATFORM`). See `fibonacci_elf.rs`.

Observations:
- Could instate some safeguards to use the same platform throughout the pipeline. One example: presently it's technically possible to use one platform for building the `VM` and another for the constraint system.

### DIFF
--- a/ceno_zkvm/src/tables/ram.rs
+++ b/ceno_zkvm/src/tables/ram.rs
@@ -1,4 +1,4 @@
-use ceno_emul::{Addr, CENO_PLATFORM, VMState, WORD_SIZE};
+use ceno_emul::{Addr, Platform, VMState, WORD_SIZE};
 use ram_circuit::{DynVolatileRamCircuit, NonVolatileRamCircuit, PubIORamCircuit};
 
 use crate::{instructions::riscv::constants::UINT_LIMBS, structs::RAMType};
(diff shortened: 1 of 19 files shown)


ceno_zkvm/src/tables/ram/ram_circuit.rs
```diff
@@ -128,17 +128,20 @@ pub trait DynVolatileRamTable {
     const RAM_TYPE: RAMType;
     const V_LIMBS: usize;
 
-    const OFFSET_ADDR: Addr;
-    const END_ADDR: Addr;
+    // const OFFSET_ADDR: Addr;
+    // const END_ADDR: Addr;
+
+    fn offset_addr(platform: &Platform) -> Addr;
+    fn end_addr(platform: &Platform) -> Addr;
```
naure, line 135: `end_addr` is not actually needed in the trait.

ceno_zkvm/src/tables/ram/ram_circuit.rs
```diff
@@ -128,17 +128,20 @@ pub trait DynVolatileRamTable {
...
 
     fn name() -> &'static str;
 
-    fn max_len() -> usize {
-        (Self::END_ADDR - Self::OFFSET_ADDR) as usize / WORD_SIZE
+    fn max_len(platform: &Platform) -> usize {
+        (Self::end_addr(platform) - Self::offset_addr(platform)) as usize / WORD_SIZE
```
naure, line 140: This default body is not needed since it is overridden in the only implementation just above.

ceno_zkvm/src/tables/ram.rs
```diff
@@ -13,15 +13,21 @@ pub struct MemTable;
...
+    }
 
     fn name() -> &'static str {
         "MemTable"
     }
 
-    fn max_len() -> usize {
-        let max_size = (Self::END_ADDR - Self::OFFSET_ADDR) / WORD_SIZE as Addr;
+    fn max_len(platform: &Platform) -> usize {
+        let max_size = (Self::end_addr(platform) - Self::offset_addr(platform)) / WORD_SIZE as Addr;
```
naure, line 30: This is inclusive end (possibly `u32::MAX`). More reliable is:

```suggestion
        let max_size = (Self::end_addr(platform) - Self::offset_addr(platform)).div_ceil(WORD_SIZE as Addr);
```

ceno_zkvm/src/tables/ram.rs
```diff
@@ -56,7 +62,7 @@ impl NonVolatileTable for StaticMemTable {
     const V_LIMBS: usize = 1; // See `MemoryExpr`.
     const WRITABLE: bool = true;
 
-    fn len() -> usize {
+    fn len(_platform: &Platform) -> usize {
         // TODO: take as program parameter.
```
naure, line 66: So close! Might as well solve these TODOs right away.

naure:
This approach is simple and a neat improvement. Let’s open it (vs "draft")?

ceno_zkvm/src/tables/ram/ram_circuit.rs
mcalancea: It's used in `max_len()` (the overwritten impl as well), no?

kunxian-xia (approved):
Nice job!

PR #611, opened 2024-11-20
fix/program-size: fix padding depending on program table size

naure:
This produces invalid proofs (too few rounds) if the actual program size is less than `PROGRAM_SIZE / 2`.

### DIFF
--- a/ceno_zkvm/src/tables/program.rs
+++ b/ceno_zkvm/src/tables/program.rs
@@ -173,7 +173,16 @@ impl<E: ExtensionField, const PROGRAM_SIZE: usize> TableCircuit<E>
             });
 
         assert_eq!(INVALID as u64, 0, "cannot use 0 as program padding");
-        Self::padding_zero(&mut fixed, num_fixed).expect("padding error");
+        fixed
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/tables/program.rs
```diff
@@ -172,7 +172,17 @@ impl<E: ExtensionField, const PROGRAM_SIZE: usize> TableCircuit<E>
                 }
             });
 
-        Self::padding_zero(&mut fixed, num_fixed).expect("padding error");
+        assert_eq!(INVALID as u64, 0, "cannot use 0 as program padding");
```
matthiasgoergens, line 175: Sorry, why is 0 not allowed?

naure: Yes that would work too. The spirit is to have a generic padding strategy, treating all fields at once, not aware of type InsnKind. Ultimately I’d like to refactor it into such generic `padding` function.

naure: The comment could be "0 must be valid padding".

PR #615, opened 2024-11-21
fix/program-size2: refactor padding_zero

naure:
Follow-up to #611.


ceno_zkvm/src/witness.rs
```diff
@@ -87,12 +87,12 @@ impl<T: Sized + Sync + Clone + Send + Copy> RowMajorMatrix<T> {
 
     pub fn par_batch_iter_padding_mut(
         &mut self,
-        num_rows: usize,
+        skip: usize,
```
hero78119, line 90: `skip` better be optional, and we can keep `let valid_instance = self.num_instances();` if it's None

naure: Done.

hero78119 (approved):
LGTM!

Issue #614, opened 2024-11-21
Private Input circuit


naure:

hero78119:
Related to marked of TODOs here 
https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/tables/ram/ram_impl.rs#L402-L417

In the end it seems to support {Dyn, Fix} x {Volatile, Non Volatile}, in totally 4 combination

hero78119:
Done via #617 & #622

PR #617, opened 2024-11-21
feat/private-input: Unconstrained memory init

naure:
_Issues #614 and #607_

Initialize memory with unconstrained prover hints.

- It is practically identical to the zero-init circuit; just without the zero-init part.
- The address range must be contiguous, but its size is dynamic.
- The `u32` range check is already done by the LOAD and STORE circuits.

e2e integration will be done in another PR. Need #608 to configure the address space.

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/memory/load.rs
+++ b/ceno_zkvm/src/instructions/riscv/memory/load.rs
@@ -84,6 +84,7 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for LoadInstruction<E,
     ) -> Result<Self::InstructionConfig, ZKVMError> {
         let rs1_read = UInt::new_unchecked(|| "rs1_read", circuit_builder)?; // unsigned 32-bit value
         let imm = circuit_builder.create_witin(|| "imm"); // signed 12-bit value
+        // Memory initialization is not guaranteed to contain u32. Decompose in [u16; 2] here.
         let memory_read = UInt::new(|| "memory_read", circuit_builder)?;
 
         let memory_addr = match I::INST_KIND {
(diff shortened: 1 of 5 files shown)


hero78119 (approved):
LGTM!

ceno_zkvm/src/instructions/riscv/memory/load.rs
```diff
@@ -84,6 +84,7 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for LoadInstruction<E,
     ) -> Result<Self::InstructionConfig, ZKVMError> {
         let rs1_read = UInt::new_unchecked(|| "rs1_read", circuit_builder)?; // unsigned 32-bit value
         let imm = circuit_builder.create_witin(|| "imm"); // signed 12-bit value
+        // Memory initialization is not guaranteed to contain u32. Decompose in [u16; 2] here.
```
kunxian-xia, line 87: Each memory slot MUST be initialized with u32, right?

ceno_zkvm/src/instructions/riscv/memory/store.rs
```diff
@@ -65,6 +65,7 @@ impl<E: ExtensionField, I: RIVInstruction, const N_ZEROS: usize> Instruction<E>
     ) -> Result<Self::InstructionConfig, ZKVMError> {
         let rs1_read = UInt::new_unchecked(|| "rs1_read", circuit_builder)?; // unsigned 32-bit value
         let rs2_read = UInt::new_unchecked(|| "rs2_read", circuit_builder)?;
+        // Memory initialization is not guaranteed to contain u32. Decompose in [u16; 2] here.
```
kunxian-xia, line 68: ditto.

ceno_zkvm/src/instructions/riscv/memory/load.rs
naure: Actually not, the private input circuit itself will allow non-u32.

It is the LOAD or STORE that touch a private input value that will range-check it to `u32`. This was already in place, I just highlighted it with this comment.

kunxian-xia (approved):

ceno_zkvm/src/instructions/riscv/memory/load.rs
kunxian-xia: I still don't get it why prover want to assign non-u32 to memory slot.

naure: Because they’re evil and trying to make fake proofs.

kunxian-xia: ok, i see. Load instruction needs to range check the value it reads from write set. Since both store instructions and memory init circuits are allowed to insert write record into write set, it's possible a malicious prover will try to insert some random number here.

PR #620, opened 2024-11-21
cleanup/exclusive-ranges

naure:
- Switch to half-open / usual range for platform addresses.
- Expose the `Range` type (#621).
- Introduce methods `iter_addresses` which is not the same as `Range::iter`.

(This became easy to fix after recent changes and useful for upcoming changes.)

### DIFF
--- a/ceno_emul/src/addr.rs
+++ b/ceno_emul/src/addr.rs
@@ -14,7 +14,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fmt, ops};
+use std::{
+    fmt,
+    ops::{self, Range},
+};
 
 pub const WORD_SIZE: usize = 4;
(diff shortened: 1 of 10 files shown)


KimiWu123 (approved):
LGTM!

lispc (approved):
i read this and think it behaves identical to prev version.

Only exception is the `cfg!(feature = "forbid_overflow")` feature flag.

Can be merged if you think it is ok

naure:
Thanks for checking!

Nice catch of `forbid_overflow`. I removed this check because it is already at the place that needs it ([store circuit](https://github.com/scroll-tech/ceno/blob/0efb9d1b380e8eaa49add1fa3a5ca9cecc132c96/ceno_zkvm/src/instructions/riscv/memory/store.rs#L82)).

PR #621, opened 2024-11-21
cleanup/platform-range

naure:
_Follow-up to #620_

Simplify the `Platform` type by exposing the `Range` type.


KimiWu123 (approved):
Nice cleanup!

PR #622, opened 2024-11-21
Feat: Private input integration

naure:
_Issue #614_.
_Pending a solution in #625_

- Load the private input from a file.
- Configuration of the address range.
- Integrate the circuit  (#617) in MMU.

### DIFF
--- a/ceno_emul/src/addr.rs
+++ b/ceno_emul/src/addr.rs
@@ -14,7 +14,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fmt, ops};
+use std::{
+    fmt,
+    ops::{self, Range},
+};
 
 pub const WORD_SIZE: usize = 4;
 pub const PC_WORD_SIZE: usize = 4;
@@ -190,5 +193,31 @@ impl ops::AddAssign<usize> for ByteAddr {
 impl ops::AddAssign<u32> for ByteAddr {
     fn add_assign(&mut self, rhs: u32) {
         self.0 += rhs;
+    }
+}
+
+pub trait IterAddresses {
+    fn iter_addresses(&self) -> impl Iterator<Item = Addr>;
+}
+
+impl IterAddresses for Range<Addr> {
+    fn iter_addresses(&self) -> impl Iterator<Item = Addr> {
+        self.clone().step_by(WORD_SIZE)
+    }
+}
+
+impl<'a, T: GetAddr> IterAddresses for &'a [T] {
+    fn iter_addresses(&self) -> impl Iterator<Item = Addr> {
+        self.iter().map(T::get_addr)
+    }
+}
+
+pub trait GetAddr {
+    fn get_addr(&self) -> Addr;
+}
+
+impl GetAddr for Addr {
+    fn get_addr(&self) -> Addr {
+        *self
     }
 }
--- a/ceno_emul/src/addr.rs
+++ b/ceno_emul/src/addr.rs
@@ -197,17 +197,17 @@ impl ops::AddAssign<u32> for ByteAddr {
 }
 
 pub trait IterAddresses {
-    fn iter_addresses(&self) -> impl Iterator<Item = Addr>;
+    fn iter_addresses(&self) -> impl ExactSizeIterator<Item = Addr>;
 }
 
 impl IterAddresses for Range<Addr> {
-    fn iter_addresses(&self) -> impl Iterator<Item = Addr> {
+    fn iter_addresses(&self) -> impl ExactSizeIterator<Item = Addr> {
         self.clone().step_by(WORD_SIZE)
     }
 }
 
 impl<'a, T: GetAddr> IterAddresses for &'a [T] {
-    fn iter_addresses(&self) -> impl Iterator<Item = Addr> {
+    fn iter_addresses(&self) -> impl ExactSizeIterator<Item = Addr> {
         self.iter().map(T::get_addr)
     }
 }
--- a/ceno_emul/src/platform.rs
+++ b/ceno_emul/src/platform.rs
@@ -12,6 +12,7 @@ pub struct Platform {
     pub rom: Range<Addr>,
     pub ram: Range<Addr>,
     pub public_io: Range<Addr>,
+    pub private_io: Range<Addr>,
     pub stack_top: Addr,
     /// If true, ecall instructions are no-op instead of trap. Testing only.
     pub unsafe_ecall_nop: bool,
@@ -21,6 +22,7 @@ pub const CENO_PLATFORM: Platform = Platform {
     rom: 0x2000_0000..0x3000_0000,
     ram: 0x8000_0000..0xFFFF_0000,
     public_io: 0x3000_1000..0x3000_2000,
+    private_io: 0x4000_0000..0x5000_0000,
     stack_top: 0xC0000000,
     unsafe_ecall_nop: false,
 };
@@ -38,6 +40,10 @@ impl Platform {
 
     pub fn is_pub_io(&self, addr: Addr) -> bool {
         self.public_io.contains(&addr)
+    }
+
+    pub fn is_priv_io(&self, addr: Addr) -> bool {
+        self.private_io.contains(&addr)
     }
 
     /// Virtual address of a register.
@@ -60,7 +66,7 @@ impl Platform {
     // Permissions.
 
     pub fn can_read(&self, addr: Addr) -> bool {
-        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr)
+        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr) || self.is_priv_io(addr)
     }
 
     pub fn can_write(&self, addr: Addr) -> bool {
--- a/ceno_emul/src/platform.rs
+++ b/ceno_emul/src/platform.rs
@@ -1,3 +1,5 @@
+use std::ops::Range;
+
 use crate::addr::{Addr, RegIdx};
 
 /// The Platform struct holds the parameters of the VM.
@@ -7,20 +9,18 @@ use crate::addr::{Addr, RegIdx};
 /// - codes of environment calls.
 #[derive(Clone, Debug)]
 pub struct Platform {
-    pub rom_start: Addr,
-    pub rom_end: Addr,
-    pub ram_start: Addr,
-    pub ram_end: Addr,
+    pub rom: Range<Addr>,
+    pub ram: Range<Addr>,
+    pub public_io: Range<Addr>,
     pub stack_top: Addr,
     /// If true, ecall instructions are no-op instead of trap. Testing only.
     pub unsafe_ecall_nop: bool,
 }
 
 pub const CENO_PLATFORM: Platform = Platform {
-    rom_start: 0x2000_0000,
-    rom_end: 0x3000_0000,
-    ram_start: 0x8000_0000,
-    ram_end: 0xFFFF_0000,
+    rom: 0x2000_0000..0x3000_0000,
+    ram: 0x8000_0000..0xFFFF_0000,
+    public_io: 0x3000_1000..0x3000_2000,
     stack_top: 0xC0000000,
     unsafe_ecall_nop: false,
 };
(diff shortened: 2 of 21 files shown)


ceno_zkvm/src/bin/e2e.rs
```diff
@@ -94,6 +100,12 @@ fn main() {
     let elf_bytes = fs::read(&args.elf).expect("read elf file");
     let mut vm = VMState::new_from_elf(platform.clone(), &elf_bytes).unwrap();
 
+    tracing::info!("Loading private input file: {:?}", args.private_input);
+    let priv_io = memory_from_file(&args.private_input);
+    for (addr, value) in zip(platform.private_io.iter_addresses(), &priv_io) {
+        vm.init_memory(addr.into(), *value);
+    }
```
lispc, line 107: should we add a comment to say, if len(private_io) != len(private_input_file), then it is right-padded with 0s?

lispc (approved):

ceno_zkvm/src/bin/e2e.rs
naure: Good point. It is **not** padded. Or precisely it is zero-padded to the next power-of-two size.

The rest of memory does not exist - unsatisfiable if the program tries to use it.

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
```diff
@@ -61,6 +71,7 @@ impl<E: ExtensionField> MmuConfig<E> {
         );
 
         fixed.register_table_circuit::<PubIOCircuit<E>>(cs, &self.public_io_config, io_addrs);
+        fixed.register_table_circuit::<PrivateIOCircuit<E>>(cs, &self.private_io_config, &());
```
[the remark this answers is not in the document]
naure: It is an associated type not used in this case.

```rust
impl TableCircuit for DynVolatileRamCircuit {
    type FixedInput = ();
…
    fn generate_fixed_traces(…, _: &Self::FixedInput)
```

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: For technical reasons, [if we are using rkyv, it would be better to pad from the left by default.](https://github.com/scroll-tech/ceno/discussions/607)

That's means the last byte that's specified should always appear in the same position in the VM.

> The rest of memory does not exist - unsatisfiable if the program tries to use it.

That's good.

matthiasgoergens (commented):
Repeating my comment on the PR here for visibility:

For rkyv it would be best if we could pad on the left, ie the last byte specified by the user of the prover should always appear in the same address in the VM.  Where the first byte appears should be a function of the length of the input specified.

As far as I can tell, we pad the input given with 0s to the next power of two, and then we pad the rest with ‘inaccessible bytes’.  That’s good, but we need them from the left.

We can either hard code this to come from the left, or we can make it so that the ‘core’ functions always need the placement specified exactly, and then have a relatively thin shell of helper functionality on the outside that does the necessary calculations.

(The latter would be useful, if we want to explore other deserialisation schemes, because I suspect rkyv is a bit special in wanting left padding.  But we can also just hardcode it now, because having extra mostly unused customisation options around is also a hassle in testing and maintenance.)

ceno_emul/src/addr.rs
```diff
@@ -197,17 +197,17 @@ impl ops::AddAssign<u32> for ByteAddr {
 }
 
 pub trait IterAddresses {
-    fn iter_addresses(&self) -> impl Iterator<Item = Addr>;
+    fn iter_addresses(&self) -> impl ExactSizeIterator<Item = Addr>;
```
matthiasgoergens, line 200: Interesting!  Why do we need the additional guarantees?

(I wonder if we should also add `+ DoubleEndedIterator`, or perhaps just give a concrete type?  But that's just idle musing.)

ceno_emul/src/platform.rs
```diff
@@ -60,7 +66,7 @@ impl Platform {
     // Permissions.
 
     pub fn can_read(&self, addr: Addr) -> bool {
-        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr)
+        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr) || self.is_priv_io(addr)
```
matthiasgoergens, line 69: We might want to make the emulator match what the circuits allow?  Mostly to simplify reasoning about soundness.

As far as I can tell from what you wrote, only some parts of the `private_io` address range are actually readable?  (And I suspect it's similar for the other address ranges as well?)

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -41,6 +43,11 @@ struct Args {
     /// The preset configuration to use.
     #[arg(short, long, value_enum, default_value_t = Preset::Ceno)]
     platform: Preset,
+
+    /// The private input or hints. This is a raw file mounted as a memory segment.
```
matthiasgoergens, line 47: I'd say the raw file is 'mapped' as a memory segment.  Mounting is generally used for file systems.

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -41,6 +43,11 @@ struct Args {
...
+    /// Zero-padded to the next power-of-two size.
```
matthiasgoergens, line 48: Please be explicitly in which direction we are padding.

As written elsewhere, left padding would be most useful to use.

Also as far as I can tell, we are zero-padding to the next power of two, end then 'inaccessible'-padding to fill up the total range?  We should probably mention the two different kinds of padding in the doc comment.

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -94,6 +101,17 @@ fn main() {
     let elf_bytes = fs::read(&args.elf).expect("read elf file");
     let mut vm = VMState::new_from_elf(platform.clone(), &elf_bytes).unwrap();
 
+    tracing::info!("Loading private input file: {:?}", args.private_input);
```
matthiasgoergens, line 104: On the discussion about private IO, someone suggested to use the term 'hint' throughout?

('Private input' works perfectly fine for me; but I remember some people having strong opinions on this.)

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -332,6 +359,18 @@ fn main() {
     };
 }
 
+fn memory_from_file(path: &Option<String>) -> Vec<u32> {
+    path.as_ref()
+        .map(|path| {
+            let mut buf = fs::read(path).expect("could not read file");
+            buf.resize(buf.len().next_multiple_of(WORD_SIZE), 0);
```
matthiasgoergens, line 366: Be careful about a buffer length of 0 here, please.  0 ain't a power of two, but I think we can probably leave that unpadded?

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
```diff
@@ -48,7 +52,13 @@ impl<E: ExtensionField> MmuConfig<E> {
         io_addrs: &[Addr],
     ) {
         assert!(
-            chain!(static_mem_init.iter_addresses(), io_addrs.iter_addresses()).all_unique(),
+            chain!(
+                static_mem_init.iter_addresses(),
+                io_addrs.iter_addresses(),
+                // TODO: optimize with min_max and Range.
```
matthiasgoergens, line 58: Why does specifically this one need optimisation?

matthiasgoergens (approved):
Please see my comments.

Specifically I want to highlight two points:
1. left padding is more useful that right padding for us.
2. could we unify the padding to be 'inaccessible' values only, instead of a mix of 0 and 'inaccessible'?

If we do right padding by default, please leave a comment that it's because of rkyv.  If you make it configurable, then we'll leave the comment on one of the outer layer helper functions.

ceno_emul/src/addr.rs
naure: For `len()`

ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs
naure: This one is a Range, others are sets.

ceno_zkvm/src/bin/e2e.rs
naure: 0 stays 0 here no?

ceno_zkvm/src/tables/ram/ram_impl.rs
```diff
@@ -376,8 +376,11 @@ impl<DVRAM: DynVolatileRamTable + Send + Sync + Clone> DynVolatileRamTableConfig
             .par_iter_mut()
             .with_min_len(MIN_PAR_SIZE)
             .zip(final_mem.into_par_iter())
-            .for_each(|(row, rec)| {
+            .enumerate()
```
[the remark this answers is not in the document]
naure: We should pass the real size here because the matrix type keeps track of given vs padded sizes. 0 is acceptable and padded to 2.

ceno_zkvm/src/bin/e2e.rs
naure: Interesting idea of dynamic memory growing leftwards. Can you write an issue for it?

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: `0.next_power_of_two()` is 1, I think.

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: See https://github.com/scroll-tech/ceno/issues/628

ceno_zkvm/src/bin/e2e.rs
naure: This is `next_multiple_of`.

matthiasgoergens: Sorry, I misread.

ceno_emul/src/platform.rs
naure: That would be ideal yes.

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: Why do you give a thumbs up, but then don't do anything about it?

naure: I meant to include it but I was focused on making the feature work.

ceno_emul/src/platform.rs
matthiasgoergens: Tracked in https://github.com/scroll-tech/ceno/issues/630

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: Ok, no worries, I'll create a PR.

matthiasgoergens: https://github.com/scroll-tech/ceno/pull/633

ceno_zkvm/src/bin/e2e.rs
matthiasgoergens: See also https://github.com/scroll-tech/ceno/issues/634

ceno_zkvm/src/bin/e2e.rs
kunxian-xia: Yeah, I favor `hint` over `private input`. Some places that use this terminology:
1. https://docs.cairo-lang.org/hello_cairo/program_input.html#hints
2. https://github.com/succinctlabs/sp1/blob/dev/crates/zkvm/entrypoint/src/syscalls/io.rs#L61

naure: Thanks for the ref. If the term "hint" is understood I’ll rename things to that.

kunxian-xia: No worries, you can do that after you finish the sproll-evm test.

naure: https://github.com/scroll-tech/ceno/pull/636

PR #623, opened 2024-11-22
feat/private-input-2: fix padding assignment

naure:
Did not work when `len=1` because it’s padded to length 2.
Fixed and simplified.


hero78119 (approved):
👍

lispc (approved):

PR #624, opened 2024-11-22
Refactor row major matrix 

### LINKED ISSUE #600: Refactor `RowMajorMatrix` with a safe API
This type uses `MaybeInit` with the expectation that assignment code will set values in every cell. Moreover, the matrix is usually larger than requested in `new(…)` because it is padded to a power-of-two size. All in all that is error-prone and hard to debug (been there, #597).

Rewrite this with a safe and easy API.

One idea is to use `Vec::with_capacity` and `extend`. Another way could be based on iterators and use `collect_vec`.

Regarding padding, the type should take the responsibility to fill the padding rows. It is not actually necessary to allocate that padding space. The padding can be represented with _a single value to repeat_, and the padding is done at the last possible time in `de_interleaving`. It could look like a method `set_padding_strategy(…)`, see the existing `enum InstancePaddingStrategy`.

mcalancea:
_Written December 10th, most content and conversation predates this description._

Fixes #600.
Makes several changes to the way `RowMajorMatrix` works:

1. It no longer uses `MaybeUninit`. Rather, its contents are initialized to `T::default()` using parallel iterators.
2. Padding is no longer allocated in the constructor. It is also not the concern of any particular user of the matrix to ensure correct padding. Rather, users only pass the `InstancePaddingStrategy` argument which describes the type of padding that they want. This padding is then performed at the very latest stage (in the call to `self.into_mles()`).
3. `self.into_mles()` does more than before. `de_interleaving` is removed because the indirection through a bi-dimensional matrix is unnecessary. Notably, `self.into_mles()` parallelizes the work differently than `de_interleaving` used to.


hero78119:
This is a great effort! As some moment I am also curious what would be the impact if we remove raw assignment and use raw  vector as we only have one implementation.

I just do a very quick tried on this branch against master branch on riscv_add benchmark with command on remote ceno benchmark machine, with command

```
cargo bench --bench riscv_add --package ceno_zkvm -- --baseline baseline
```

and the result turns out to be ~+6% slower on e2e latency.

```
## master branch
add_op_20/prove_add/prove_add_log2_20
                        time:   [1.5172 s 1.5615 s 1.6110 s]
                        change: [-3.4124% -0.5170% +2.7238%] (p = 0.78 > 0.05)

### this branch
add_op_20/prove_add/prove_add_log2_20
                        time:   [1.6454 s 1.6695 s 1.6898 s]
                        change: [+4.1692% +6.3642% +8.4701%] (p = 0.00 < 0.05)
```

I believed with fibonacchi it might  be even much slower. So it's still nessesary to this feature for keeping high performance.

hero78119:
If we just talked about capture this unssignment error, I think we can achieve it from different path: in unittest we support init vector into 2 different default value with same execution trance, and then compare 2 witnesses which should be identical. An unequality indicate there are some unassigment bug. And with that, we can capture the unassignment error in early stage.

ceno_zkvm/src/witness.rs
```diff
@@ -28,98 +29,104 @@ use crate::{
...
+            values: vec![T::default(); num_rows * num_col],
             num_col,
+            padding_strategy,
         }
     }
 
-    pub fn num_instances(&self) -> usize {
-        self.values.len() / self.num_col - self.num_padding_rows
+    pub fn len(&self) -> usize {
+        self.values.len()
```
naure, line 61: Confusing name. Maybe `total_len`, `num_cells`, …?

ceno_zkvm/src/witness.rs
```diff
@@ -28,98 +29,104 @@ use crate::{
...
 
 impl<F: Field> RowMajorMatrix<F> {
     pub fn into_mles<E: ff_ext::ExtensionField<BaseField = F>>(
         self,
     ) -> Vec<DenseMultilinearExtension<E>> {
-        self.de_interleaving().into_mles()
+        let start = Instant::now();
+        let padding_row = match self.padding_strategy {
+            InstancePaddingStrategy::RepeatLast => {
+                self.values[self.values.len() - self.num_col..].to_vec()
```
naure, line 100: What if empty?

There was a special case `if steps.is_empty()` somewhere:

```rust
                InstancePaddingStrategy::RepeatLast if steps.is_empty() => {
                    tracing::debug!("No {} steps to repeat, using zero padding", Self::name());
                    vec![MaybeUninit::new(E::BaseField::ZERO); num_witin]
                }
```

although that was not completely correct either; better than crashing.

mcalancea:
@hero78119 

I got:

naure (commented):
Looking nice so far 👍

matthiasgoergens:
Could you please add to the PR description a very brief overview over what we are doing, and more importantly _why_?

(The what can be really brief, because the text of the PR answers that question in detail.)

Thanks!

matthiasgoergens:
> This is a great effort! As some moment I am also curious what would be the impact if we remove MaybeUint feature and use raw vector as we only have one implementation.

Yes, I have been rather suspicious of `MaybUninit`.  It looks like a bit of a minefield, and I'm not sure it's actually worth it.

hero78119:
> @hero78119
> 
> Thank you! This is still work in progress; I'm examining some ways to make it faster while keeping it clean/safe. My bench results are a bit better for some reason:
> 
> ```
> add_op_20/prove_add/prove_add_log2_20
>                         time:   [4.7420 s 4.7740 s 4.8031 s]
>                         change: [+1.3563% +2.2286% +3.0911%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> add_op_21/prove_add/prove_add_log2_21
>                         time:   [9.9516 s 10.189 s 10.450 s]
>                         change: [-1.7437% +0.5352% +2.8403%] (p = 0.70 > 0.05)
>                         No change in performance detected.
> ```
> 
> What regression % (evaluated at yours) would you consider acceptable?

I noticed the different with yours vs remote ceno benchmark server probably on environment number of cores. 
In multicore environment (16x2 cores), removing `MaybeUinit` the regression will be more significant, e.g. ~6%.
To me any regression of % is somehow unacceptable 😂 

So I would suggest to go from another way: in unittest initialized with 2 different default value with same witness, then check the witness polynomial equality. This help us to identify the problem while not sacrificing performance.

matthiasgoergens:
If you want some inspiration for how to do this kind of change, you might want to have a look at how plonky3 changed the organisation of its data compared to plonky2.  (It might be a bit much too read, though.)

naure:
@hero78119 There is a draft of the approach based on testing here: https://github.com/scroll-tech/ceno/pull/597

naure:
To get back to optimal performance without unsafe types, there could be another constructor that accepts a vector or an iterator. Callers can built their Vec with e.g. flatten and collect.

See also the issue https://github.com/scroll-tech/ceno/issues/600

hero78119 (commented):
Awesome job, and this PR bring more insighful takeaways for the future implementation guidelines 🔥 🔥

hero78119:
Benchmarks result is awesome (still trying to figure out how witness optimisation bring overall e2e improvement so much 😱 🔥 )

On fibonacchi witness generation vs master branch => ~ **-20%** throughput improvement

command
```
cargo bench --bench fibonacci_witness -- --save-baseline fibo_wit
cargo bench --bench fibonacci_witness -- --baseline fibo_wit
```

results (run result twice, each time flushing with complete new baseline)
```
1st result
fib_wit_max_steps_18446744073709551615/fibonacci_witness/fib_wit_max_steps_18446744073709551615
            time:  [1.8106 s 1.8265 s 1.8428 s]
            change: [-22.274% -21.483% -20.588%] (p = 0.00 < 0.05)
            Performance has improved.
2nd result 
fib_wit_max_steps_18446744073709551615/fibonacci_witness/fib_wit_max_steps_18446744073709551615
                        time:   [1.8091 s 1.8224 s 1.8338 s]
                        change: [-23.306% -22.657% -22.023%] (p = 0.00 < 0.05)
                        Performance has improved.
```


On fibonacchi e2e with 2^20 instance: ~ **-15%** throughput improvement

> this is very amazing, still trying to figure out how witness optimisation bring such improvement on overall

results (run result twice, each time flushing with complete new baseline)
```
                        time:   [4.1083 s 4.1283 s 4.1487 s]
                        change: [-16.756% -15.297% -13.856%] (p = 0.00 < 0.05)
                        Performance has improved.

                        time:   [4.0866 s 4.1098 s 4.1323 s]
                        change: [-17.146% -15.677% -14.221%] (p = 0.00 < 0.05)
                        Performance has improved.
```

ceno_zkvm/src/witness.rs
```diff
@@ -28,97 +27,94 @@ use crate::{
...
-    }
-
-    pub fn de_interleaving(mut self) -> Vec<Vec<T>> {
-        (0..self.num_col)
-            .map(|i| {
-                self.values
-                    .par_iter_mut()
-                    .skip(i)
-                    .step_by(self.num_col)
-                    .map(|v| unsafe { mem::replace(v, mem::MaybeUninit::uninit()).assume_init() })
```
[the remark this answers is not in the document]
mcalancea: :+1: Just noting it's pure speculation, I haven't investigated it. But false sharing is perhaps something to keep in mind throughout the code-base.

ceno_zkvm/src/witness.rs
```diff
@@ -28,97 +27,94 @@ use crate::{
...
         self.values.chunks(self.num_col)
     }
 
-    pub fn iter_mut(&mut self) -> ChunksMut<MaybeUninit<T>> {
+    pub fn iter_mut(&mut self) -> ChunksMut<T> {
         self.values.chunks_mut(self.num_col)
     }
 
-    pub fn par_iter_mut(&mut self) -> rayon::slice::ChunksMut<MaybeUninit<T>> {
+    pub fn par_iter_mut(&mut self) -> rayon::slice::ChunksMut<T> {
```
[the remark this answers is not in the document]
mcalancea: :+1: made an[ issue](https://github.com/scroll-tech/ceno/issues/730) to track it.

ceno_zkvm/src/witness.rs
```diff
@@ -28,97 +26,102 @@ use crate::{
...
+        let padding_iter = (num_instances..num_instances + num_padding_instances).map(|i| {
+            match &self.padding_strategy {
+                InstancePaddingStrategy::Custom(fun) => F::from(fun(i as u64, column as u64)),
+                InstancePaddingStrategy::RepeatLast => last_element,
+                _ => F::ZERO,
+            }
+        });
+
+        self.values
+            .iter()
```
hero78119, line 105: "self.values" including non-padding + padding. the iteration here will go through padding part, and the chaining of padding_iter will make padding occur twice.

we can fix here
- self.values range should be [0..num_instances * num_col] for non-default (non zero) padding
- for padding strategy as default (zero), we can skip chaining with padding_iter(), as vector now is initialized deterministics.

hero78119: Just clarified per offline discussion, self.values only contain non-padding value 👍

ceno_zkvm/src/instructions.rs
```diff
@@ -1,26 +1,28 @@
...
-    scheme::constants::MIN_PAR_SIZE,
     witness::{LkMultiplicity, RowMajorMatrix},
 };
-use ff::Field;
 
 pub mod riscv;
 
+#[derive(Clone)]
 pub enum InstancePaddingStrategy {
     Zero,
```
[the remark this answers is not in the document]
mcalancea: agreed :+1:

ceno_zkvm/src/tables/ram/ram_impl.rs
```diff
@@ -369,7 +388,23 @@ impl<DVRAM: DynVolatileRamTable + Send + Sync + Clone> DynVolatileRamTableConfig
     ) -> Result<RowMajorMatrix<F>, ZKVMError> {
         assert!(final_mem.len() <= DVRAM::max_len(&self.params));
         assert!(DVRAM::max_len(&self.params).is_power_of_two());
-        let mut final_table = RowMajorMatrix::<F>::new(final_mem.len(), num_witness);
+
+        let params = self.params.clone();
+        // TODO: make this more robust
+        let addr_column = 0;
```
hero78119, line 394: we can clean up this TODO via
`let addr_column = self.addr.id as u64;`

hero78119 (approved):
epic works and LGTM! 
(just one comment of TODO cleanup)

ceno_zkvm/src/tables/ram/ram_impl.rs
mcalancea: Ah, indeed :+1:

Issue #179, opened 2024-09-02
riscv cross-compiler toolchain integration


naure:

matthiasgoergens:
Done, but it's somewhat manual: you have to copy and paste a few files into a new guest project.

PR #631, opened 2024-11-26
Hint IO in the SDK

matthiasgoergens:
Here we are implementing hints for the SDK.  See this example [Ceno application](https://github.com/scroll-tech/ceno/tree/matthias/build-guest-hint) for how to use the new functionality.

> By Ceno application, we mean the combination of guest program and host support.

See https://github.com/scroll-tech/ceno/discussions/607

Closes https://github.com/scroll-tech/ceno/issues/639

Works towards https://github.com/scroll-tech/ceno/issues/179

Builds on https://github.com/scroll-tech/ceno/pull/754


ceno_host/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+edition.workspace = true
+license.workspace = true
+name = "ceno_host"
+version.workspace = true
+
+[dependencies]
+rkyv = { version = "0.8.9", features = ["unaligned"] }
```
matthiasgoergens, line 8: I need to check whether going without 'unaligned' improves performance in the guest.

It's bit more hassle to make that work, though.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        let archived = rkyv::access::<ArchivedTest, Failure>(&bytes[..]).unwrap();
+        assert_eq!(archived, &value);
+
+        // And you can always deserialize back to the original type
+        let deserialized = deserialize::<Test, Error>(archived).unwrap();
+        assert_eq!(deserialized, value);
+
+        let mut rng = rand::thread_rng();
+
+        {
```
matthiasgoergens, line 172: @naure See these two experiments about padding left and padding right.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        &mut self,
+        item: &impl for<'a> Serialize<HighSerializer<AlignedVec, ArenaHandle<'a>, Error>>,
+    ) -> Result<(), Error> {
+        let bytes = to_bytes::<Error>(item)?;
+        self.items.push(bytes.to_vec());
+        Ok(())
+    }
+
+    pub fn finalise(&self) -> Result<AlignedVec, Error> {
+        to_bytes::<Error>(&self.items)
```
hero78119, line 33: serialization just happened once in the end, which imply iterating the data from begin to end.
can we do streaming processing for more CPU cache friendly?

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        assert_eq!(number, &0xaf_u8);
+        let test2 = read::<ArchivedToast>(&mut iter);
+        assert_eq!(test2, &Toast {
+            stuff: Some(vec!["hello scroll".to_string()]),
+        });
+    }
+
+    #[test]
+    fn test_prepare_and_consume_items() {
+        let stdin = make_stdin();
```
hero78119, line 139: IIUC line 139 & 140 is the host/guest boundary.
In real case, we will write `stdin` to a file then load in guest memory.

Then further question, why left/right padding matters?
In guest we load `stdin` into memory region with fixed start & dynamic end (end can be different per execution trace). Then in guest we just need get start/end, then restore AlignedVec from this specific address `mem[start:end]`. In this case, the padding is transparent & agnostic to `rkyv`?

ceno_host/src/lib.rs
matthiasgoergens: What goal are you trying to accomplish?

Serialisation happens in the host program running on native hardware.  That's plenty fast enough, even if we slowed it down 1000x times.  (And the total amount of data we want to serialise is limited by the fact that we need to deserialise that data in the VM, if it is to be useful.  So it will always be a relatively small amount of data.)

ceno_host/src/lib.rs
hero78119: Introduce left padding is not free in terms of complexity for our current proving system, especially the verifier logic. It make the verifier memory range overlap check harder and more error pruning. I would suggest we better encaptulate it in SDK, as it's just minor effort, and I can already foresee it simplicity.

We can have end address defined in first word, then retrieved as `let end_hint_addr = mem[HINT_IO_REGION_START]`, then use mem[HINT_IO_REGION_START + 4: end_hint_addr] to restore AlignedVec. With that, there is almost no sacrificing with efficient, but we keep backend simple and consistent.

matthiasgoergens: I have a prototype of the new version on GitHub now.  It's a bit more of a mess, because instead of letting rkyv handle two layers of serialisation, we handle the outer layer ourselves, so that we can put its 'entry point' at the start of the hints memory region.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// The equivalent of this function would run in the guest.
+    ///
+    /// `buf` would be the memory mapped region for private hints.
+    pub fn consume(buf: AlignedVec) {
+        println!("\nConsuming...");
+        let (prefix, lengths, suffix) = unsafe { buf.align_to::<u32>() };
+        assert!(prefix.is_empty());
+        assert!(suffix.is_empty());
+        let mut iter = lengths.iter().map(|len| {
```
matthiasgoergens, line 138: This would be part of our library, and not visible in normal guest code.

ceno_host/src/lib.rs
matthiasgoergens: You can already have a look at the new version, if you are feeling brave, but at the moment it's held together with hope and duct-tape.  Let me apply some polish to make it more palatable.

matthiasgoergens: It's saner now, and you can have a look.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn write(
+        &mut self,
+        item: &impl for<'a> Serialize<HighSerializer<AlignedVec, ArenaHandle<'a>, Error>>,
+    ) -> Result<(), Error> {
+        let bytes = to_bytes::<Error>(item)?;
+        self.write_slice(bytes);
+        Ok(())
+    }
+
+    pub fn finalise(&self) -> SerialisedCenoStdin {
```
matthiasgoergens, line 37: Sorry, this is a bit of a mess.  I think it might be comparatively easier to understand the memory format by looking at the readers first, and then you can try to decipher this function.  (Or you can wait for the cleanup up version. :)

The main complexity here comes from me having mixed all the padding logic throughout.  What we are actually doing ain't very complicated conceptually.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn read_slice<'a>(&'a mut self) -> &'b [u8] {
+        self.next().unwrap()
+    }
+}
+
+impl<'a> Iterator for SerialisedCenoStdinIter<'a> {
+    type Item = &'a [u8];
+    fn next(&mut self) -> Option<Self::Item> {
+        let len = u32::from_le_bytes(
+            self.buf.0[self.next..][..size_of::<u32>()]
```
matthiasgoergens, line 91: @kunxian-xia So we actually only store 'pointers' to the end of the slices at the start of our hints memory region.

That's because rkyv doesn't mind, if we give it a slice that has extra garbage at the beginning, so without loss of generality, all the slices here start at the beginning of the hints memory region.  They are overlapping.

We can get away with this, because it's all read-only.  If you want to go read write, you'll want non-overlapping slices, which is also doable at only a slight extra overhead.

matthiasgoergens: Crucially, I have also not put in any information to delineate the end of the first section that stores the slice lengths / end points: a well-intentioned prover will provide all the entries that the guest program will want to read anyway, and a malicious prover could just give you misleading information.

If you accidentally read beyond the intended end, your deserialisation will (likely) fail.

However, I'm open to adding an explicit end, if you think that's useful for eg debugging, even if it makes no difference for security.  Just let me know, what you think, please.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+impl<'b> SerialisedCenoStdinIter<'b> {
+    pub fn read<'a, T>(&'a mut self) -> &'b T
+    where
+        T: Portable + for<'c> CheckBytes<HighValidator<'c, Error>>,
+    {
+        rkyv::access::<T, Error>(self.read_slice()).unwrap()
+    }
+
+    pub fn read_slice<'a>(&'a mut self) -> &'b [u8] {
+        self.next().unwrap()
```
matthiasgoergens, line 83: We intentionally convert the error into a panic here, because we guest programs shouldn't do any elaborate error recovery anyway: an honest prover will prepare all the data/witnesses so that everything goes off without a hitch, and a malicious prover needs to be only caught reliably and get a program abort.

Recovering from errors only makes the latter task harder.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+
+pub struct SerialisedCenoStdinIter<'a> {
+    buf: &'a SerialisedCenoStdin,
+    next: usize,
+}
+
+impl<'b> SerialisedCenoStdinIter<'b> {
+    pub fn read<'a, T>(&'a mut self) -> &'b T
+    where
+        T: Portable + for<'c> CheckBytes<HighValidator<'c, Error>>,
```
matthiasgoergens, line 77: In the guest we would change `Error` to `Failure` here.  `rancor::Error` keeps track of error messages, which is useful for the prototype and in the host.  `rancor::Failure` only keeps track of whether we have an error or not, so that's simpler and faster for the guest.

ceno_host/src/lib.rs
matthiasgoergens: Btw, the 'inner serialisation' happens one by one in something like a streaming fashion (both in the original and in the new design).  At the end we just collate all the individual items, which is mostly a bunch of mem-copies and a tiny bit of logic for the offsets.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    {
+        rkyv::access::<T, Error>(self.read_slice()).unwrap()
```
hero78119, line 79: if I remember correctly, from your previous example, we can throw the whole deliberate slice into rkyv and iterator through it, without implementing a wrapping iterator by ourself?

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+            .collect();
+        let offsets_u8: Vec<u8> = offsets.iter().copied().flat_map(u32::to_le_bytes).collect();
+        let mut buf: AlignedVec = AlignedVec::new();
+        buf.extend_from_slice(&offsets_u8);
+        println!("buf.len() after offsets: {}", buf.len());
+        buf.extend_from_slice(&vec![0; buf.len().next_multiple_of(16) - buf.len()]);
+        println!("buf.len() after offset padding: {}", buf.len());
+        for (offset, item) in izip!(offsets, &self.items) {
+            buf.extend_from_slice(item);
+            buf.extend_from_slice(&vec![0; buf.len().next_multiple_of(16) - buf.len()]);
```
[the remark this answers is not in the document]
matthiasgoergens: Thanks for the suggestion, let me look into that.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+        &mut self,
+        item: &impl for<'a> Serialize<HighSerializer<AlignedVec, ArenaHandle<'a>, Error>>,
+    ) -> Result<(), Error> {
+        let bytes = to_bytes::<Error>(item)?;
+        self.write_slice(bytes);
+        Ok(())
+    }
+
+    pub fn finalise(&self) -> SerialisedCenoStdin {
+        // TODO: perhaps don't hardcode 16 here.
```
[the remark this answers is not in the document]
matthiasgoergens: Rkyv specifically wants 16 bytes (unless you use the `unaligned` feature), which is 4 words on u32.

No worries, the alignment to 16 bytes is only a requirement for the backing storage.  Internally rkyv only uses one eg 4 bytes to store a `u32` in a struct or array somewhere, they don't blow it up to 16 bytes for every single element.

ceno_host/src/lib.rs
matthiasgoergens: Yes, in principle.  But that only works, if we make sure that the end of the mapped data always shows up at a consistent fixed memory address.  `rkyv` finds its data from the end.

Since we decided that our VM can only put the start of our mapped data at a fixed consistent address, we would need to eg store the length of the mapped data at the fixed start of the region.

But once we are mucking around with bytes ourselves, we might as well do the whole outer layer by ourselves, since it's very simple.

There's some additional complexity required for directly storing rkyv-serialised data inside of rkyv in such a way that you can zero-copy deserialise both layers.  The extra complexity is mostly around the 16 byte alignment requirement.  That's why my original prototype used the `unaligned` feature.

It's totally doable, but then probably not really simpler than coding up this outer layer ourselves.

However, I'm open to using rkyv for the outer layer as well, if you think that's cleaner.  The code in here is just a prototype.

icemelon:
Could you add some example that showcases how `CenoStdin` can be used?

matthiasgoergens:
> Could you add some example that showcases how `CenoStdin` can be used?

Yes, I already added them. Have a look at the PR description for the link.

I can add some nob-standalone examples, too.

icemelon:
> > Could you add some example that showcases how `CenoStdin` can be used?
> 
> Yes, I already added them. Have a look at the PR description for the link.
> 
> I can add some nob-standalone examples, too.

Yeah, sorry missed that in the description. Why not include this example app in this PR?

ceno_emul/src/host_utils.rs
```diff
@@ -0,0 +1,32 @@
+use std::iter::from_fn;
+
+use crate::{ByteAddr, EmuContext, VMState, WordAddr};
+
+const WORD_SIZE: usize = 4;
+const INFO_OUT_ADDR: WordAddr = ByteAddr(0xC000_0000).waddr();
```
icemelon, line 6: is this a placeholder for now until we have the public IO?

matthiasgoergens: It's a placeholder of sorts, but for debug output, not for public IO.

We have an open issue to improve debug output.

matthiasgoergens:
> > > Could you add some example that showcases how `CenoStdin` can be used?
> > 
> > Yes, I already added them. Have a look at the PR description for the link.
> > 
> > I can add some non-standalone examples, too.
> 
> Yeah, sorry missed that in the description. Why not include this example app in this PR?

Yes, I will add a non-standalone example in this PR today. (I can't really add the standalone example in this PR, because it's supposed to be something that lives in a separate repository, to show that we can do that.)

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+        to_bytes::<Error>(item).map(|bytes| self.write_slice(bytes))
+    }
+
+    pub fn finalise(&self) -> Vec<u32> {
+        let initial_offset = (size_of::<u32>() * self.items.len()).next_multiple_of(RKYV_ALIGNMENT);
+        let offsets: Vec<u32> = self
+            .items
+            .iter()
+            .scan(initial_offset, |acc, bytes| {
+                let output = (*acc + bytes.len()) as u32;
```
icemelon, line 42: why you don't increase the `bytes.len()` up to multiple of `RKYV_ALIGNMENT`?

icemelon (commented):
A high-level comment or question. Should we restructure and rename the packages like having a `ceno_sdk` directory that encloses `ceno_emul`, `ceno_host`, and `ceno_guest`?

ceno_host/src/lib.rs
matthiasgoergens: Do you mean extending the buffer itself?  We could do that, but then we'd need to keep track of the original offsets (= lengths of the buffers) ourselves.

That's because rkyv wants its buffers to start at a 16 alignment, but it also wants the buffers to end exactly how they come out of its own `.to_bytes` function.  You can't just arbitrarily pad them to the right.

You can however pad them to the left just fine, as long as you keep the 16 byte alignment in mind.

matthiasgoergens:
> Yeah, sorry missed that in the description. Why not include this example app in this PR?

Done!

ceno_host/src/lib.rs
icemelon: My question is that in Line 43, the `acc` increases by the byte lens rounded to multiples of `RKYV_ALIGNMENT`. It seems to contradict the `output`, which calculates the offsets, right?

icemelon (commented):
We should still move `ceno_rt` to the top level directory, in parallel with `ceno_host`. The hint program example is better to put in the `$ROOT/examples` if possible.

matthiasgoergens:
> We should still move `ceno_rt` to the top level directory, in parallel with `ceno_host`. The hint program example is better to put in the `$ROOT/examples` if possible.

See https://github.com/scroll-tech/ceno/pull/754

ceno_emul/src/host_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+    .collect()
+}
+
+fn read_message(state: &VMState, offset: WordAddr) -> String {
+    let out_addr = INFO_OUT_ADDR + offset;
+    let byte_len = state.peek_memory(out_addr) as usize;
+
+    String::from_utf8_lossy(
+        &(out_addr + 1_usize..)
+            .map(|memory| state.peek_memory(memory))
```
lispc, line 26: `.map(|memory| state.peek_memory(memory))`: rename it to `address`?

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+            );
+        }
+        let (prefix, hints, postfix): (_, &[u32], _) = unsafe { buf.align_to() };
+        assert_eq!(prefix, &[]);
+        assert_eq!(postfix, &[]);
+        hints.to_vec()
+    }
+}
+
+pub fn run(platform: Platform, elf: &[u8], hints: &CenoStdin) -> Vec<String> {
```
lispc, line 69: not sure whether it is a good idea to use Vec<String> as "output" of program instead of raw Vec<u8>.

It is more friendly for debugging or human, but less flexiable for future changes.

anyway not bad for now.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    pub fn write(
+        &mut self,
+        item: &impl for<'a> Serialize<HighSerializer<AlignedVec, ArenaHandle<'a>, Error>>,
+    ) -> Result<(), Error> {
+        to_bytes::<Error>(item).map(|bytes| self.write_slice(bytes))
+    }
+
+    pub fn finalise(&self) -> Vec<u32> {
```
lispc, line 36: i think the codec format should be described briefly in doc string?

ceno_rt/src/mmio.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub fn read_slice<'a>() -> &'a [u8] {
+    unsafe {
+        let len: u32 = core::ptr::read(NEXT_HINT_LEN_AT as *const u32);
+        NEXT_HINT_LEN_AT = NEXT_HINT_LEN_AT.add(4);
+
+        let hints_region = {
+            let total_length = (&raw const _hints_end).offset_from(&_hints_start) as usize;
+            from_raw_parts(&_hints_start, total_length)
+        };
```
lispc, line 25: cannot understand here.. 

i guess `read_slice` will be called multi time, so `_hints_start` and `_hints_end` (or one of them) are like cursors, should not be const, they should be changed somewhere.. but i cannot find where they are changed..

lispc (approved):

lispc:
one requested change: https://github.com/scroll-tech/ceno/pull/631#discussion_r1886062232. and some other minor comments

ceno_emul/src/host_utils.rs
matthiasgoergens: Done.

ceno_rt/src/mmio.rs
matthiasgoergens: It's slightly more complicated.

In principle we could construct the slice directly like:
```rust
pub fn read_slice<'a>() -> &'a [u8] {
    unsafe {
        let len: u32 = core::ptr::read(NEXT_HINT_LEN_AT);
        NEXT_HINT_LEN_AT = NEXT_HINT_LEN_AT.add(4);
        &from_raw_parts(&_hints_start, len as usize)
    }
}
```

I hope that variant is easy to understand?

Why do we bother with the more complicated version?

Because `len` is a value that we get given by the prover, so we don't rust it!  So instead we construct the `hints_region`, which is constant and always the same.  Then we return a sub-slice via `&hints_region[..len as usize]`.  That subslicing makes sure to catch any shenanigans a prover might want to play.  (The one I can think of is trying to give a `len` longer than the hints region, but there might be others I'm unaware of.)

As you recognised we compute `hints_region` purely from statically know inputs, and it's always the same.  The only reason it's not declared static is because Rust won't let us.  (But I hope that the optimiser in LLVM is smart enough to recognise that it doesn't need to re-compute.)

Briefly talking about cursors:

You can think of `_hints_start` and `_hints_end` as constant cursors that define the `hints_region`.

`NEXT_HINT_LEN_AT` and `len` are like mutable cursors.  We use `NEXT_HINT_LEN_AT` to find the current value of `len`.

Hope that is clearer?  I'm working making a comment in the code to explain this, too.  (But with some small code changes to make the comment simpler.)

ceno_host/src/lib.rs
matthiasgoergens: This is only really meant for human consumption (and very simple test assertions) at the moment.  We have an issue to overhaul the system more properly.

ceno_host/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+// We want to get access to the default value of `AlignedVec::ALIGNMENT`, and using it directly like this
+//   pub const RKVY_ALIGNMENT: usize = rkyv::util::AlignedVec::ALIGNMENT;
+// doesn't work:
+pub const RKYV_ALIGNMENT: usize = {
+    type AlignedVec = rkyv::util::AlignedVec;
+    AlignedVec::ALIGNMENT
+};
+
+/// A structure for building the hints input to the Ceno emulator.
```
matthiasgoergens, line 23: @lispc Please see if the comment makes any sense to you.  Thanks!

ceno_rt/src/mmio.rs
```diff
@@ -0,0 +1,53 @@
...
+    static mut NEXT_HINT_LEN_AT: *const usize = &raw const _lengths_of_hints_start;
+    unsafe {
+        let len: usize = core::ptr::read(NEXT_HINT_LEN_AT);
+        NEXT_HINT_LEN_AT = NEXT_HINT_LEN_AT.add(1);
+        len
+    }
+}
+
+pub fn read_slice<'a>() -> &'a [u8] {
+    &hints_region()[..hint_len()]
```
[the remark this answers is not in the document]
matthiasgoergens: Please see the [comment](https://github.com/scroll-tech/ceno/pull/631/files#r1886274179) that's trying to explain this.

> say the hints_region() is static constant, be [2,3,4,5,67,14,65] (if we don't talk about alignment for simplicity).
> 
> lens is [2, 3,..]
> 
> 1st time, read_slice will return [2,3], first 2 elements.
> 2nd time, read_slice will return [4,5,67] or [2,3,4]?

It would return [2,3,4].

> my understanding is that it should return [4,5,67], but if hints_region() is the full hints bytes, it will return [2,3,4]?

Yes, exactly, it'll return [2, 3, 4] and that's as intended.

We always return from the very start of the hints region.  I added a comment in the code (linked above) to explain this:

> Our guest programs have two requirements on the format:
> 1. The start of the hints buffer consists of a sequence of `usize` values, each representing the
>    length of the next hint (from the start of the whole buffer).
> 2. `hints[..current_hint_len]` can deserialise into the expected type via rkyv.
> 
> Note how we overlap the two areas, and don't specify starts for our hints.  That's a simplification
> and performance improvement we can make because of how rkyv works: you can add arbitrary padding to
> the left of a serialised buffer, and it will still work.

Let me know if it makes sense.

PR #637, opened 2024-11-26
fix/program-size4: fix padding in edge cases

naure:
This bug was fixed then broken again in a cleanup attempt. 🤦 

Note that the whole topic is being refactored in #624 so hopefully we won’t have this issue anymore.

### DIFF
--- a/ceno_zkvm/src/instructions.rs
+++ b/ceno_zkvm/src/instructions.rs
@@ -73,7 +73,7 @@ pub trait Instruction<E: ExtensionField> {
             })
             .collect::<Result<(), ZKVMError>>()?;
 
-        let num_padding_instances = raw_witin.num_padding_instances();
+        let num_padding_instances = raw_witin.num_padding_instances(None);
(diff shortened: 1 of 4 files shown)


mcalancea (approved):

hero78119 (changes requested):
Sorry I disable auto merge because I was thinking to identify the root cause instead of quick patching 😃 .

question, does the number of instance size different with real padding just exist to deal with program table special case?

naure:
Yes at the moment this problem only exists the program table because it has a separate configuration of max size versus content. In other places (e.g. `NonVolatileTableConfig`) we just required an exact match.

Although you’re right that there is a more elegant fix which is to require a tight fit of the size. This wasn’t possible in the original fix because the size was not configurable yet. Let me try that.

matthiasgoergens:
Thanks for trying to fix this.

Could we please add a test that would fail if/when this break again?

naure:
Better not introduce this problem in the first place: #638

PR #644, opened 2024-11-27
fix/pow-table-index: Correct table index calculation

naure:
_Blocking sproll-evm._

PowTable content is organized differently than binary ops. This breaks the assignment of multiplicities. Fixed by overriding the mapping between table indexes and entries.

This detail was not covered by the table checks of MockProver, hence why discovered only now.

### DIFF
--- a/ceno_zkvm/src/tables/ops.rs
+++ b/ceno_zkvm/src/tables/ops.rs
@@ -113,7 +113,7 @@ mod tests {
     use goldilocks::{GoldilocksExt2 as E, SmallField};
 
     #[test]
-    fn test_ops_table_assign() {
+    fn test_ops_pow_table_assign() {
         let mut cs = ConstraintSystem::<E>::new(|| "riscv");
         let mut cb = CircuitBuilder::new(&mut cs);
 
--- a/ceno_zkvm/src/tables/ops.rs
+++ b/ceno_zkvm/src/tables/ops.rs
@@ -87,7 +87,49 @@ impl OpsTable for PowTable {
     }
 
     fn content() -> Vec<[u64; 3]> {
-        (0..Self::len() as u64).map(|b| [2, b, 1 << b]).collect()
+        (0..Self::len() as u64)
+            .map(|exponent| [2, exponent, 1 << exponent])
+            .collect()
+    }
+
+    fn pack(base: u64, exponent: u64) -> u64 {
+        assert_eq!(base, 2);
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/tables/ops.rs
```diff
@@ -87,7 +87,49 @@ impl OpsTable for PowTable {
     }
 
     fn content() -> Vec<[u64; 3]> {
-        (0..Self::len() as u64).map(|b| [2, b, 1 << b]).collect()
+        (0..Self::len() as u64)
+            .map(|exponent| [2, exponent, 1 << exponent])
```
hero78119, line 91: may I know what the different before/after, this logic looks same to me😮

ceno_zkvm/src/tables/ops.rs
```diff
@@ -87,7 +87,49 @@ impl OpsTable for PowTable {
...
+            .collect()
+    }
+
+    fn pack(base: u64, exponent: u64) -> u64 {
+        assert_eq!(base, 2);
+        exponent
+    }
+
+    fn unpack(exponent: u64) -> (u64, u64) {
```
hero78119, line 100: does pack & unpack only for test purpose? if yes probably we can define in test mod, or mark function with cfg[test]

naure: This is how `LkMultiplicity` finds the table index: https://github.com/scroll-tech/ceno/blob/846033ad805ae115507ed6923acf37e3ff478279/ceno_zkvm/src/witness.rs#L152

hero78119: I c, thanks! now I realized "override" means

hero78119 (approved):

ceno_zkvm/src/tables/ops.rs
```diff
@@ -87,7 +87,49 @@ impl OpsTable for PowTable {
...
+mod tests {
+    use super::*;
+    use crate::{
+        circuit_builder::{CircuitBuilder, ConstraintSystem},
+        tables::TableCircuit,
+    };
+    use goldilocks::{GoldilocksExt2 as E, SmallField};
+
+    #[test]
+    fn test_ops_table_assign() {
```
hero78119, line 116: one nitpick: I think this test only for pow_table, so probably `test_ops_pow_table_assign` more precisely

PR #646, opened 2024-11-27
fix tower sumcheck default logic with minimal table size 2

hero78119:
A follow-up PR after https://github.com/scroll-tech/ceno/pull/622

In #622 integration test was failed in first place. A proposed fix in comment https://github.com/scroll-tech/ceno/pull/622#discussion_r1855953467 is to skip the entire proof when table size should be 0. This fix is reasonable and fit the expectation, however, the integration test should NOT failed even BEFORE this fix. It should be garbage-in, garbage out and shouldn't break the proof.

### Root cause
The root cause is when table size is 1 (we will padding next size which is 2) or 2, in tower sumcheck there will be empty rounds. In tower sumcheck verifier, there are bookkeeping variables to tracking each layer (evaluation, point), and bookkeeping variables will keep updated when goes through layer by layer verification.  However when table size is 2, there is only one layer, so bookkeeping variables skip updated. However in this case, default value was wrong. The correct "default" value of bookkeeping variable should be the output layer evaluations. 

This PR correcting the default value design. Also add simple test to verify the logic.


naure (approved):
As explained the new variable `prod_spec_point_n_eval` holds the evaluation to return in case `max_num_variables=1` instead of default.

PR #649, opened 2024-11-28
refactor `assert_satisfied_full`

zemse:
- uses MockProver utils inside `assert_satisfied_full` which enables additional missing checks for things like assert zero expressions.
- removes duplicate code to check of multiplicity
- enables MockProver to check for multiplicity of all instances (previously only instance 0 was considered)

### DIFF
--- a/ceno_zkvm/src/circuit_builder.rs
+++ b/ceno_zkvm/src/circuit_builder.rs
@@ -115,13 +115,10 @@ pub struct ConstraintSystem<E: ExtensionField> {
     pub w_table_expressions_namespace_map: Vec<String>,
 
     /// lookup expression
-    // TODO remove lk_table and use lk instead
     pub lk_expressions: Vec<Expression<E>>,
+    pub lk_table_expressions: Vec<LogupTableExpression<E>>,
     pub lk_expressions_namespace_map: Vec<String>,
     pub lk_expressions_items_map: Vec<(ROMType, Vec<Expression<E>>)>,
-    pub lk_table_expressions: Vec<LogupTableExpression<E>>,
-    pub lk_table_expressions_namespace_map: Vec<String>,
-    pub lk_table_expressions_items_map: Vec<(ROMType, Vec<Expression<E>>)>,
 
     /// main constraints zero expression
     pub assert_zero_expressions: Vec<Expression<E>>,
@@ -166,11 +163,9 @@ impl<E: ExtensionField> ConstraintSystem<E> {
             w_table_expressions: vec![],
             w_table_expressions_namespace_map: vec![],
             lk_expressions: vec![],
+            lk_table_expressions: vec![],
             lk_expressions_namespace_map: vec![],
             lk_expressions_items_map: vec![],
-            lk_table_expressions: vec![],
-            lk_table_expressions_namespace_map: vec![],
-            lk_table_expressions_items_map: vec![],
             assert_zero_expressions: vec![],
             assert_zero_expressions_namespace_map: vec![],
             assert_zero_sumcheck_expressions: vec![],
@@ -329,10 +324,10 @@ impl<E: ExtensionField> ConstraintSystem<E> {
             table_len,
         });
         let path = self.ns.compute_path(name_fn().into());
-        self.lk_table_expressions_namespace_map.push(path);
+        self.lk_expressions_namespace_map.push(path);
         // Since lk_expression is RLC(record) and when we're debugging
         // it's helpful to recover the value of record itself.
-        self.lk_table_expressions_items_map.push((rom_type, record));
+        self.lk_expressions_items_map.push((rom_type, record));
 
         Ok(())
     }
--- a/ceno_zkvm/src/circuit_builder.rs
+++ b/ceno_zkvm/src/circuit_builder.rs
@@ -115,10 +115,13 @@ pub struct ConstraintSystem<E: ExtensionField> {
     pub w_table_expressions_namespace_map: Vec<String>,
 
     /// lookup expression
+    // TODO remove lk_table and use lk instead
     pub lk_expressions: Vec<Expression<E>>,
     pub lk_expressions_namespace_map: Vec<String>,
+    pub lk_expressions_items_map: Vec<(ROMType, Vec<Expression<E>>)>,
     pub lk_table_expressions: Vec<LogupTableExpression<E>>,
     pub lk_table_expressions_namespace_map: Vec<String>,
+    pub lk_table_expressions_items_map: Vec<(ROMType, Vec<Expression<E>>)>,
 
     /// main constraints zero expression
     pub assert_zero_expressions: Vec<Expression<E>>,
@@ -136,7 +139,6 @@ pub struct ConstraintSystem<E: ExtensionField> {
     pub chip_record_beta: Expression<E>,
 
     pub debug_map: HashMap<usize, Vec<Expression<E>>>,
-    pub lk_expressions_items_map: Vec<(ROMType, Vec<Expression<E>>)>,
 
     pub(crate) phantom: PhantomData<E>,
 }
@@ -165,8 +167,10 @@ impl<E: ExtensionField> ConstraintSystem<E> {
             w_table_expressions_namespace_map: vec![],
             lk_expressions: vec![],
             lk_expressions_namespace_map: vec![],
+            lk_expressions_items_map: vec![],
             lk_table_expressions: vec![],
             lk_table_expressions_namespace_map: vec![],
+            lk_table_expressions_items_map: vec![],
             assert_zero_expressions: vec![],
             assert_zero_expressions_namespace_map: vec![],
             assert_zero_sumcheck_expressions: vec![],
@@ -176,7 +180,6 @@ impl<E: ExtensionField> ConstraintSystem<E> {
             chip_record_beta: Expression::Challenge(1, 1, E::ONE, E::ZERO),
 
             debug_map: HashMap::new(),
-            lk_expressions_items_map: vec![],
 
             phantom: std::marker::PhantomData,
         }
@@ -329,7 +332,7 @@ impl<E: ExtensionField> ConstraintSystem<E> {
         self.lk_table_expressions_namespace_map.push(path);
         // Since lk_expression is RLC(record) and when we're debugging
         // it's helpful to recover the value of record itself.
-        self.lk_expressions_items_map.push((rom_type, record));
+        self.lk_table_expressions_items_map.push((rom_type, record));
 
         Ok(())
     }
--- a/ceno_zkvm/src/scheme/mock_prover.rs
+++ b/ceno_zkvm/src/scheme/mock_prover.rs
@@ -474,10 +474,11 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
         challenge: Option<[E; 2]>,
         lkm: Option<LkMultiplicity>,
     ) -> Result<(), Vec<MockProverError<E, u64>>> {
-        let program = Arc::new(Program::from_insn_codes(program));
+        let program = Arc::new(Program::from(program));
         let (table, challenge) = Self::load_tables_with_program(cb.cs, program, challenge);
 
-        Self::run_maybe_challenge_with_table(cb.cs, &table, wits_in, pi, 1, challenge, lkm, None)
+        Self::run_maybe_challenge_with_table(cb.cs, &table, wits_in, pi, 1, challenge, lkm)
+            .map(|_| ())
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -489,8 +490,8 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
         num_instances: usize,
         challenge: [E; 2],
         expected_lkm: Option<LkMultiplicity>,
-        mut shared_lkm: Option<&mut LkMultiplicityRaw<E>>,
-    ) -> Result<(), Vec<MockProverError<E, u64>>> {
+    ) -> Result<LkMultiplicityRaw<E>, Vec<MockProverError<E, u64>>> {
+        let mut shared_lkm = LkMultiplicityRaw::<E>::default();
         let mut errors = vec![];
 
         // Assert zero expressions
@@ -582,10 +583,8 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
             }
 
             // Increment shared LK Multiplicity
-            if let Some(shared_lkm) = shared_lkm.as_mut() {
-                for element in expr_evaluated {
-                    shared_lkm.increment(*rom_type, *element);
-                }
+            for element in expr_evaluated {
+                shared_lkm.increment(*rom_type, *element);
             }
         }
 
@@ -644,7 +643,7 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
         }
 
         if errors.is_empty() {
-            Ok(())
+            Ok(shared_lkm)
         } else {
             Err(errors)
         }
@@ -848,7 +847,7 @@ Hints:
                 let lkm_from_assignments = witnesses
                     .get_lk_mlt(circuit_name)
                     .map(LkMultiplicityRaw::deep_clone);
-                let errors = Self::run_maybe_challenge_with_table(
+                match Self::run_maybe_challenge_with_table(
                     cs,
                     &lookup_table,
                     &witness,
@@ -856,10 +855,10 @@ Hints:
                     num_rows,
                     challenges,
                     lkm_from_assignments,
-                    Some(&mut lkm_opcodes),
-                );
-                match errors {
-                    Ok(_) => {}
+                ) {
+                    Ok(multiplicities) => {
+                        lkm_opcodes += multiplicities;
+                    }
                     Err(errors) => {
                         tracing::error!("Mock proving failed for opcode {}", circuit_name);
                         print_errors(&errors, &witness, &cs.witin_namespace_map, true);
@@ -898,10 +897,11 @@ Hints:
                     .to_vec();
 
                     for (key, multiplicity) in izip!(lk_table, multiplicity) {
-                        let multiplicity = multiplicity.to_canonical_u64();
-                        if multiplicity > 0 {
-                            lkm_tables.set_count(*rom_type, key, multiplicity as usize);
-                        }
+                        lkm_tables.set_count(
+                            *rom_type,
+                            key,
+                            multiplicity.to_canonical_u64() as usize,
+                        );
                     }
                 }
             }
@@ -1188,7 +1188,7 @@ fn compare_lkm<E, K>(
 ) -> Vec<MockProverError<E, K>>
 where
     E: ExtensionField,
-    K: LkMultiplicityKey + Ord,
+    K: LkMultiplicityKey + Default + Ord,
 {
     let lkm_a = lkm_a.into_finalize_result();
     let lkm_b = lkm_b.into_finalize_result();
--- a/ceno_zkvm/src/scheme/mock_prover.rs
+++ b/ceno_zkvm/src/scheme/mock_prover.rs
@@ -573,8 +573,7 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
 
             // Check each lookup expr exists in t vec
             for (inst_id, element) in enumerate(expr_evaluated) {
-                let vec = element.to_canonical_u64_vec();
-                if !table.contains(&vec) {
+                if !table.contains(&element.to_canonical_u64_vec()) {
                     errors.push(MockProverError::LookupError {
                         rom_type: *rom_type,
                         expression: expr.clone(),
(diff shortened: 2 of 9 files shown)


ceno_zkvm/src/instructions.rs
```diff
@@ -47,31 +47,31 @@ pub trait Instruction<E: ExtensionField> {
...
-        raw_witin_iter
+        let lk_multiplicity = raw_witin_iter
             .zip(steps.par_chunks(num_instance_per_batch))
             .flat_map(|(instances, steps)| {
-                let mut lk_multiplicity = lk_multiplicity.clone();
                 instances
                     .chunks_mut(num_witin)
                     .zip(steps)
                     .map(|(instance, step)| {
+                        let mut lk_multiplicity = LkMultiplicity::default();
```
hero78119, line 68: This new design although idea is nice and capture lkm on row (per instance) level, however this design will bring quite lot of overhead because we will create #instance of lkm, comparing with original design we just create ~ #opcode + #opcode lkm which is < 50.

So I suggest we just rollback to single lkm version. In current assignment design, lkm check failed will imply on all row contributed to the failed, because we dont have per row conditional logic. With single lkm, we can still know which table failed, while retain the efficiency of the design

ceno_zkvm/src/instructions/riscv/arith_imm.rs
```diff
@@ -113,7 +113,8 @@ mod test {
         let insn_code = encode_rv32(InsnKind::ADDI, 2, 0, 4, imm_i(3));
         let (raw_witin, lkm) = AddiInstruction::<GoldilocksExt2>::assign_instances(
             &config,
-            cb.cs.num_witin as usize,
+            #[allow(clippy::explicit_auto_deref)]
+            (*cb.cs).num_witin as usize,
```
hero78119, line 117: this change seems unrelated and just optional, is it trigger any clippy warning?

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -543,94 +543,67 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
...
+                                        (expr.clone(), insts)
+                                    })
+                                    .collect::<Vec<(Expression<E>, Vec<u64>)>>(),
+                            )
                         })
-                    });
+                        .collect::<Vec<(ROMType, Vec<(Expression<E>, Vec<u64>)>)>>(),
+                );
+                // Remove the last instance, which is the padding instance.
+                let instances = result.len();
```
hero78119, line 571: The logic to retrieve padding instance is somehow incorrect. 

The correct way to retrieve number of padding row is via https://github.com/scroll-tech/ceno/blob/cec7b82c29f0d06041bf20aec3e3c0932e6643fa/ceno_zkvm/src/scheme/mock_prover.rs#L789

and https://github.com/scroll-tech/ceno/blob/cec7b82c29f0d06041bf20aec3e3c0932e6643fa/ceno_zkvm/src/scheme/mock_prover.rs#L829

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -392,7 +392,7 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
         cb: &CircuitBuilder<E>,
```
hero78119, line 392: I suggest in this PR, we can refactor duplicate part from another mock_prover util 
https://github.com/scroll-tech/ceno/blob/cec7b82c29f0d06041bf20aec3e3c0932e6643fa/ceno_zkvm/src/scheme/mock_prover.rs#L748

In summary, in `assert_satisfied_full`
- test on all opcode + table circuit
- output error per opcode/table
- per opcode/table circuit, it only evaluate expression values on errors happened, and just output first row raw value

Since `assert_satisfied_full` is like a super set of ALL opcode circuit of `run_maybe_challenge` for single opcode circuit. So we should extract common part to `run_maybe_challenge`, then `assert_satisfied_full` just call `run_maybe_challenge` per opcode. With that, we remove all the duplicated, while keep a good debugging ability

hero78119 (changes requested):

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -711,24 +712,25 @@ Hints:
...
         challenge: Option<[E; 2]>,
         lkm: Option<LkMultiplicity>,
     ) {
-        Self::assert_with_expected_errors(cb, wits_in, programs, &[], challenge, lkm);
+        Self::assert_with_expected_errors(cb, wits_in, program, &[], challenge, lkm);
     }
 
     pub fn assert_satisfied_full(
-        cs: &ZKVMConstraintSystem<E>,
+        cs: &mut ZKVMConstraintSystem<E>,
```
hero78119, line 729: it's not very intuive to me for why `cs` need to be mutable, does it nessesary?

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -861,70 +863,26 @@ Hints:
...
+                    &CircuitBuilder {
+                        cs: op_cs,
+                        params: cs.params.clone(),
+                    },
+                    witness,
+                    program.clone(),
+                    None,
+                    Some(witnesses.get_lk_mlt(name).unwrap()),
+                );
             }
```
hero78119, line 886: `rom_tables` checking logic are missing in new refactor.
We need to check all `opcode` rom_type multiplicity sum are matching with `table` rom_type multiplicity

ceno_zkvm/src/scheme/mock_prover.rs
zemse: I am trying to call `assert_satisfied_raw` inside `assert_satisfied_full`. Currently, `assert_satisfied_raw` takes `CircuitBuilder` struct which wants a `mut ConstraintSystem`.

```rs
Self::assert_satisfied_raw(
    &CircuitBuilder {
        cs: op_cs,
        params: cs.params.clone(),
    },
    ...
);
```

To remove the mutable reference, `assert_satisfied_raw` API need to be changed to take `ConstraintSystem` directly and would cause more changes everywhere.

hero78119: I c the reason, it's because `assert_satisfied_raw` api take `CircuitBuilder`.
But `assert_satisfied_full` take ConstainSystem which cause confuse to be mutable. I realized a small change to make it work

```diff
diff --git a/ceno_zkvm/src/scheme/mock_prover.rs b/ceno_zkvm/src/scheme/mock_prover.rs
index 75340435..f5fe7cda 100644
--- a/ceno_zkvm/src/scheme/mock_prover.rs
+++ b/ceno_zkvm/src/scheme/mock_prover.rs
@@ -726,7 +726,7 @@ Hints:
     }

     pub fn assert_satisfied_full(
-        cs: &mut ZKVMConstraintSystem<E>,
+        cs: &ZKVMConstraintSystem<E>,
         mut fixed_trace: ZKVMFixedTraces<E>,
         witnesses: &ZKVMWitnesses<E>,
         pi: &PublicValues<u32>,
@@ -864,7 +864,7 @@ Hints:
         }

         // assert all opcodes and check lk multiplicity
-        for (name, op_cs) in (cs.circuit_css).iter_mut() {
+        for (name, op_cs) in &cs.circuit_css {
             let is_opcode = op_cs.lk_table_expressions.is_empty()
                 && op_cs.r_table_expressions.is_empty()
                 && op_cs.w_table_expressions.is_empty();
@@ -875,7 +875,7 @@ Hints:
             if let Some(witness) = witnesses.get_table_witness(name) {
                 Self::assert_satisfied_raw(
                     &CircuitBuilder {
-                        cs: op_cs,
+                        cs: &mut op_cs.clone(),
                         params: cs.params.clone(),
                     },
                     witness,
```

zemse: I've resolved this by removing the requirement of CircuitBuilder as it is not necessary.

ceno_zkvm/src/scheme/mock_prover.rs
zemse: I've added the check.

ceno_zkvm/src/structs.rs
```diff
@@ -254,6 +254,10 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
         self.witnesses_tables.get(name).cloned()
     }
 
+    pub fn get_lk_mlt(&self, name: &String) -> Option<LkMultiplicity> {
```
hero78119, line 257: we should move `deep_clone()` logic to mock_prover and here should just return Option<&LkMultiplicity> in case of being abuse in production code in the future, since deep_clone is very heavy

ceno_zkvm/src/structs.rs
```diff
@@ -284,7 +288,17 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
         let mut combined_lk_mlt = vec![];
         let keys = self.lk_mlts.keys().cloned().collect_vec();
         for name in keys {
-            let lk_mlt = self.lk_mlts.remove(&name).unwrap().into_finalize_result();
+            let lk_mlt = if std::env::var("MOCK_PROVING").is_ok() {
```
hero78119, line 291: are we able to retain original logic and move this logic to mock_prover?
the goal is avoid introducing unnessesary debug condition check in production flow

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -404,58 +456,55 @@ impl<'a, E: ExtensionField + Hash> MockProver<E> {
...
 
+    #[allow(clippy::too_many_arguments)]
+    fn run_with_challenge_plain(
+        cs: &ConstraintSystem<E>,
+        table: &HashSet<Vec<u64>>,
+        wits_in: &[ArcMultilinearExtension<'a, E>],
+        pi: &[ArcMultilinearExtension<'a, E>],
+        num_instances: usize,
+        challenge: [E; 2],
+        lkm: Option<LkMultiplicity>,
```
hero78119, line 494: not related to this PR, but to differentiate with `shared_lkm`, how about rename `lkm` as `expected_lkm`

ceno_zkvm/src/circuit_builder.rs
```diff
@@ -136,7 +139,6 @@ pub struct ConstraintSystem<E: ExtensionField> {
     pub chip_record_beta: Expression<E>,
 
     pub debug_map: HashMap<usize, Vec<Expression<E>>>,
```
hero78119, line 141: - the above TODO seems unnessesary 
- we can remove `lk_table_expressions_items_map` because we dont extract each `record` of lk table in mock_prover

zemse: I've also removed the `lk_table_expressions_namespace_map` as `lk_expressions_namespace_map` has the same type.

ceno_zkvm/src/structs.rs
zemse: Unfortunately, this logic is called before invoking the mock prover, by that time `self.lk_mlts` is empty. So I don't think we can do that.

ceno_zkvm/src/scheme/mock_prover.rs
zemse: I've renamed it.

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -51,9 +52,27 @@ pub const MOCK_PC_START: ByteAddr = ByteAddr({
     CENO_PLATFORM.pc_base()
 });
 
+/// Allow LK Multiplicity's key to be used with `u64` and `GoldilocksExt2`.
+pub trait LkMultiplicityKey: Copy + Clone + Debug + Eq + Hash + Send {
+    /// If key is u64, return Some(u64), otherwise None.
```
matthiasgoergens, line 57: The other possibility is `GoldilocksExt2`?

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -1205,6 +1185,84 @@ Hints:
...
+) where
+    E: ExtensionField,
+    K: LkMultiplicityKey,
+{
+    let lkm_a = lkm_a.into_finalize_result();
+    let lkm_b = lkm_b.into_finalize_result();
+
+    // Compare each LK Multiplicity.
+    for (rom_type, a_map, b_map) in izip!(ROMType::iter(), &lkm_a, &lkm_b) {
+        if *a_map != *b_map {
```
matthiasgoergens, line 1201: We can simplify the logic here a lot:

```rust
fn compare_lkm<E, K>(
    lkm_a: LkMultiplicityRaw<K>,
    lkm_b: LkMultiplicityRaw<K>,
) -> Vec<MockProverError<E, K>>
where
    E: ExtensionField,
    K: LkMultiplicityKey + Ord,
{
    let lkm_a = lkm_a.into_finalize_result();
    let lkm_b = lkm_b.into_finalize_result();

    // Compare each LK Multiplicity.
    izip!(ROMType::iter(), &lkm_a, &lkm_b)
        .flat_map(|(rom_type, a_map, b_map)| {
            // We use a BTreeSet, instead of a HashSet, to ensure deterministic order.
            let keys: BTreeSet<_> = chain!(a_map.keys(), b_map.keys()).collect();
            keys.into_iter().filter_map(move |key| {
                let count =
                    *a_map.get(key).unwrap_or(&0) as isize - *b_map.get(key).unwrap_or(&0) as isize;

                (count != 0).then_some(MockProverError::LkMultiplicityError {
                    rom_type,
                    key: *key,
                    count,
                })
            })
        })
        .collect()
}
```

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -1205,6 +1185,84 @@ Hints:
...
+    for (count, error) in errors.iter().dedup_with_count() {
+        error.print(wits_in, wits_in_name);
+        if count > 1 {
+            println!("Error: {} duplicates hidden.", count - 1);
+        }
+    }
+    println!("Error: {} constraints not satisfied", errors.len());
+    println!("======================================================");
+    if panic_on_error {
+        panic!("(Unexpected) Constraints not satisfied");
```
matthiasgoergens, line 1262: If `panic_on_error` is set, we should probably put the full error message into the panic message?  (In that case, we don't need to println the error message, just put it in the panic only.)

ceno_zkvm/src/structs.rs
matthiasgoergens: Can't we change that?

ceno_zkvm/src/witness.rs
```diff
@@ -130,10 +131,59 @@ impl<F: Sync + Send + Copy> Index<usize> for RowMajorMatrix<F> {
...
 #[allow(clippy::type_complexity)]
-pub struct LkMultiplicity {
-    multiplicity: Arc<ThreadLocal<RefCell<[HashMap<u64, usize>; mem::variant_count::<ROMType>()]>>>,
+pub struct LkMultiplicityRaw<K: Copy + Clone + Debug + Eq + Hash + Send> {
+    multiplicity: Arc<ThreadLocal<RefCell<[HashMap<K, usize>; mem::variant_count::<ROMType>()]>>>,
+}
+
+impl<K: Copy + Clone + Debug + Eq + Hash + Send> LkMultiplicityRaw<K> {
+    /// Merge result from multiple thread local to single result.
+    pub fn into_finalize_result(self) -> [HashMap<K, usize>; mem::variant_count::<ROMType>()] {
```
matthiasgoergens, line 140: These are just for-loops at their core.  No need to pretend we are more functional than we are:

```rust
    /// Merge result from multiple thread local to single result.
    pub fn into_finalize_result(self) -> [HashMap<K, usize>; mem::variant_count::<ROMType>()] {
        let mut results = array::from_fn(|_| HashMap::new());
        for y in Arc::try_unwrap(self.multiplicity).unwrap() {
            for (result, addition) in izip!(&mut results, y.into_inner()) {
                for (key, value) in addition {
                    *result.entry(key).or_insert(0) += value;
                }
            }
        }
        results
    }
```

ceno_zkvm/src/witness.rs
```diff
@@ -130,10 +131,59 @@ impl<F: Sync + Send + Copy> Index<usize> for RowMajorMatrix<F> {
...
+    }
+
+    pub fn set_count(&mut self, rom_type: ROMType, key: K, count: usize) {
+        let multiplicity = self
+            .multiplicity
+            .get_or(|| RefCell::new(array::from_fn(|_| HashMap::new())));
+        multiplicity.borrow_mut()[rom_type as usize].insert(key, count);
+    }
+
+    /// Clone inner, expensive operation.
```
matthiasgoergens, line 170: This seems all too complicated, but let me think a bit more about it.  (We can merge first.)

matthiasgoergens (commented):
Please see https://github.com/scroll-tech/ceno/pull/793 for most of the suggestions that are also in the comments.

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -240,36 +277,49 @@ impl<E: ExtensionField> MockProverError<E> {
                 } else {
                     "Lookup".to_string()
                 };
-                let location = if *count > 0 {
-                    "constraint system"
+
+                let (location, element) = if let Some(key) = key.to_u64() {
+                    let location = if *count > 0 {
```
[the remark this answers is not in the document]
zemse: `count` is recorded to print errors in the following fashion:

```sh
# when you have Uint::new in construct while Value::new_unchecked in assign
LkMultiplicityError:
2 Lookups of U16 missing in assignments
Element: 0

# when you have Uint::new_unchecked in construct while Value::new in assign
LkMultiplicityError:
2 Lookups of U16 missing in constraint system
Element: 0
```

But I agree this could be more self explaining. Instead of count being a signed int, it could be an enum, something like-

```rs
enum LkMultiplicityErrorCount {
    MissingInAssignments(u64),
    MissingInConstraintSystem(u64),
}
```

ceno_zkvm/src/scheme/mock_prover.rs
zemse: Sometimes the errors can be a lot and it can go beyond the terminal lines limit (if it's set low) and the following default rust panic stderr line would not be easily visible or lost.

```
thread 'main' panicked at /Users/.../mock_prover.rs:919:9:
```

That's why the panic message was purposefully kept short summary so that when program terminates, it can be obvious it was a panic and at which line, and dev can look at logs before to learn more details about what caused "(Unexpected) Constraints not satisfied".

But if it doesn't seem helpful, for better code we can put everything in panic.

ceno_zkvm/src/scheme/mock_prover.rs
zemse: LkMultiplicity's key previously was just `u64`. From the e2e witness we also get case where we want key as `GoldilocksExt2`. To enable reuse of the same code in MockProver, the trait `LkMultiplicityKey` helps.

zemse:
@matthiasgoergens Thanks for review and suggestion PR!

> Can't we change that? `let lk_mlt = if std::env::var("MOCK_PROVING").is_ok() {`

That part of the code executes [before the mock prover is triggered in the e2e test](https://github.com/scroll-tech/ceno/blob/97237a1224f0cbce2c57cd975f2e58a83e504073/ceno_zkvm/src/e2e.rs#L455). Since by the time we are in mock prover, the `lk_mlt` data is deleted. Unless we add a similar condition before data is deleted, it's difficult to recover the data.

ceno_zkvm/src/structs.rs
zemse: We can prevent `self.lk_mlts` from being emptied.

---

Original behavior: consume `self.lk_mlts` (regardless of `MOCK_PROVING` env)

This PR behaviour: do not consume `self.lk_mlts` if `MOCK_PROVING=true` but can consume it if `MOCK_PROVING=false`.

Alternative behavuour to remove debug env stuff from that file: do not consume `self.lk_mlts`. But this might affect performance.

ceno_zkvm/src/scheme/mock_prover.rs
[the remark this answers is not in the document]
matthiasgoergens: Yes.  Perhaps something for the next PR?

hero78119 (approved):
LGTM now 👍  
In summary
- in `assert_satisfied_raw`, each opcode we extract raw value from rlc expression and checked with expected lookup multiplicity
- in zkvm, the lookup value was in rlc format and so as table, and we just do raw to raw comparison.

some lefted TODO after this PR
- have better naming on `assert_satisfied_raw` and `assert_satisfied_full`. E.g. `assert_satisfied_full` -> `assert_satisfied_zkvm`
- retain this debug message https://github.com/scroll-tech/ceno/blob/1d7818ec23c6857ee557532853489ffb23297aa4/ceno_zkvm/src/scheme/mock_prover.rs#L900-L930 => this break down raw value is very useful, e.g. once program table lookup failed we can quickly targeting on the opcode.

PR #653, opened 2024-11-29
[WIP] experiment devirgo-style sumcheck + FRI fold

hero78119:
An experiment to see how much gain we have on e2e latency

Also related to #568


hero78119:
Hi @chaosma @yczhangsjtu I found before/after this change, benchmark e2e remain the same 😂 
Few reason 
- I didn't apply complex threading model as [prover_v2](https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/prover_v2.rs#L88-L89) in ceno main prove part sumcheck, so in rayon, different thread might take different part of vector, so the cache hit rate is not good
- basefold openining overhead is small comparing with other part

Hope this PR can help on WHIR implementation, if you need any input or review on WHIR, I will be happy to learn that and review your implementation

hero78119:
Closed in favour of #891

PR #657, opened 2024-11-29
Add timing tree 

### LINKED ISSUE #655: Integrate plonky3-like timing tree
Right now we need to examine textual logs to understand the distribution of runtime. The flamechart layer currently in use has [several problems](https://github.com/scroll-tech/ceno/pull/572).

We would like something similar to what Plonky3 uses:
image(1)
This seems to be done using `tracing-forest`

mcalancea:
Fixes #655.

Integrates `tracing-forest` for its nice timing-tree summary.
In addition, I added a trick to control "granularity": spans have a `profiling_{LEVEL}` field which one can prune from the command line. For example, compare:

```
cargo run --release --package ceno_zkvm --bin e2e -- --profiling=1 --max-steps=10000 --platform=sp1 ceno_zkvm/examples/fibonacci.elf 

INFO     ZKVM_create_proof [ 790ms | 0.00% / 100.00% ]
INFO     ┝━ commit_to_fixed_commit [ 129µs | 0.02% ] profiling_1: true
INFO     ┝━ commit_to_traces [ 147ms | 18.62% ] profiling_1: true
INFO     ┕━ main_proofs [ 643ms | 81.36% ] profiling_1: true
```
to
```
cargo run --release --package ceno_zkvm --bin e2e -- --profiling=2 --max-steps=10000 --platform=sp1 ceno_zkvm/examples/fibonacci.elf 

INFO     ZKVM_create_proof [ 1.10s | 0.00% / 100.00% ]
INFO     ┝━ commit_to_fixed_commit [ 162µs | 0.01% ] profiling_1: true
INFO     ┝━ commit_to_traces [ 151ms | 0.02% / 13.78% ] profiling_1: true
INFO     │  ┝━ commit to iteration [ 7.66ms | 0.70% ] circuit_name: "ADD" | profiling_2: true
INFO     │  ┝━ commit to iteration [ 4.17ms | 0.38% ] circuit_name: "ADDI" | profiling_2: true
INFO     │  ┝━ commit to iteration [ 163ns | 0.00% ] circuit_name: "AND" | profiling_2: true
INFO     │  ┝━ commit to iteration [ 297µs | 0.03% ] circuit_name: "ANDI" | profiling_2: true
INFO     │  ┝━ commit to iteration [ 201µs | 0.02% ] circuit_name: "AUIPC" | profiling_2: true
[... more lines follow]
```

I haven't gone too far in assigning adequate levels to spans, I'm not sure what's relevant.

LE: I also set the `entered_span!` log level to TRACE to keep them silent outside the profiling option.

### DIFF
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -59,10 +59,6 @@ fn main() {
         args
     };
 
-    type E = GoldilocksExt2;
-    type Pcs = Basefold<GoldilocksExt2, BasefoldRSParams>;
-    type ExampleProgramTableCircuit<E> = ProgramTableCircuit<E>;
-
     // default filter
     let default_filter = EnvFilter::builder()
         .with_default_directive(LevelFilter::DEBUG.into())
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -111,8 +111,8 @@ fn main() {
         .with(fmt_layer)
         // if some profiling granularity is specified, use the profiling filter,
         // otherwise use the default
-        .with((args.profiling.is_some()).then(|| filter_by_profiling_level))
-        .with((args.profiling.is_none()).then(|| default_filter))
+        .with((args.profiling.is_some()).then_some(filter_by_profiling_level))
+        .with((args.profiling.is_none()).then_some(default_filter))
         .init();
 
     let elf_bytes = fs::read(&args.elf).expect("read elf file");
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -25,14 +25,9 @@ use std::{
     time::Instant,
 };
 use tracing::level_filters::LevelFilter;
-use tracing_flame::FlameLayer;
 use tracing_forest::ForestLayer;
 use tracing_subscriber::{
-    EnvFilter, Layer, Registry,
-    filter::{self, filter_fn},
-    fmt,
-    layer::SubscriberExt,
-    util::SubscriberInitExt,
+    EnvFilter, Registry, filter::filter_fn, fmt, layer::SubscriberExt, util::SubscriberInitExt,
 };
 use transcript::Transcript;
 
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -26,7 +26,14 @@ use std::{
 };
 use tracing::level_filters::LevelFilter;
 use tracing_flame::FlameLayer;
-use tracing_subscriber::{EnvFilter, Registry, fmt, layer::SubscriberExt};
+use tracing_forest::ForestLayer;
+use tracing_subscriber::{
+    EnvFilter, Layer, Registry,
+    filter::{self, filter_fn},
+    fmt,
+    layer::SubscriberExt,
+    util::SubscriberInitExt,
+};
 use transcript::Transcript;
 
 /// Prove the execution of a fixed RISC-V program.
@@ -39,6 +46,11 @@ struct Args {
     /// The maximum number of steps to execute the program.
     #[arg(short, long)]
     max_steps: Option<usize>,
+
+    // Profiling granularity.
+    // Setting any value restricts logs to profiling information
+    #[arg(long)]
+    profiling: Option<usize>,
 
     /// The preset configuration to use.
     #[arg(short, long, value_enum, default_value_t = Preset::Ceno)]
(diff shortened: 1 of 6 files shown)


ceno_zkvm/src/bin/e2e.rs
```diff
@@ -40,6 +42,11 @@ struct Args {
     #[arg(short, long)]
     max_steps: Option<usize>,
 
+    // Profiling granularity.
+    // Setting any value restricts logs to profiling information
```
matthiasgoergens, line 46: I think you might want to control log verbosity and profiling information independently?

(It's reasonable to default log verbosity to silent, if you turn on profiling, though.)

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -77,22 +84,36 @@ fn main() {
     type Pcs = Basefold<GoldilocksExt2, BasefoldRSParams>;
     type ExampleProgramTableCircuit<E> = ProgramTableCircuit<E>;
 
-    // set up logger
-    let (flame_layer, _guard) = FlameLayer::with_file("./tracing.folded").unwrap();
```
matthiasgoergens, line 81: It looks like you are removing the flame related stuff here.  Do we still use it elsewhere?

(If not, we can probably remove the relevant crates from the dependencies in the `Cargo.toml`?)

You might also want to augment the PR description to mention that you are removing the flame related stuff?

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -77,22 +84,36 @@ fn main() {
...
+    let default_filter = EnvFilter::builder()
+        .with_default_directive(LevelFilter::DEBUG.into())
+        .from_env_lossy();
+
+    // filter by profiling level;
+    // spans with level i contain the field "profiling_{i}"
+    // this restricts statistics to first (args.profiling) levels
+    let profiling_level = args.profiling.unwrap_or(1);
+    let filter_by_profiling_level = filter_fn(move |metadata| {
+        (1..)
```
matthiasgoergens, line 97: With something like this you don't need `.take(profiling_level)`?

```suggestion
        (1..=profiling_level)
```

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -77,22 +84,36 @@ fn main() {
...
+        .with_thread_ids(false)
+        .with_thread_names(false)
+        .without_time();
+
+    Registry::default()
+        .with(ForestLayer::default())
+        .with(fmt_layer)
+        // if some profiling granularity is specified, use the profiling filter,
+        // otherwise use the default
+        .with((args.profiling.is_some()).then_some(filter_by_profiling_level))
```
matthiasgoergens, line 114: The parens look a bit confusing here?

Btw, instead of `is_some().then_some()` you can probably just use `.map(|_| filter_by_profiling_level)`?

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -52,13 +52,14 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
...
     /// create proof for zkvm execution
-    #[tracing::instrument(skip_all, name = "ZKVM_create_proof")]
+    #[tracing::instrument(skip_all, name = "ZKVM_create_proof", fields(profiling_1))]
     pub fn create_proof(
         &self,
         witnesses: ZKVMWitnesses<E>,
         pi: PublicValues<u32>,
         mut transcript: Transcript<E>,
     ) -> Result<ZKVMProof<E, PCS>, ZKVMError> {
+        let span = entered_span!("commit_to_fixed_commit", profiling_1 = true);
```
matthiasgoergens, line 62: Instead of pairs of and `entered_span` and `exit_span` like this

```rust
let span = entered_span(some_name);
// [... some code ...]
exit_span!(span);
```
in most cases we can probably use a less error-prone mechanism:

```rust
with_span!(some_name) {
  // [... some code ...]
}
```

(That requires to quickly write a `with_span` macro.  The exact syntax is only a suggestion.)

matthiasgoergens (commented):
Looks mostly good.  Please see comments.

matthiasgoergens:
Btw, does this produce wallclock time or report CPU seconds spent?

ceno_zkvm/src/bin/e2e.rs
mcalancea: I reverted the change regarding the flamegraph. I think it's not too useful right now, but it is outside the scope of this PR.

ceno_zkvm/src/bin/e2e.rs
mcalancea: Removed parens. `.map()` is terser, but not as readable in this case I'd say.

ceno_zkvm/src/scheme/prover.rs
mcalancea: I get your point, but I dislike the effect on indentation. Often, `[... some code...]` is not an expression, but a very large block. Because it's not a clear-cut decision, I would defer it to a separate issue/PR.

mcalancea:
It's wall time. It uses https://doc.rust-lang.org/std/time/struct.Instant.html and `elapsed`.

ceno_zkvm/src/bin/e2e.rs
mcalancea: Ideally you would, but the only way to keep the profiling data readable is to filter out some of the spans. And filtering out some of the spans has an unclear impact (to me at least) on the general logging context. I think it's simpler and not that inconvenient to keep them separate.

hero78119 (approved):
LGTM! it's very useful especially on level field as a good alternative of (buggy) flamegraph.

ceno_zkvm/src/scheme/prover.rs
matthiasgoergens: Yes, we can do that in a separate PR.

hero78119:
There might be profiling accuracy problem and I reported here first, run on `ceno server`
+ @kunxian-xia 

```command
cargo run --release --package ceno_zkvm --bin e2e -- --profiling=1 --max-steps=1048576 --platform=sp1 ceno_zkvm/examples/fibonacci.elf
```

```
INFO     ZKVM_create_proof [ 17.1s | 0.00% / 100.00% ]
INFO     - commit_to_fixed_commit [ 115ms | 0.00% ] profiling_1: true
INFO     - commit_to_traces [ 2.59s | 15.15% ] profiling_1: true
INFO     - main_proofs [ 14.5s | 84.85% ] profiling_1: true
```

hero78119:
Hi @mcalancea 

With command `RUST_LOG="debug" cargo run --release --package ceno_zkvm --bin e2e -- --max-steps=1048576 --platform=sp1 ceno_zkvm/examples/fibonacci.elf`

by default all the tracing span message will goes to stdout which make results unreadable.

Are we able to enable a filter setting from command to keep only "tracing::debug" output and excluding time span output?

PR #661, opened 2024-12-02
Update Cargo.toml

matthiasgoergens:


matthiasgoergens:
@spherel says that the current status is actually better than the PR.

PR #660, opened 2024-12-02
Refactor/using field operations on MUL to reduce witness

KimiWu123:
This PR was inspired by the implementation of `MULH`, which computes the result directly in the field. This methodology is applied to the `MUL` opcode to reduce the witness size.

lookups from 13 --> 11
witnesses from 28 --> 25

```
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| opcode_name   | num_instances | lookups | reads | witnesses | writes | 0_expr_deg | 0_expr_sumcheck_deg |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
- current
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| MUL           | 0             | 13      | 4     | 28        | 4      | [1: 5]     | [2: 2]              |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
- after
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| MUL           | 0             | 11      | 4     | 25        | 4      | [1: 3]     | [2: 2]              |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
```

PR #663, opened 2024-12-02
Shift directly

matthiasgoergens:


ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -307,8 +307,7 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for MulhInstructionBas
                 let rs2_s = rs2_signed.assign_instance(instance, lk_multiplicity, &rs2_val)?;
                 rd_signed.assign_instance(instance, lk_multiplicity, &rd_val)?;
```
matthiasgoergens, line 308: I don't know why we are bothering with signs here: RiscV only has one MUL instruction for both signed and unsigned values, just like the only have one ADD and one SUB instruction, because the bit patterns are the same.

ceno_zkvm/src/instructions/riscv/mulh.rs
```diff
@@ -296,49 +296,39 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for MulhInstructionBas
...
-                ((prod as u64) % (1u64 << BIT_WIDTH)) as u32
             }
             MulhSignDependencies::UU { constrain_rd } => {
                 // assign nonzero value (u32::MAX - rd)
                 let rd_f = E::BaseField::from(rd as u64);
                 let avoid_f = E::BaseField::from((1u64 << BIT_WIDTH) - 1);
                 constrain_rd.assign_instance(instance, rd_f, avoid_f)?;
-
-                let prod = rs1_val.as_u64() * rs2_val.as_u64();
-                (prod % (1u64 << BIT_WIDTH)) as u32
```
matthiasgoergens, line 320: Turns out, we don't actually need all the shifting.

matthiasgoergens:
Most of this is now in https://github.com/scroll-tech/ceno/pull/660

So this PR served its purpose of showing what's possible.

PR #666, opened 2024-12-02
Extract end-to-end logic from `bin/`

mcalancea:
Refactors code such that main logic of end-to-end execution moves to `src/` and CLI handling remains in `bin/`.


naure (commented):
That makes sense.

naure (approved):

PR #669, opened 2024-12-03
cleanup verbose log

hero78119:
With RUST_LOG="debug" by default there will log many verbose message which make result unreadable.
Most of debug log are left over to debug in https://github.com/scroll-tech/ceno/pull/368

As PR 368 already done, this PR is aim for debug log cleanup


matthiasgoergens (approved):

hero78119:
PR forcely updated to fix one clippy and synced with master

PR #672, opened 2024-12-03
Simplify tracing-span macros

matthiasgoergens:
Just something that came up when I tried to understand how https://github.com/scroll-tech/ceno/pull/657 works.

### DIFF
--- a/sumcheck/src/lib.rs
+++ b/sumcheck/src/lib.rs
@@ -1,7 +1,8 @@
 #![deny(clippy::cargo)]
+#![feature(decl_macro)]
 #[cfg(feature = "non_pow2_rayon_thread")]
 pub mod local_thread_pool;
-mod macros;
+pub mod macros;
 mod prover;
 mod prover_v2;
 pub mod structs;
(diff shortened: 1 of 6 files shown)


sumcheck/src/lib.rs
```diff
@@ -1,7 +1,8 @@
 #![deny(clippy::cargo)]
+#![feature(decl_macro)]
 #[cfg(feature = "non_pow2_rayon_thread")]
 pub mod local_thread_pool;
-mod macros;
+pub mod macros;
```
matthiasgoergens, line 5: Now this is explicitly marked as public.  We actually already used these macros across a different crate.

sumcheck/src/prover.rs
```diff
@@ -16,7 +16,7 @@ use transcript::{Challenge, Transcript, TranscriptSyncronized};
 use crate::local_thread_pool::{LOCAL_THREAD_POOL, create_local_pool_once};
 
 use crate::{
-    entered_span, exit_span,
+    macros::{entered_span, exit_span},
```
matthiasgoergens, line 19: We could also export these macros on the level of the crate, but leaving them in their own macros sub-module seems cleaner.

mcalancea (approved):

PR #678, opened 2024-12-03
remove tracing-flame

mcalancea:
Removes the `tracing-flame` dependency. 

As discussed [here](https://github.com/scroll-tech/ceno/pull/572) the flamegraphs generated via `tracing-flame` suffered from at least two problems:
- Unexplainable time gaps which made the graph hard to read/misleading. Possibly a bug in the crate itself.
- Collapsing all spans with the same name (but different fields) into one column, making it impossible to distinguish particular iterations in loops (such as particular circuits during construction).

In addition, the generation of  `tracing.folded` itself can take significant time and distort profiling for short executions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,7 +289,6 @@ dependencies = [
  "tempfile",
  "thread_local",
  "tracing",
- "tracing-flame",
  "tracing-forest",
  "tracing-subscriber",
  "transcript",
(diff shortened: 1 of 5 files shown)


hero78119 (approved):
LGTM! Thanks for quick addressed the issue.
Also verified this PR in ceno server and there is no tracing.fold log anymore 👍

Issue #630, opened 2024-11-26
Make the emulator match the circuits


matthiasgoergens:
Programming is hard.  Programming in an emulator like ours without proper debugging tools is even harder.  Programming without any debugging tools at all, just by 'proofed failed / succeeded' is hardest.

At the moment, our emulator and circuits don's implement the same machine.  So you might do a lot of debugging to barely get your code working in the emulator, but then it still fails proving.

One example: if your guest program accesses random uninitialised memory in the emulator, you get a 0.  But in the prover you might get a 0, or your program might fail, depending on the address and the phase of the moon.  (Eg some parts of the private IO space behave like that.)

The circuits have their limitations (mostly!) for good technical reasons.  So let's make the emulator reflect the limitations of the circuits.

The goal is that ideally, if your execution works in the emulator, you should be able to prove it.

_Originally posted by @matthiasgoergens in https://github.com/scroll-tech/ceno/pull/622#discussion_r1855845216_

lispc:
in this function,  we assert adjacent memory access value should be same https://github.com/scroll-tech/zkevm-circuits/blob/de04c6668125b732b47668d3bcd04723125762eb/zkevm-circuits/src/witness/rw.rs#L55

matthiasgoergens:
@lispc Thanks, I'll have a look.  Could you give more context, please?

PR #688, opened 2024-12-04
Match emulator to circuits: use a Harvard Architecture

matthiasgoergens:
Our circuits use a [modified](Modified_Harvard_architecture) [Harvard architecture](https://en.wikipedia.org/wiki/Harvard_architecture) where the memory space for instructions and data are completely separated, even though they are both addressed with 32 bits.  That is a Good Thing.

That means writing to the data cell with address x does not change the value of the instruction cell with address x.

In contrast, our emulator uses a [von Neumann architecture](https://en.wikipedia.org/wiki/Von_Neumann_architecture) at its core, but goes through quite a few gymnastics to hide that fact.

> In a von Neumann architecture, data memory and stored program memory are intermixed.  Modern computers have moved away from this architecture for [security](https://en.wikipedia.org/wiki/W%5EX) and performance reasons.

In this PR, we turn our emulator into a Harvard Architecture machine as well to match what our circuits are doing.  That means we change the emulator to explicitly load instructions not from the data RAM, but from the program ROM, which we are already storing separately anyway.

> Instead of thinking about separate address spaces for data and code, you can also imagine that we have an [instruction cache and a data cache](https://en.wikipedia.org/wiki/Modified_Harvard_architecture#Split-cache_(or_almost-von-Neumann)_architecture), and that we load the instructions into their cache once at the start of the program, and then never invalidate that cache.

This works towards https://github.com/scroll-tech/ceno/issues/630


icemelon:
Since you change the emulator to Harvard Architecture, should you also update the memory address space [here](https://github.com/scroll-tech/ceno/blob/master/ceno_emul/src/platform.rs#L21-L28)?

icemelon:
Is this PR ready for review? I see there're many todos left

matthiasgoergens:
> Since you change the emulator to Harvard Architecture, should you also update the memory address space [here](https://github.com/scroll-tech/ceno/blob/master/ceno_emul/src/platform.rs#L21-L28)?

Well, we could just totally remove the whole ROM section, because there's no problem with executing code anywhere, now that we separated the address spaces.  But I want to keep this PR as small as possible, and there's also no immediate problem with keeping the ROM section contained, either.  Or am I missing something?

> Is this PR ready for review? I see there're many todos left

The TODOs are for future work, I want to keep this PR small.  (I can convert them into issues, too, if you think that's better, but for these small organisational items I prefer to have them discoverable in the code.)

matthiasgoergens:
@icemelon Let me make the follow up PR quickly that addresses the TODOs.  Think of this PR as pre-emptively extracted from the big PR the TODOs hint at.  Smaller PRs are easier to review and get consensus on.

lispc (approved):
lgtm

icemelon:
> there's also no immediate problem with keeping the ROM section contained, either.
@matthiasgoergens Yes, you're right. I was thinking of having different memory space for ROM and RAM. But it's not necessary.

lispc:
> writing to the data cell with address x does not change the value of the instruction cell with address x

I think this pr finishes what it claims to do "writing to the data cell with address x does not change the value of the instruction cell with address x,  so make emulator be consistent with prover".

the TODOs are good refactor, but not directly part of this purpose.  So i think it is ok now.

matthiasgoergens:
> I don't mind making PR small. But why not finish the implementation of the Hardvard architecture emulator in this PR? I don't think it will make the pr too large. But I'll leave this up to @lispc

This PR finishes the Harvard architecture.  The TODOs are for cleanup enabled by the switch.

PR #689, opened 2024-12-04
Only use forest layer when profiling

mcalancea:
Forest layer currently clashes with `fmt_layer`. This PRs disables forest layer for non-profiling use-cases, which is fine because profiling visualization was the only reason we introduced it.

### DIFF
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -80,7 +80,7 @@ fn main() {
         .without_time();
 
     Registry::default()
-        .with(ForestLayer::default())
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/bin/e2e.rs
```diff
@@ -80,7 +80,7 @@ fn main() {
         .without_time();
 
     Registry::default()
-        .with(ForestLayer::default())
+        .with(args.profiling.is_some().then_some(ForestLayer::default()))
```
matthiasgoergens, line 83: Just use `map`?

matthiasgoergens (approved):
Thanks for writing a good description for the PR.

PR #690, opened 2024-12-04
add fibonacchi bench

hero78119:
### Context
Existing benchmark `riscv_add` are just for testing single opcode performance. This PR aim to have a first e2e bench which is representative to a workload in real world.

### What has been cover
- add fibonacci bench with sp1 platform hardcode
- refactor e2e flow into sub-function to maximize code sharing

### what's not being covered
Hey @mcalancea, key generation + witness assignment are still encaptulated in same function and not including in bench latency meature. Break down key generation and extract witness assignment will expose massive intermediate variables in between, so probably need another refactor and design to support #624

### DIFF
--- a/ceno_zkvm/benches/fibonacci.rs
+++ b/ceno_zkvm/benches/fibonacci.rs
@@ -1,0 +1,87 @@
+use std::{
+    fs,
+    path::PathBuf,
+    time::{Duration, Instant},
+};
+
+use ceno_emul::{CENO_PLATFORM, Platform, Program, WORD_SIZE};
+use ceno_zkvm::{
+    self,
+    e2e::{run_e2e_gen_witness, run_e2e_proof},
+};
+use criterion::*;
+
+use goldilocks::GoldilocksExt2;
+use mpcs::BasefoldDefault;
+
+criterion_group! {
+  name = op_add;
+  config = Criterion::default().warm_up_time(Duration::from_millis(20000));
+  targets = bench_e2e
+}
+
+criterion_main!(op_add);
+
+const NUM_SAMPLES: usize = 10;
+
+fn bench_e2e(c: &mut Criterion) {
+    type Pcs = BasefoldDefault<E>;
+    let mut file_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    file_path.push("examples/fibonacci.elf");
+    let stack_size = 32768;
+    let heap_size = 2097152;
+    let elf_bytes = fs::read(&file_path).expect("read elf file");
+    let program = Program::load_elf(&elf_bytes, u32::MAX).unwrap();
+
+    // use sp1 platform
+    let platform = Platform {
+        // The stack section is not mentioned in ELF headers, so we repeat the constant STACK_TOP here.
+        stack_top: 0x0020_0400,
+        rom: program.base_address
+            ..program.base_address + (program.instructions.len() * WORD_SIZE) as u32,
+        ram: 0x0010_0000..0xFFFF_0000,
+        unsafe_ecall_nop: true,
+        ..CENO_PLATFORM
+    };
+
+    for max_steps in [1usize << 20, 1usize << 21, 1usize << 22] {
+        // expand more input size once runtime is acceptable
+        let mut group = c.benchmark_group(format!("fibonacci_max_steps_{}", max_steps));
+        group.sample_size(NUM_SAMPLES);
+
+        // Benchmark the proving time
+        group.bench_function(
+            BenchmarkId::new(
+                "prove_fibonacci",
+                format!("fibonacci_max_steps_{}", max_steps),
+            ),
+            |b| {
+                b.iter_with_setup(
+                    || {
+                        run_e2e_gen_witness::<E, Pcs>(
+                            program.clone(),
+                            platform.clone(),
+                            stack_size,
+                            heap_size,
+                            vec![],
+                            max_steps,
+                        )
+                    },
+                    |(prover, _, zkvm_witness, pi, _, _, _)| {
+                        let timer = Instant::now();
+                        let _ = run_e2e_proof(prover, zkvm_witness, pi);
+                        println!(
+                            "Fibonacci::create_proof, max_steps = {}, time = {}",
+                            max_steps,
+                            timer.elapsed().as_secs_f64()
+                        );
+                    },
+                );
+            },
+        );
+
+        group.finish();
+    }
+
+    type E = GoldilocksExt2;
+}
--- a/ceno_zkvm/benches/fibonacci.rs
+++ b/ceno_zkvm/benches/fibonacci.rs
@@ -15,12 +15,12 @@ use goldilocks::GoldilocksExt2;
 use mpcs::BasefoldDefault;
 
 criterion_group! {
-  name = op_add;
+  name = fibonacci;
   config = Criterion::default().warm_up_time(Duration::from_millis(20000));
   targets = bench_e2e
 }
 
-criterion_main!(op_add);
+criterion_main!(fibonacci);
 
 const NUM_SAMPLES: usize = 10;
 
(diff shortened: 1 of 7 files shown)


ceno_zkvm/benches/fibonacci.rs
```diff
@@ -0,0 +1,87 @@
...
+    self,
+    e2e::{run_e2e_gen_witness, run_e2e_proof},
+};
+use criterion::*;
+
+use goldilocks::GoldilocksExt2;
+use mpcs::BasefoldDefault;
+
+criterion_group! {
+  name = op_add;
```
mcalancea, line 18: Needs renaming.

ceno_zkvm/benches/fibonacci.rs
```diff
@@ -0,0 +1,87 @@
...
+        // The stack section is not mentioned in ELF headers, so we repeat the constant STACK_TOP here.
+        stack_top: 0x0020_0400,
+        rom: program.base_address
+            ..program.base_address + (program.instructions.len() * WORD_SIZE) as u32,
+        ram: 0x0010_0000..0xFFFF_0000,
+        unsafe_ecall_nop: true,
+        ..CENO_PLATFORM
+    };
+
+    for max_steps in [1usize << 20, 1usize << 21, 1usize << 22] {
```
mcalancea, line 47: Do you think it makes sense to have a smaller number like `1 << 10` as well? It works much faster and it can be a good sanity check. 

By the way, do you know a simple way to not run for all three sizes every time? Or even to pass size from the command-line?

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -131,14 +135,77 @@ fn main() {
...
+
+    let (prover, verifier, zkvm_witness, pi, cycle_num, e2e_start, exit_code) =
+        run_e2e_gen_witness::<E, Pcs>(
+            program,
+            platform,
+            args.stack_size,
+            args.heap_size,
+            hints,
+            max_steps,
+        );
```
mcalancea, line 150: I see you've moved the high-level logic back to `bin/`. Is this temporary or do you want it to stay this way?

mcalancea (approved):
It looks alright to me! 

I added an issue tracking the general task [here](https://github.com/scroll-tech/ceno/issues/691). I have work in progress [here](https://github.com/scroll-tech/ceno/tree/refactor/breakdown-e2e), increasing granularity.

ceno_zkvm/src/bin/e2e.rs
hero78119: Most of the logic stay in src/e2e.rs. bin/e2e.rs will call sub step function.
Here I move 2 part of logic
- timing statistics print out, since this information are across different sub function
- sanity check for verifier with malicious public input. This part more like a sanity check, thus it's not very general to stay in src/e2e.rs. Besides, to make generics type E, PCS work with `panic::catch_unwind` we need to add quite of trait bound. Above 2 reason makes the decision to stay in high level e2e :)

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -131,14 +135,77 @@ fn main() {
...
+        witgen_time,
+        cycle_num as f64 / witgen_time / 1000.0,
+        e2e_time,
+        cycle_num as f64 / e2e_time / 1000.0,
+        rayon::current_num_threads()
     );
+
+    run_e2e_verify(&verifier, zkvm_proof.clone(), exit_code, max_steps);
+
+    // do sanity check
```
hero78119, line 174: Below are sanity check logic, which might not be general in other use caee :)

ceno_zkvm/src/bin/e2e.rs
mcalancea: Makes sense :+1:

ceno_zkvm/benches/fibonacci.rs
hero78119: For "1 << 20" per iteration it take ~ 20 secs to warm up (on ceno server), so for sanity check purpose it's also fine without wait too long time. (most of the time are on key step actually, 1 << 20 proof time is < 5 secs)

For run with custom start/end, I tried to have Args + pass from CLI, but unfortunately the argument parsing will conflicts with `Criteria` crate, both complain unexpected argument, atm I dont have idea, will try look back to it later :)

ceno_zkvm/benches/fibonacci.rs
hero78119: nice catch and fixed https://github.com/scroll-tech/ceno/pull/690/commits/19b6164b62d09cc2ca094e21c82d5c3b5cfd6183

ceno_zkvm/benches/fibonacci.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+criterion_main!(fibonacci);
+
+const NUM_SAMPLES: usize = 10;
+
+fn bench_e2e(c: &mut Criterion) {
+    type Pcs = BasefoldDefault<E>;
+    let mut file_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    file_path.push("examples/fibonacci.elf");
```
matthiasgoergens, line 30: Please see how `test_elf.rs` builds and reads in the examples.

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -131,14 +135,77 @@ fn main() {
...
+            max_steps,
+        );
+
+    let timer = Instant::now();
+    let mut zkvm_proof = run_e2e_proof(prover, zkvm_witness, pi);
+    let proving_time = timer.elapsed().as_secs_f64();
+    let e2e_time = e2e_start.elapsed().as_secs_f64();
+    let witgen_time = e2e_time - proving_time;
+    println!(
+        "Proving finished.\n\
```
matthiasgoergens, line 158: You can just use 'natural' newlines?

```suggestion
        "Proving finished.
```

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -131,14 +135,77 @@ fn main() {
...
+        cycle_num as f64 / e2e_time / 1000.0,
+        rayon::current_num_threads()
     );
+
+    run_e2e_verify(&verifier, zkvm_proof.clone(), exit_code, max_steps);
+
+    // do sanity check
+    let transcript = Transcript::new(b"riscv");
+    // change public input maliciously should cause verifier to reject proof
+    zkvm_proof.raw_pi[0] = vec![<E as ff_ext::ExtensionField>::BaseField::ONE];
```
matthiasgoergens, line 177: That should be easier to read:

```suggestion
    zkvm_proof.raw_pi[0] = vec![Goldilocks::ONE];
```

(You need to import Goldilocks.)

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -131,14 +135,77 @@ fn main() {
...
+    zkvm_proof.raw_pi[1] = vec![<E as ff_ext::ExtensionField>::BaseField::ONE];
+
+    // capture panic message, if have
```
matthiasgoergens, line 180: Ah, I recognise this code from elsewhere.  The whole file is mostly copy-and-pasted, isn't it?

Please consider less duplication.

matthiasgoergens: See https://github.com/scroll-tech/ceno/pull/603/files for a cleanup that's already in the works.  (Feel free to take it over, or clean it up yourself; but making copies of the old code will make it harder to clean up in multiple places.)

ceno_zkvm/benches/fibonacci.rs
hero78119: Build and read elf is outside of the scope. there are other ongoing task for build fibonacci elf.

ceno_zkvm/src/bin/e2e.rs
hero78119: Indeed it improve readibility, fixed in https://github.com/scroll-tech/ceno/pull/690/commits

ceno_zkvm/src/bin/e2e.rs
hero78119: PR 603 `with_panic_hook` idea is worthy and improve on readibilty. I refactored two place to share this utility and fix in https://github.com/scroll-tech/ceno/commit/dc1f24515dcffb42c833704297ac42190923cfba

PR #692, opened 2024-12-04
Turn Transcript into trait

noel2004:
This PR try to turn `Transcript` into a trait so we can apply different implenents (For example, the `TranscriptSyncronized`) in the future.

We try to split the trait into a basic `Transcript` trait and a extended `Forkable` (with the `fork` method being provided) one.
The original `Transcript` struct has been renamed into `BasicTranscript`.

### DIFF
--- a/transcript/src/basic.rs
+++ b/transcript/src/basic.rs
@@ -5,64 +5,45 @@ use poseidon::poseidon_permutation::PoseidonPermutation;
 
 use crate::Challenge;
 
-#[derive(Clone)]
-pub struct Transcript<E: ExtensionField> {
-    permutation: PoseidonPermutation<E::BaseField>,
-}
-
-impl<E: ExtensionField> Transcript<E> {
-    /// Create a new IOP transcript.
-    pub fn new(label: &'static [u8]) -> Self {
-        let mut perm = PoseidonPermutation::new(core::iter::repeat(E::BaseField::ZERO));
-        let label_f = E::BaseField::bytes_to_field_elements(label);
-        perm.set_from_slice(label_f.as_slice(), 0);
-        perm.permute();
-        Self { permutation: perm }
-    }
-}
-
-impl<E: ExtensionField> Transcript<E> {
-    /// Fork this transcript into n different threads.
-    pub fn fork(self, n: usize) -> Vec<Self> {
-        let mut forks = Vec::with_capacity(n);
-        for i in 0..n {
-            let mut fork = self.clone();
-            fork.append_field_element(&(i as u64).into());
-            forks.push(fork);
+/// The Transcript trait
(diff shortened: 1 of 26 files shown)


transcript/src/basic.rs
```diff
@@ -5,64 +5,45 @@ use poseidon::poseidon_permutation::PoseidonPermutation;
...
-
-impl<E: ExtensionField> Transcript<E> {
-    /// Create a new IOP transcript.
-    pub fn new(label: &'static [u8]) -> Self {
-        let mut perm = PoseidonPermutation::new(core::iter::repeat(E::BaseField::ZERO));
-        let label_f = E::BaseField::bytes_to_field_elements(label);
-        perm.set_from_slice(label_f.as_slice(), 0);
-        perm.permute();
-        Self { permutation: perm }
+/// The Transcript trait
```
icemelon, line 8: Maybe move the trait to `lib.rs` or a separate file like `common.rs`?

noel2004: Done. Traits have been moved into `lib.rs`

transcript/src/lib.rs
```diff
@@ -14,3 +14,96 @@ mod hasher;
...
+
+    fn read_field_element(&self) -> E::BaseField;
+
+    fn read_challenge(&mut self) -> Challenge<E>;
+
+    fn send_challenge(&self, challenge: E);
+
+    fn commit_rolling(&mut self) {
+        // do nothing
+    }
```
icemelon, line 94:
```suggestion
    fn commit_rolling(&mut self);
```

transcript/src/lib.rs
```diff
@@ -14,3 +14,96 @@ mod hasher;
...
+        }
+    }
+
+    /// Append the challenge to the transcript.
+    fn append_challenge(&mut self, challenge: Challenge<E>) {
+        self.append_field_element_ext(&challenge.elements)
+    }
+
+    // // Append the message to the transcript.
+    // pub fn append_serializable_element<S: Serialize>(
```
icemelon, line 62: remove this function?

lispc (approved):

matthiasgoergens:
This seems to have broken integration tests.  See https://github.com/scroll-tech/ceno/issues/715 and https://github.com/scroll-tech/ceno/pull/714

icemelon:
Why CI didn't catch the integration test failure?

matthiasgoergens:
> Why CI didn't catch the integration test failure?

Because the CI is running the tests with `--release` which turns off debug assertions.

My PR fixes that and runs the integration tests in the dev profile, but with optimisations turned on.

(However my PR doesn't fix the bug. I don't understand the problem well enough to quickly hammer out a fix for that unfortunately.)

PR #695, opened 2024-12-05
WIP Feat/sharding

lispc:


ceno_zkvm/src/scheme/prover.rs
```diff
@@ -90,33 +92,59 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
         }
         exit_span!(span);
 
+        // TODO: is it better to set different size of different opcode?
+        let shard_size = 1048576;
```
matthiasgoergens, line 96: This might be easier to read in hex.  (Because it's not just a random number.)

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -90,33 +92,59 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
...
+
         // commit to main traces
-        let mut commitments = BTreeMap::new();
-        let mut wits = BTreeMap::new();
+        // TODO: (1) is it ok to store mle? (2) replace tuple with struct?
+        #[allow(clippy::type_complexity)]
+        let mut wits_and_commitments: BTreeMap<
+            String,
```
matthiasgoergens, line 102: What's the meaning of the key here?  A  type synonym might be useful here?

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -90,33 +92,59 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
...
         exit_span!(span);
 
+        // TODO: is it better to set different size of different opcode?
+        let shard_size = 1048576;
+
         // commit to main traces
-        let mut commitments = BTreeMap::new();
-        let mut wits = BTreeMap::new();
+        // TODO: (1) is it ok to store mle? (2) replace tuple with struct?
+        #[allow(clippy::type_complexity)]
```
matthiasgoergens, line 100: Perhaps use some type synonym or so?

ceno_zkvm/src/scheme/verifier.rs
```diff
@@ -61,10 +61,11 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMVerifier<E, PCS>
         does_halt: bool,
     ) -> Result<bool, ZKVMError> {
         // require ecall/halt proof to exist, depending whether we expect a halt.
+        // TODO: make it less adhoc
```
matthiasgoergens, line 64: Especially once we support more ecalls.  Or perhaps we should introduce a specific 'halt-successfully' introduction.

ceno_zkvm/src/witness.rs
```diff
@@ -107,6 +110,42 @@ impl<T: Sized + Sync + Clone + Send + Copy> RowMajorMatrix<T> {
             })
             .collect()
     }
+
+    // TODO: should we consume or clone `self`?
```
matthiasgoergens, line 114: If it's natural to consume `self`, then do so.  The caller can clone.

matthiasgoergens (commented):
A few comments and questions.  I'm still trying to understand what's going on.

hero78119 (commented):
sharding looks pretty straightforward and simple 👍

naure:
This makes sense, nice and simple. Is this PR only for discussion/example or would you like to merge it?

PR #703, opened 2024-12-05
Further break down e2e logic

### LINKED ISSUE #691: Break down e2e logic into smaller components
Currently end-to-end logic is contained in one long function (`run_e2e` in `src/e2e.rs`). We want to break it down into stand-alone components such as, for example, `circuit generation + fixed traces + witness generation + keygen + proof + verify`.

mcalancea:
Fixes #691.

- Refactors e2e into smaller steps. 
- Introduces a `run_partial` method which takes a `prefix` argument, indicating when to stop the run of the pipeline. According to this argument, different parts of the state are yielded back. This part isn't too pretty type-wise, but ultimately I think it's not dangerous and a reasonable compromise for the moment.

Later edit: I've also included a benchmark for witness generation of the Fibonacci program, since it is a consumer of this refactor and helps to exemplify it.


ceno_zkvm/src/e2e.rs
```diff
@@ -219,38 +142,257 @@ pub fn run_e2e_gen_witness<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>
...
+
+    // keygen
+    let pcs_param = PCS::setup(1 << MAX_NUM_VARIABLES).expect("Basefold PCS setup");
+    let (pp, vp) = PCS::trim(pcs_param, 1 << MAX_NUM_VARIABLES).expect("Basefold trim");
+    let pk = system_config
+        .zkvm_cs
+        .clone()
+        .key_gen::<PCS>(pp.clone(), vp.clone(), zkvm_fixed_traces.clone())
+        .expect("keygen failed");
+    let vk = pk.get_vk();
```
hero78119, line 368: Hye, from line 361 to line 368 it's belong to key setup phase, so we probably need to move this part to line 350 right after `let zkvm_fixed_traces = generate_fixed_traces(&system_config, &init_full_mem, &program);` then counting flow til this point as "PreKeySetup"


Witness simulation + assignment should include into proving flow.

ceno_zkvm/src/e2e.rs
```diff
@@ -219,38 +142,257 @@ pub fn run_e2e_gen_witness<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>
...
+}
+
+pub fn run_partial<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>>(
+    program: Program,
+    platform: Platform,
+    stack_size: u32,
+    heap_size: u32,
+    hints: Vec<u32>,
+    max_steps: usize,
+    prefix: PipelinePrefix,
```
hero78119, line 320: It seems that prefix can be Option, and without giving, it's run as a whole.
Probably better naming of function is `run_e2e_with_checkpoints` :) ?

ceno_zkvm/src/e2e.rs
hero78119: Just update comment to avoid confusion.

Btw probably we can have one more character to differentiate `Pre<v>(ious)` or `Pre<f>(ix)` 😄

ceno_zkvm/src/e2e.rs
mcalancea: Agreed with name change. 

It seems more natural to have a `Complete` option for the `Checkpoint` enum, because the function needs to return the `CheckpointOutput` type anyway, and these two enums look nicer with identical variants.

ceno_zkvm/src/e2e.rs
```diff
@@ -6,135 +6,64 @@ use crate::{
...
+) -> EmulationResult {
+    let InitMemState {
+        mem: mem_init,
+        io: io_init,
+        reg: reg_init,
+        priv_io: _,
+    } = init_mem_state;
 
-    let mut vm = VMState::new(platform.clone(), program);
+    let mut vm: VMState = VMState::new(platform.clone(), program.clone());
```
hero78119, line 55: program clone here might be an overhead for e2e. probably
- plan a: change VMState to new with program reference, which need to introduce new life cycle.
- plan b or we can reuse same VMState with a reset() function

Plan a seems bring less change comparing with plan b

ceno_zkvm/src/e2e.rs
```diff
@@ -219,38 +142,257 @@ pub fn run_e2e_gen_witness<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>
...
+    // Keygen
+    let pcs_param = PCS::setup(1 << MAX_NUM_VARIABLES).expect("Basefold PCS setup");
+    let (pp, vp) = PCS::trim(pcs_param, 1 << MAX_NUM_VARIABLES).expect("Basefold trim");
+    let pk = system_config
+        .zkvm_cs
+        .clone()
+        .key_gen::<PCS>(pp.clone(), vp.clone(), zkvm_fixed_traces.clone())
+        .expect("keygen failed");
+    let vk = pk.get_vk();
+
```
hero78119, line 360: checkpoint 
```rust
let prover = ZKVMProver::new(pk);
let verifier = ZKVMVerifier::new(vk);
if let Checkpoint::PreProving = checkpoint {
    return CheckpointOutput::PreProving((prover, verifier, pi, exit_code));
}
```
proper place is here.

Then
emulate_program + generate_witness + run_e2e_proof 
those steps all belong to proving flow

ceno_zkvm/src/e2e.rs
mcalancea: `VMState` just wanted an `Arc` which I propagated to the top level. Clones are cheap now.

PR #706, opened 2024-12-06
Upgrade `pprof`

matthiasgoergens:
See https://github.com/scroll-tech/ceno/security/dependabot/3 for why.


hero78119 (approved):
👍 thanks for address this issue!

PR #707, opened 2024-12-06
Update toolchain

matthiasgoergens:
Just to avoid falling too far behind.  There seem to be no issues with the update.

### DIFF
--- a/ceno_emul/src/addr.rs
+++ b/ceno_emul/src/addr.rs
@@ -206,7 +206,7 @@ impl IterAddresses for Range<Addr> {
     }
 }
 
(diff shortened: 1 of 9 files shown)


ceno_emul/src/addr.rs
```diff
@@ -206,7 +206,7 @@ impl IterAddresses for Range<Addr> {
     }
 }
 
-impl<'a, T: GetAddr> IterAddresses for &'a [T] {
```
matthiasgoergens, line 209: Default clippy rules have updated.

hero78119 (approved):

PR #708, opened 2024-12-06
Upgrade plonky2

matthiasgoergens:
Plonky2 has released their 1.0 at last.


hero78119 (commented):
Notice this got major version upgrade, would suggest to
- share some highlights for upgraded items
- benchmark to tell the impact scope

matthiasgoergens:
@hero78119 Looking at the [changelog](https://github.com/0xPolygonZero/plonky2/blob/main/CHANGELOG.md) it doesn't look like too much has changed.  (I recognise quite a few of my own contributions.)

It looks like this is not so much an incompatible change like semantic versioning would suggest, but just them saying that the project is stable enough now to deserve a big 1 in front of the version and graduate from the 0.

hero78119:
> @hero78119 Looking at the [changelog](https://github.com/0xPolygonZero/plonky2/blob/main/CHANGELOG.md) it doesn't look like too much has changed. (I recognise quite a few of my own contributions.)
> 
> It looks like this is not so much an incompatible change like semantic versioning would suggest, but just them saying that the project is stable enough now to deserve a big 1 in front of the version and graduate from the 0.

Thanks for the sharing and attached link of change log. I think for critical library dependency bum, we can develop an SOP to  conduct a at least e2e benchmark manually to address the impact of before/after (before we have CI to automate it). 
Could you try to make effort on the benchmark of Ceno server?

matthiasgoergens:
> Thanks for the sharing and attached link of change log. I think for critical library dependency bump, we can develop an SOP to conduct a at least e2e benchmark manually to address the impact of before/after (before we have CI to automate it). Could you try to make effort on the benchmark of Ceno server?

I don't have access to the Ceno server. @kunxian-xia could you kick off a benchmark, please?

icemelon:
Why didn't we use plonky3? I think the utils and goldilocks we used from plonky2 are supported by plonky3 as well and plonky3 should be more future-proof.

matthiasgoergens:
I don't know why we didn't originally use plonky3.

I can tell you that I didn't move to plonky3 here: because that's a completely different library, and I just wanted to do a very simple and quick version upgrade.

We can look into using plonky3 in the future, but that would be a more involved effort.

hero78119:
Closed and addressed in #832

PR #711, opened 2024-12-09
Implement the TODOs in the Harvard Architecture PR

matthiasgoergens:
In the [Harvard Architecture PR](https://github.com/scroll-tech/ceno/pull/688) we left a few TODOs.  Here we make good on them.

Alas, unlike [@icemelon's hopes](https://github.com/scroll-tech/ceno/pull/688#issuecomment-2519283914) the change is slightly more involved.

There's a few things happening in this PR:

- We remove the fake instruction `EANY`.  `EANY` was only ever an artifact of Risc0's incredible 'interesting' approach to decoding.  We use the real instructions `ECALL` and `EBREAK` instead.

- Remove `AUIPC` and `LUI`.  Both of them are now implemented as pseudo-instructions, that translate to `ADDI` during decoding.

- Use the same library as SP1 for RiscV decoding, instead of copy-and-pasting-and-editing Risc0's 'interesting' decoder.  That simplifies our code, and comes with a lot more tests than we ever had.  Both because of explicit tests in the library, and because of the usage in SP1 and other projects. This gets rid of much error prone bit manipulating code.

- Use `struct Instruction` throughout the code when handling and testing instructions, instead of `u32`.  That makes specifying tests a lot simpler and more readable.  No more `0b_000000001010_00000_000_00001_0010011, // addi x1, x0, 10` in the code.

- Remove the notion of executable vs non-executable ROM.  This is only necessary for a von-Neumann architecture: everything that's in our instruction-cache is meant to be executable already.  (We can re-implement this restriction later by controlling what is allowed to make it into the instruction cache when we eg decode the ELF.  But it's unnecessary: we already honour the executable flag for memory sections in the ELF.)

Please review [with whitespace changes hidden](https://github.com/scroll-tech/ceno/pull/711/files?diff=unified&w=1) for an easier reading experience.


ceno_emul/src/rv32im.rs
```diff
@@ -182,335 +173,96 @@ pub enum InsnKind {
     SB,
     SH,
     SW,
-    /// ECALL and EBREAK etc.
-    EANY,
```
matthiasgoergens, line 186: `EANY` was never a real instruction.  It was only ever there to work around Risc0's _interesting_ ideas about decoding.

ceno_emul/Cargo.toml
```diff
@@ -15,6 +15,7 @@ elf = "0.7"
 itertools.workspace = true
 num-derive.workspace = true
 num-traits.workspace = true
+rrs_lib = { package = "rrs-succinct", version = "0.1.0" }
```
icemelon, line 18: why not use the standard `rrs_lib`?

matthiasgoergens: That's what I tried at first, but then I noticed that Succinct had added support for `ecall` and a few other niceties.

ceno_emul/src/rv32im_encode.rs
```diff
@@ -1,116 +1,29 @@
-use crate::{InsnKind, rv32im::InsnFormat};
+use crate::{InsnKind, Instruction};
```
icemelon, line 1: maybe just merge this file with `rv32im.rs`?

ceno_emul/src/rv32im.rs
```diff
@@ -520,25 +264,22 @@ impl Emulator {
         &self,
         ctx: &mut M,
         kind: InsnKind,
-        decoded: &DecodedInstruction,
+        decoded: &Instruction,
```
icemelon, line 267: suggest to rename var to `insn`

ceno_emul/src/rv32im_encode.rs
matthiasgoergens: Yes, good idea.  I left it in the original position to make the PR simpler and smaller.  But we can just move this now.

icemelon (approved):
LGTM

naure:
> Remove AUIPC and LUI.

The LUI circuit is more efficient than ADDI. No register read, fewer range checks, no addition... I recommend putting it back.

You can implement AUIPC with the LUI circuit though (absolute addressing).

matthiasgoergens:
If you think it's worth the hassle, we can restore the LUI circuit and also compile ADDI with rs1=0 to LUI.

naure:
Indeed.

PR #712, opened 2024-12-09
[Feat] Calculate proof size and number of hashes

noel2004:
This PR is base on #692 and provided required features to calculate the size of proofs and number of hashes

Now it has been implied in the e2e test `test_single_add_instance_e2e` and `fibonacci` bench, obtained following output:

> encoded zkvm proof size: 2756371, hash_num: 4411

The encoding of proof is done by `bincode` to kept the serialization as compact as possible


ceno_zkvm/Cargo.toml
```diff
@@ -44,6 +44,7 @@ tempfile = "3.14"
 thread_local = "1.1"
 
 [dev-dependencies]
+bincode = "1"
```
matthiasgoergens, line 47: We already use serde and rkyv, I don't think we need to introduce yet another serialisation framework?  (Perhaps just use serde?)

ceno_zkvm/benches/fibonacci.rs
```diff
@@ -54,6 +85,10 @@ fn fibonacci_prove(c: &mut Criterion) {
                 format!("fibonacci_max_steps_{}", max_steps),
             ),
             |b| {
+                // use itertools::Itertools;
```
matthiasgoergens, line 88: No need to introduce commented out code, please remove.

ceno_zkvm/src/e2e.rs
```diff
@@ -531,6 +531,7 @@ pub fn run_e2e_verify<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>>(
             .verify_proof_halt(zkvm_proof, transcript, exit_code.is_some())
             .expect("verify proof return with error"),
     );
+
```
matthiasgoergens, line 534: Unrelated change, remove.

ceno_zkvm/src/scheme/tests.rs
```diff
@@ -178,6 +179,12 @@ fn test_rw_lk_expression_combination() {
                 &verifier_challenges,
             )
             .expect("verifier failed");
+
+        drop(v_transcript);
```
matthiasgoergens, line 183: Interesting!  Why do we need the manual drop here?

ceno_zkvm/benches/fibonacci.rs
```diff
@@ -43,6 +44,36 @@ fn setup() -> (Program, Platform) {
 fn fibonacci_prove(c: &mut Criterion) {
     let (program, platform) = setup();
     for max_steps in [1usize << 20, 1usize << 21, 1usize << 22] {
+        // estimate proof size data first
```
matthiasgoergens, line 47: Sorry, I don't understand.  Why do we need this estimate first?

ceno_zkvm/benches/fibonacci.rs
```diff
@@ -43,6 +44,36 @@ fn setup() -> (Program, Platform) {
...
+        let (sanity_check_state, _) = run_e2e_with_checkpoint::<E, Pcs>(
+            program.clone(),
+            platform.clone(),
+            vec![],
+            max_steps,
+            Checkpoint::PrepSanityCheck,
+        );
+        assert!(
```
matthiasgoergens, line 55: You can combine your `assert` and your `if let`, since they are checking for the same condition.

Something like `let (proof, verifier)) = sanity_check_state.expect("PrepSanityCheck does not provide proof and verifier");` should do the trick.  (But I don't care about the specifics.  It's just weird to have the same check twice.)

transcript/src/statistics.rs
matthiasgoergens, line 1: Could you add some comments explaining what we are doing here and why, please?

matthiasgoergens (commented):
See https://github.com/scroll-tech/ceno/pull/770 for some groundwork.

ceno_zkvm/src/scheme/tests.rs
```diff
@@ -155,8 +155,9 @@ fn test_rw_lk_expression_combination() {
             .expect("create_proof failed");
 
         // verify proof
+        let static_recorder = StatisticRecorder::new();
```
matthiasgoergens, line 158: What is 'static' about this recorder?

matthiasgoergens (commented):
Please see https://github.com/scroll-tech/ceno/pull/771

matthiasgoergens:
I merged in latest master (and resolved the conflict).

ceno_zkvm/Cargo.toml
matthiasgoergens: Having said all that, if you want to keep bincode, you can keep it.  The only real downside I see is a slightly longer build time, but it's probably not that bad.

matthiasgoergens (approved):
Please remove the commented out code etc (see my comments), then we can merge.  Thanks!

ceno_zkvm/Cargo.toml
noel2004: If we wish to compare the size of our proof with sp1 directly, bincode would provide a bit more confident to the comparision since sp1 use bincode to serialize their proof.

And yes, we can only apply this crate as a dev-dependency

ceno_zkvm/benches/fibonacci.rs
kunxian-xia: Yeah, I think we should move this to `e2e.rs` after we generated the proof. The benchmark we have below is only used to see if there is progress or regression. The proof size should be identical in multiple proving instances.

transcript/src/statistics.rs
```diff
@@ -0,0 +1,90 @@
...
+type SharedStatisticRecorder = RefCell<StatisticRecorder>;
+
+impl StatisticRecorder {
+    pub fn new() -> Rc<SharedStatisticRecorder> {
+        Rc::new(RefCell::new(Default::default()))
+    }
+}
+
+#[derive(Clone)]
+pub struct BasicTranscriptWitStat<E: ExtensionField> {
```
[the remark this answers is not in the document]
matthiasgoergens: Most likely!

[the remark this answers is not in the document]
noel2004: Have been fixed in #777

Issue #715, opened 2024-12-09
Integration tests are broken on master


matthiasgoergens:
238516f652467ebb2738f7cb02052723d5d8a179 / https://github.com/scroll-tech/ceno/pull/692 seems to be the first bad commit that fails with:

```terminal
$ RUST_BACKTRACE=1 cargo run --package ceno_zkvm --example riscv_opcodes -- --start 10 --end 11
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.05s
     Running `target/debug/examples/riscv_opcodes --start 10 --end 11`
thread 'main' panicked at ff_ext/src/lib.rs:53:13:
assertion `left == right` failed
  left: 8
 right: 2
stack backtrace:
   0: rust_begin_unwind
             at /rustc/c94848c046d29f9a80c09aae758e27e418a289f2/library/std/src/panicking.rs:681:5
   1: core::panicking::panic_fmt
             at /rustc/c94848c046d29f9a80c09aae758e27e418a289f2/library/core/src/panicking.rs:75:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/c94848c046d29f9a80c09aae758e27e418a289f2/library/core/src/panicking.rs:364:5
   4: ff_ext::impl_goldilocks::<impl ff_ext::ExtensionField for goldilocks::structs::GoldilocksExt2>::from_bases
             at ./ff_ext/src/lib.rs:53:13
   5: <transcript::basic::BasicTranscript<E> as transcript::Transcript<E>>::read_challenge
             at ./transcript/src/basic.rs:35:17
   6: ceno_zkvm::scheme::prover::ZKVMProver<E,PCS>::create_proof
             at ./ceno_zkvm/src/scheme/prover.rs:125:13
   7: riscv_opcodes::main
             at ./ceno_zkvm/examples/riscv_opcodes.rs:304:30
   8: core::ops::function::FnOnce::call_once
             at /home/matthias/.rustup/toolchains/nightly-2024-12-06-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

_Originally posted by @matthiasgoergens in https://github.com/scroll-tech/ceno/issues/714#issuecomment-2527571873_

matthiasgoergens:
[...]

PR #714, opened 2024-12-09
Enable debug assertions in integration tests

matthiasgoergens:
To catch some bugs, like the recent breakage.

PR #719, opened 2024-12-10
Make `cargo build` work

matthiasgoergens:
Right now a simple `cargo build` fails.  So let's fix that.

This PR also disentangles `ceno_rt`, so that we can work on building guest programs much easier.

Update: I extracted the remove of the target matrix into https://github.com/scroll-tech/ceno/pull/725 for ease of review.

This PR also enables debug assertions for our tests.


matthiasgoergens:
Tests are failing, because they are broken on master.

.github/workflows/integration.yml
```diff
@@ -32,13 +32,6 @@ jobs:
     timeout-minutes: 30
     runs-on: [self-hosted, Linux, X64]
 
-    strategy:
```
matthiasgoergens, line 35: Our strategy matrix only has one entry.  So we might as well not bother.

.github/workflows/integration.yml
```diff
@@ -50,16 +43,17 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
+            guest/target/
           key: integration-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: dtolnay/rust-toolchain@nightly
 
       - name: Run example
         env:
-          RAYON_NUM_THREADS: 2
```
matthiasgoergens, line 58: This isn't read anywhere.

.github/workflows/integration.yml
matthiasgoergens: Extracted into https://github.com/scroll-tech/ceno/pull/725

guest/examples/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+categories.workspace = true
+description = "Ceno runtime library"
```
lispc, line 3: Ceno runtime library => Ceno examples

Makefile.toml
```diff
@@ -11,10 +9,7 @@ args = [
   "--bins",
   "--tests",
   "--examples",
-  "--release",
```
lispc, line 14: not sure whether it is good to remove the `--release`.   Feel free to close this if that is ok.

lispc (approved):
only some small comments

Makefile.toml
matthiasgoergens: Good spotting.  In this case, it's what we want to do:

`--release` disables debug assertions.  Recently we had a problem with [debug assertions being broken](https://github.com/scroll-tech/ceno/issues/715).  And we already [enabled them for integration tests](https://github.com/scroll-tech/ceno/pull/714/files).

guest/examples/Cargo.toml
matthiasgoergens: Good catch!

Makefile.toml
matthiasgoergens: In compensation, we upped the default optimisation level for the dev-build to 2.  As a compromise between speed of building vs speed of execution.

examples/.cargo/config.toml
matthiasgoergens, line 1: This file moved, it wasn't deleted.  Command line git shows that, but not github.  Alas.

PR #721, opened 2024-12-10
[FIX] Broken integration test caused by transcript (#715)

noel2004:
This PR has fixed issue #715 by respect the assertion in `from_base` method of the Godilock extension field

Notice there is two method `from_base` and `form_limbs` in the `ExtensionField` trait which is not well illustrated and maybe we should update them later.

The two running mentioned in #715 have been success under this PR

`cargo run --package ceno_zkvm --bin e2e -- --platform=sp1 ceno_zkvm/examples/fibonacci.elf`

`RUST_BACKTRACE=1 cargo run --package ceno_zkvm --example riscv_opcodes -- --start 10 --end 11`

### DIFF
--- a/transcript/src/basic.rs
+++ b/transcript/src/basic.rs
@@ -32,7 +32,7 @@ impl<E: ExtensionField> Transcript<E> for BasicTranscript<E> {
     }
 
     fn read_challenge(&mut self) -> Challenge<E> {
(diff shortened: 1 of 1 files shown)


matthiasgoergens (approved):
Thanks for the quick fix.

PR #722, opened 2024-12-10
Remove broken Make target `clippy-all`

matthiasgoergens:
We have mutually incompatible features, so running anything with `--all-features` is bound to fail.


yczhangsjtu (approved):

yczhangsjtu:
If this check is failing, why are the checks still passing for existing PRs?

matthiasgoergens:
> If this check is failing, why are the checks still passing for existing PRs?

We aren't using this check. We use a related, but different one.

PR #723, opened 2024-12-10
Remove broken features `non_pow2_rayon_thread` and `riv64`

matthiasgoergens:
Features `non_pow2_rayon_thread` and `riv64` are not tested: we can't even compile with them.

Because `riv64` is gone, we can make `riv32` invisible: it's effects are now just the default.


roylou (approved):
approve assuming the 2 features are not used currently.

matthiasgoergens:
> approve assuming the 2 features are not used currently.

Well, they can’t have been used, because if you enable any of them, you get compiler errors. :)

PR #726, opened 2024-12-10
Security-harden the guest's memory allocator

matthiasgoergens:
Here we harden the security of the guest's memory allocator: we no longer rely on manual initialisation code for its proper operation.

Because we remove code, it also simplifies the allocator.  Hopefully making it easier to understand.

Extracted from https://github.com/scroll-tech/ceno/pull/631


ceno_rt/src/allocator.rs
```diff
@@ -4,26 +4,34 @@
 use core::alloc::{GlobalAlloc, Layout};
 
 struct SimpleAllocator {
-    next_alloc: usize,
+    next_alloc: *mut u8,
```
matthiasgoergens, line 7: This type is more truthful, and simplifies our code.

ceno_rt/src/allocator.rs
```diff
@@ -4,26 +4,34 @@
...
         let align = layout.align();
         // `Layout` contract forbids making a `Layout` with align=0, or align not power of 2.
-        // So we can safely use subtraction and a mask to ensure alignment without worrying about UB.
-        let offset = heap_pos & (align - 1);
-        if offset != 0 {
-            heap_pos = heap_pos.strict_add(align.strict_sub(offset));
-        }
+        core::hint::assert_unchecked(align.is_power_of_two());
+        core::hint::assert_unchecked(align != 0);
+        heap_pos = heap_pos.add(heap_pos.align_offset(align));
```
matthiasgoergens, line 29: We use the proper Rust functions for working with pointers now.  That makes both the type checker happier (when working with pointers instead of `usize`), and hopefully also clues in the optimiser a bit more.

But the latter would just be a welcome side effect, if it does happen.

ceno_rt/src/allocator.rs
```diff
@@ -36,14 +44,3 @@ unsafe impl GlobalAlloc for SimpleAllocator {
...
 }
-
-#[global_allocator]
-// We initialize `next_alloc` to 0xFFFF_FFFF to indicate that the heap has not been initialized.
-// The value is chosen to make any premature allocation fail.
-static mut HEAP: SimpleAllocator = SimpleAllocator {
-    next_alloc: 0xFFFF_FFFF,
-};
-
-pub unsafe fn init_heap() {
```
matthiasgoergens, line 47: We no longer need init heap.

ceno_rt/src/allocator.rs
```diff
@@ -14,16 +14,14 @@ unsafe impl GlobalAlloc for SimpleAllocator {
...
-        let ptr = heap_pos as *mut u8;
-        // Panic on overflow.  We don't want to wrap around, and overwrite stack etc.
+        core::hint::assert_unchecked(align.is_power_of_two());
+        core::hint::assert_unchecked(align != 0);
+        heap_pos = heap_pos.add(heap_pos.align_offset(align));
+
+        let ptr = heap_pos;
+        // We don't want to wrap around, and overwrite stack etc.
         // (We could also return a null pointer, but only malicious programs would ever hit this.)
-        heap_pos = heap_pos.strict_add(layout.size());
```
matthiasgoergens, line 26: Alas, `ptr` type doesn't have `strict_add`, but `add` does nearly the same.

lispc (approved):
my understanding of this PR:

previously, we also need to set the `_sheap` var somewhere(also linking time?), then `allocator::init_heap()` will use that value to overwrite the default 0xffffffff. 

This PR seems a pure code simplification.  Do it make anything really better or really worse?   (if we are not talking about "more robust so in future there may be less bug in code modification", but just talk current code behavior)

matthiasgoergens:
> my understanding of this PR:
> 
> previously, we also need to set the `_sheap` var somewhere(also linking time?), then `allocator::init_heap()` will use that value to overwrite the default 0xffffffff.

To be really precise: we only ever care about the address of `_sheap`, not really the value.  And, yes, both before and after this PR, that address gets set at linking time.

In this PR, I figured out that we can use Rust's static initialisation to set `next_alloc` to the address of `_sheap`.  Previously we set `next_alloc` to a dummy value, and then used normal code to initialise `next_alloc` to a sensible value (the address of `_sheap`) at the start of normal runtime.

Then link time optimisation can kick in, and constant propagation can make `next_alloc` be initialised via program data in the ELF instead of via code.

> This PR seems a good pure code simplification change. Does it make anything really better or really worse? (if we are not talking about "more robust so in future there may be less bug in code modification", but just talk current code behavior)

Current behaviour should be the same.  The original motivation for this change was to make the standard library support simpler to implement: this way we have to worry less about what happens code wise before we run `main`.

So to be honest, the motivation given in the PR description is a bit of a retro-fit.  It's not wrong, but it's more of a side-benefit of a change we want to make for ease of standard library support.

PR #732, opened 2024-12-11
Introduce `Value::as_i32`

matthiasgoergens:
To help make https://github.com/scroll-tech/ceno/pull/596 easier to read and reason about.

Also introduce a few more conversion helpers.

### DIFF
--- a/ceno_zkvm/src/uint.rs
+++ b/ceno_zkvm/src/uint.rs
@@ -606,6 +606,24 @@ pub struct Value<'a, T: Into<u64> + From<u32> + Copy + Default> {
     pub limbs: Cow<'a, [u16]>,
 }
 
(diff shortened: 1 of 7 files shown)


ceno_zkvm/src/uint.rs
```diff
@@ -684,6 +684,11 @@ impl<'a, T: Into<u64> + From<u32> + Copy + Default> Value<'a, T> {
         self.as_u64() as u32
     }
 
+    /// Convert the limbs to a u32 value
```
bgillesp, line 687: Copy-paste error: should be `Convert the limbs to an i32 value`

bgillesp (approved):
Seems like a reasonable utility function to add -- see just the one copy error in the doc comment for the new `UInt` function.

ceno_zkvm/src/uint.rs
matthiasgoergens: Good catch!

Issue #215, opened 2024-09-12
Decide and implement details of platform, memory layout, and I/O


naure:
The platform is the specification of how programs are loaded, what are the regions of memory, how I/O works, and more details.

The current draft is essentially this [platform.rs](https://github.com/scroll-tech/ceno/blob/master/ceno_emul/src/platform.rs), this [linker script](https://github.com/scroll-tech/ceno/blob/master/ceno_rt/ceno_link.x), and a [demo of MMIO](https://github.com/scroll-tech/ceno/blob/master/ceno_rt/src/io.rs).

matthiasgoergens:
@naure What do you mean by 'virtual memory'?  I know the term in the context of real physical computers, but I don't know what it means in our context.

matthiasgoergens:
> * Support panic as a valid outcome, and include the exit code as part of public I/O (see [Feat: Ecall/Halt #258](https://github.com/scroll-tech/ceno/pull/258)).

I'd rather not.  It's extra complexity that we don't need.

naure:
I don't think this issue is closed; or it should have a follow up issue. We still need to apply the choices and in particular update the Platform type (replace rom/ram fields and such).

matthiasgoergens:
> I don't think this issue is closed; or it should have a follow up issue. We still need to apply the choices and in particular update the Platform type (replace rom/ram fields and such).

I updated the title to say 'Decide and implement [...]' to make this issue its own follow-up and re-opened it.

naure:
Implementation: https://github.com/scroll-tech/ceno/pull/735

PR #735, opened 2024-12-12
sync-platform: Configure the platform / emulator to match circuits

naure:
Over time we moved from a first version (https://github.com/scroll-tech/ceno/issues/215) which looked end-to-end like this:

```
Static memory layout 
            `-->  Program compilation
            `-->  Trace generation
            `-->  Static circuits
```

to this

```
Existing third-party programs
            `-->  Auto-detected memory layout
                        `-->  Trace generation
                        `-->  Configurable circuits
```

This worked by letting the trace generation be more powerful and more permissive than the circuits (larger memory ranges, etc).

This PR configures the trace generation to tightly match the circuit behaviors as of today.


matthiasgoergens:
Could you please expand the description with a short note on what aspects you are syncing up, please?

ceno_emul/src/platform.rs
```diff
@@ -9,33 +9,42 @@ use crate::addr::{Addr, RegIdx};
 /// - codes of environment calls.
 #[derive(Clone, Debug)]
 pub struct Platform {
-    pub rom: Range<Addr>,
-    pub ram: Range<Addr>,
+    pub prog_code: Range<Addr>,
```
matthiasgoergens, line 12: We don't need `prog_code` in the platform.  There's no restrictions we need to enforce, apart from what's in the ELF, and we enforce them during decoding.

That's all thanks to the recent move to a Harvard Architecture.

ceno_emul/src/platform.rs
```diff
@@ -9,33 +9,42 @@ use crate::addr::{Addr, RegIdx};
...
+    pub prog_data: Option<HashSet<Addr>>,
+    pub stack: Range<Addr>,
```
matthiasgoergens, line 14: I like the move to using a Range here!

ceno_emul/src/platform.rs
```diff
@@ -9,33 +9,42 @@ use crate::addr::{Addr, RegIdx};
 /// - codes of environment calls.
 #[derive(Clone, Debug)]
 pub struct Platform {
-    pub rom: Range<Addr>,
-    pub ram: Range<Addr>,
+    pub prog_code: Range<Addr>,
+    pub prog_data: Option<HashSet<Addr>>,
```
matthiasgoergens, line 13: Consider changing to `BTreeSet` which has `pub const fn new() -> BTreeSet<T>` to make `const` objects.

ceno_emul/src/platform.rs
```diff
@@ -9,33 +9,42 @@ use crate::addr::{Addr, RegIdx};
...
-    pub stack_top: Addr,
     /// If true, ecall instructions are no-op instead of trap. Testing only.
     pub unsafe_ecall_nop: bool,
 }
 
 pub const CENO_PLATFORM: Platform = Platform {
-    rom: 0x2000_0000..0x3000_0000,
-    ram: 0x8000_0000..0xFFFF_0000,
+    prog_code: 0x2000_0000..0x3000_0000,
+    prog_data: None, // This is an `Option` to allow `const` here.
```
matthiasgoergens, line 24: Move the comment to the type definition for Platform, please.  (Or change to `BTreeSet` and avoid the problem.)

ceno_emul/src/platform.rs
```diff
@@ -9,33 +9,42 @@ use crate::addr::{Addr, RegIdx};
...
-    stack_top: 0xC0000000,
     unsafe_ecall_nop: false,
 };
 
 impl Platform {
     // Virtual memory layout.
 
-    pub fn is_rom(&self, addr: Addr) -> bool {
-        self.rom.contains(&addr)
+    pub fn is_prog_code(&self, addr: Addr) -> bool {
```
matthiasgoergens, line 35: We don't need this at all in the platform.  See the comment about Harvard architecture and decoding.

ceno_emul/src/platform.rs
```diff
@@ -60,21 +69,21 @@ impl Platform {
...
     // Permissions.
 
     pub fn can_read(&self, addr: Addr) -> bool {
-        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr) || self.is_hints(addr)
+        self.can_write(addr)
     }
 
     pub fn can_write(&self, addr: Addr) -> bool {
-        self.is_ram(addr)
+        self.is_ram(addr) || self.is_pub_io(addr) || self.is_hints(addr)
```
matthiasgoergens, line 82: `hints` (and pub io) are supposed be read-only.

ceno_zkvm/src/e2e.rs
```diff
@@ -159,33 +160,79 @@ fn emulate_program(
...
+            stack: 0x0020_0400..0x0020_0400,
+            unsafe_ecall_nop: true,
+            ..CENO_PLATFORM
+        },
+    };
+
+    let prog_data = program.image.keys().copied().collect::<HashSet<Addr>>();
+    let stack = preset.stack.end - stack_size..preset.stack.end;
+    let heap = {
+        // Detect heap as starting after program data.
```
matthiasgoergens, line 189: Btw, for ceno-compiled code ELFs, we have the `_sheap` and `_stack_start` symbols, which we can use read where heap and stack actually are.

We should also introduce symbols for the end of these, like we do for hints in https://github.com/scroll-tech/ceno/pull/631/files#diff-4994466418e9592df592e3378c56430a43d10451155f5f415da95ebe1a8699a7

So no need to come up with questionable heuristics for these values here.

ceno_zkvm/src/e2e.rs
```diff
@@ -159,33 +160,79 @@ fn emulate_program(
...
+        let heap_start = program.image.keys().max().unwrap() + WORD_SIZE as u32;
+        let heap = heap_start..heap_start + heap_size;
+        // Pad the total size to the next power of two.
```
matthiasgoergens, line 192: Please compare the discussion around https://github.com/scroll-tech/ceno/pull/716

Let's pick one policy for whether we should round up or down or assert, and decide one ideally one place where we do that.

matthiasgoergens (commented):
Overall a nice consolidation and simplification!

Despite the title of the PR, I didn't see all that much new synchronisation between the platform and emulator here?  Perhaps I'm missing something.  Could you please elaborate in the PR description what new aspects we are matching that we didn't match before?  Thanks!

(I do mostly like the changes here in any case, even if I don't quite understand the title.)

matthiasgoergens:
Please mention that we are doing this partially for SP1 compatibility in the PR description.

ceno_emul/src/platform.rs
matthiasgoergens: Hmm, all the comments here for this file seem to have moved down by one line, compared to where they should be.

ceno_emul/src/platform.rs
```diff
@@ -60,21 +69,21 @@ impl Platform {
...
     pub const fn pc_base(&self) -> Addr {
-        self.rom.start
+        self.prog_code.start
     }
 
     // Permissions.
 
     pub fn can_read(&self, addr: Addr) -> bool {
-        self.is_rom(addr) || self.is_ram(addr) || self.is_pub_io(addr) || self.is_hints(addr)
+        self.can_write(addr)
```
matthiasgoergens, line 78: This completely precludes read-only memory?

ceno_emul/src/platform.rs
```diff
@@ -60,21 +69,21 @@ impl Platform {
...
     }
 
     pub fn can_write(&self, addr: Addr) -> bool {
-        self.is_ram(addr)
+        self.is_ram(addr) || self.is_pub_io(addr) || self.is_hints(addr)
     }
 
     pub fn can_execute(&self, addr: Addr) -> bool {
```
matthiasgoergens, line 85: We don't need this anymore.

matthiasgoergens (changes requested):
Please see comments.

ceno_emul/src/platform.rs
naure: Reverted.

ceno_emul/src/platform.rs
naure: Done.

ceno_emul/src/platform.rs
naure: They are not read-only so this matches correctly.

ceno_emul/src/platform.rs
naure: Yes, there are no read-only memory anywhere.

ceno_emul/src/platform.rs
naure: If you mean to merge `can_execute` into `fetch`, it should be removed in the PR that does that.

ceno_zkvm/src/e2e.rs
naure: The "questionable heuristic" is the normal behavior of the linker.

I did consider switching to a symbol-based solution potentially that but that’s not the scope here.

lispc (approved):

ceno_emul/src/platform.rs
matthiasgoergens: Done.

ceno_emul/src/platform.rs
matthiasgoergens: OK.  I created an issue at https://github.com/scroll-tech/ceno/issues/751

Btw, we do have read only sections in the ELF already, and we should honour them.

matthiasgoergens (changes requested):
Sorry, I prematurely removed my hold.  Please just use a `BTreeSet` or at least a `///` doc comment.

ceno_zkvm/src/e2e.rs
matthiasgoergens: How do you know it's not related, if you don't understand it?

ceno_emul/src/platform.rs
matthiasgoergens: Interesting.  Thanks!

Then a `HashSet` is as good as the `BTreeSet`, but please use a doc comment for the explanation, so it shows up in our tooling.  Thanks!

ceno_emul/src/platform.rs
```diff
@@ -10,20 +10,23 @@ use crate::addr::{Addr, RegIdx};
 #[derive(Clone, Debug)]
 pub struct Platform {
     pub rom: Range<Addr>,
-    pub ram: Range<Addr>,
+    // This is an `Option` to allow `const` here.
```
matthiasgoergens, line 13:
```suggestion
    /// This is an `Option` to allow `pub const CENO_PLATFORM: Platform`
```

matthiasgoergens:
I enabled auto-merge again.  (It got disabled, when I mucked around with my review.)

matthiasgoergens:
@naure Sorry, the renaming of ROM to program code (or similar) is fine.  And I recognise that the changes you made to the `can_write` function is also fine in the sense that they make the emulator reflect what the circuits currently enforce.  (A promised in the PR description.)

My request to respect read-only memory is more an aspiration for the future development.

matthiasgoergens (approved):

PR #736, opened 2024-12-12
Use `ToExpr` trait instead of ad-hoc `expr` functions

matthiasgoergens:


ceno_zkvm/src/expression.rs
```diff
@@ -767,48 +776,55 @@ impl WitIn {
...
 }
 
 impl<E: ExtensionField> ToExpr<E> for Instance {
     type Output = Expression<E>;
-    fn expr(&self) -> Expression<E> {
-        Expression::Instance(*self)
-    }
-}
-
-impl<F: SmallField, E: ExtensionField<BaseField = F>> ToExpr<E> for F {
```
matthiasgoergens, line 808: This implementation of the trait is a bit problematic, because it clashes with making implementations for eg `u64`.

ceno_zkvm/src/expression.rs
```diff
@@ -823,45 +839,41 @@ macro_rules! impl_from_via_ToExpr {
...
+                    Expression::Constant(F::from(self as u64))
                 }
             }
         )*
     };
 }
-impl_from_unsigned!(u8, u16, u32, u64, usize, RAMType, InsnKind);
+impl_ToExpr_unsigned!(u8, u16, u32, u64, usize, RAMType, InsnKind);
 
-// Implement From trait for u128 separately since it requires explicit reduction
```
matthiasgoergens, line 843: We never actually used this implementation anywhere.

PR #738, opened 2024-12-12
Use built-in `sum` instead of writing our own ad-hoc version

matthiasgoergens:
Extracted from https://github.com/scroll-tech/ceno/pull/736 for ease of review.


mpcs/src/sum_check/classic/coeff.rs
```diff
@@ -49,9 +49,7 @@ impl<E: ExtensionField> ClassicSumCheckRoundMessage<E> for Coefficients<E> {
     }
 
     fn sum(&self) -> E {
-        self[1..]
-            .iter()
-            .fold(self[0].double(), |acc, coeff| acc + coeff)
+        self[0] + self[..].iter().sum::<E>()
```
matthiasgoergens, line 52: Doubling the first element is a bit weird.  But the tests seem to expect it.  What's the reason for it?

chaosma (approved):

PR #740, opened 2024-12-12
Feat/structural witin add

10to4:
Work for #654

PR #741, opened 2024-12-12
Clean up Circuit Builder

matthiasgoergens:
@10to4 Your https://github.com/scroll-tech/ceno/pull/740 reminded me of some old improvements I had lying around.

Especially the use of the standard library's `join`, instead of implementing the functionality manually.


ceno_zkvm/src/circuit_builder.rs
```diff
@@ -41,24 +41,10 @@ impl NameSpace {
...
-            if needs_separation {
-                name += "/";
-            }
-
-            name += ns;
-            needs_separation = true;
-        }
-
-        name
+        chain!(self.get_namespaces(), once(&this)).join("/")
```
matthiasgoergens, line 47: This is mostly just realising that `join` exists and does exactly what we implemented above manually.

yczhangsjtu (approved):

Issue #642, opened 2024-11-27
WHIR wrapper


icemelon:
The goal of this task is to make a wrapper of https://github.com/scroll-tech/whir to be compatible with the traits that we have for [mpcs](https://github.com/scroll-tech/ceno/blob/master/mpcs/src/lib.rs#L110) in ceno right now.

kunxian-xia:
@yczhangsjtu @chaosma Could you start to work on this?

lispc:
Sub Tasks

[] ff <-> ark-ff 
[] ..
[] ..
[] testing with ?? exmaple, and report performance gain. (proof size, verification time..)

current plan:
Dec 23: likely testing done, in review phase.

kunxian-xia:
@yczhangsjtu please update the links to pull requests.

hero78119:
Done via #742

PR #742, opened 2024-12-12
WHIR wrapper

yczhangsjtu:
Add the implementation of WHIR from  #642


mpcs/src/whir/field_wrapper.rs
```diff
@@ -0,0 +1,600 @@
+use ark_ff::{AdditiveGroup, Field, LegendreSymbol};
```
hero78119, line 1: Introduce ark_ff field + wrapper would make slightly large technical burden.

In these case, I would suggest we go from another way to make https://github.com/scroll-tech/whir to use plonky3 field library, so it would be easiler align everything in the future in "production ready" manner

lispc:
oh the merge conflicts seems much less than i thought..

hero78119 (approved):
LGTM and good to go in this stage! in particular we haven't switch default PCS to WHIR and remain in basefold so should be no impact.

There are many chores during reviews, to improving it we need to do big refactor to migrate WHIR. Per discussed offline, its good to go with "build pass", and let do all the improvement incrementally in later PR.

PR #745, opened 2024-12-13
MPCS: Rename CommitmentWithData to CommitmentWithWitness & remove unnecessary trait requirement.

yczhangsjtu:
- The name "CommitmentWithData" is vague, renamed to `CommitmentWithWitness` to be clearer
- It doesn't actually need to be serialized anywhere in the protocol, and never need to create a default instance, remove the unnecessary trait bounds (which doesn't cause any compilation error at all), and to avoid implementing these trait unnecessarily in WHIR wrapper


matthiasgoergens (approved):
Looks good.

matthiasgoergens:
Thanks for providing a detailed motivation in the PR description.

hero78119 (approved):
👍

PR #750, opened 2024-12-13
Fix InsnKind in SLT

### LINKED ISSUE #746: Bug: fail to use InsnKind::SLTU in the lookup to program table for SLTU instruction
When I tried to run the `sproll-evm` test using latest master branch, it fails with lots of lookup errors like this

```
ERROR ceno_zkvm::scheme::mock_prover: riscv_opcode/SLTU/fetch: value [Goldilocks(2ab85c), Goldilocks(9), Goldilocks(5), Goldilocks(c), Goldilocks(10), Goldilocks(0)] mismatch lk_multiplicity: real e6 > remaining 0 in Instruction table
```

You can reproduce this bug by the following cmd:
```
MOCK_PROVING=true RAYON_NUM_THREADS=32 RUST_LOG=info cargo run --package ceno_zkvm --bin e2e --release -- --platform=sp1 --hints ./poseidon_hint.bin riscv32im-succinct-zkvm-elf
```

zemse:
Closes #746

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/slt.rs
+++ b/ceno_zkvm/src/instructions/riscv/slt.rs
@@ -83,7 +83,7 @@ impl<E: ExtensionField, I: RIVInstruction> Instruction<E> for SetLessThanInstruc
 
         let r_insn = RInstructionConfig::<E>::construct_circuit(
             cb,
-            InsnKind::SLT,
+            I::INST_KIND,
             rs1_read.register_expr(),
             rs2_read.register_expr(),
             rd_written.register_expr(),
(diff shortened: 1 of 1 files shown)


matthiasgoergens (approved):
Could you please add a test that exercise this functionality?

That is a test that would fail on current master, but succeeds here?

zemse:
There was a test but it had the same problem on line 162

ceno_zkvm/src/instructions/riscv/slt.rs
```diff
@@ -159,7 +159,7 @@ mod test {
             .unwrap()
             .unwrap();
 
-        let insn_code = encode_rv32(InsnKind::SLT, 2, 3, 4, 0);
```
zemse, line 162: It had the same problem in the test, it is updated now.

kunxian-xia (approved):
Thanks for your fix.

PR #757, opened 2024-12-14
refactor: Remove `Option` wrapper around `prog_data`

matthiasgoergens:
We found a neat way to avoid the extra `Option` wrapper around `prog_data`.

We only ever needed it, because of some technicalities with the Rust type system.


ceno_emul/src/platform.rs
```diff
@@ -10,8 +10,7 @@ use crate::addr::{Addr, RegIdx};
 #[derive(Clone, Debug)]
 pub struct Platform {
     pub rom: Range<Addr>,
-    // This is an `Option` to allow `const` here.
-    pub prog_data: Option<HashSet<Addr>>,
+    pub prog_data: BTreeSet<Addr>,
```
matthiasgoergens, line 13: `HashSet` doesn't have `const` in `pub fn new() -> HashSet<T, RandomState>`.

But `BTreeSet` has `pub const fn new() -> BTreeSet<T>`.

ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -38,7 +38,10 @@ use strum::IntoEnumIterator;
 
 const MAX_CONSTRAINT_DEGREE: usize = 2;
 const MOCK_PROGRAM_SIZE: usize = 32;
-pub const MOCK_PC_START: ByteAddr = ByteAddr(CENO_PLATFORM.pc_base());
+pub static MOCK_PC_START: ByteAddr = ByteAddr({
+    static CENO_PLATFORM: Platform = ceno_emul::CENO_PLATFORM;
```
matthiasgoergens, line 42: Because of the way `const` and `static` interact, we are statically leaking a single copy of `CENO_PLATFORM` here.  But that's only a few bytes, and the compiler optimises it away on anything higher than `-O0`.

ceno_emul/src/platform.rs
```diff
@@ -10,8 +10,7 @@ use crate::addr::{Addr, RegIdx};
 #[derive(Clone, Debug)]
 pub struct Platform {
     pub rom: Range<Addr>,
-    // This is an `Option` to allow `const` here.
-    pub prog_data: Option<HashSet<Addr>>,
```
matthiasgoergens, line 14: The minor problem here was that `None` and `Some(empty_hash_set)` are two different ways to express the same thing.

ceno_zkvm/src/scheme/mock_prover.rs
yczhangsjtu: Which one (`const` or `static`) is leaking? Why does this redefinition make a difference?

matthiasgoergens: So `const` in Rust is a bit like a macro in C: you can pretend that wherever you use a const, in the source code Rust copies in the definition of the `const`.  That's not leaking (at least not any more than copying in the definition manually would.)

In contrast `static` reserves memory for a single copy of the variable at the level of the process itself.  That single copy always stays around for as long as your process is running.  In that sense, it's leaking that single copy.

Does that make sense?

The natural way to write the definition here would be `pub const MOCK_PC_START: ByteAddr = ByteAddr(ceno_emul::CENO_PLATFORM.pc_base());`, exactly as we have before.  But because we removed the option layer, we'd now get:

```
error[E0493]: destructor of `Platform` cannot be evaluated at compile-time
  --> ceno_zkvm/src/scheme/mock_prover.rs:41:46
   |
41 | pub const MOCK_PC_START: ByteAddr = ByteAddr(ceno_emul::CENO_PLATFORM.pc_base());
   |                                              ^^^^^^^^^^^^^^^^^^^^^^^^          - value is dropped here
   |                                              |
   |                                              the destructor for this type cannot be evaluated in constants

For more information about this error, try `rustc --explain E0493`.
error: could not compile `ceno_zkvm` (lib) due to 1 previous error
```

That's exactly because a `const` can't leak.  That means, we'd need to call the destructor (when we sort-of copy-and-paste the definition of the `const` into where we use it.)  Moving to a `static` avoids having to call that destructor, but also leaks exactly one instance of the variable per process.

Does this make sense?

matthiasgoergens: I added an explanation in a comment.

yczhangsjtu: Crystal clear. Thanks!

yczhangsjtu (approved):

PR #759, opened 2024-12-14
Standard library support for guests

matthiasgoergens:
Standard library support for guests works, but this PR needs some polishing and depends on a lot of other PRs that should go in first, to make this one easier to review.  Nevertheless, you can have a look and try this one out, if you want.


ceno_rt/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    unsafe {
+        asm!(
+            "ecall",
+            in ("a0") exit_code,
+            in ("t0") 0,
+        );
+        unreachable!();
+    }
+    #[cfg(not(target_arch = "riscv32"))]
+    unimplemented!("Halt is not implemented for this target: {}", exit_code);
```
lispc, line 36: `unimplemented!("Halt is not implemented for this target, exit_code: {}", exit_code);` would be easier to understand

examples/.cargo/riscv32im-ceno-zkvm-elf.json
```diff
@@ -0,0 +1,31 @@
+{
```
lispc, line 1: in future, when users run `ceno-cli init-project ...`,   will this `riscv32im-ceno-zkvm-elf.json` be put into their workspace, (allow editting), or actually this file is not something they can change, bounded to "ceno plaform"?

for the latter case, this file should be in `rt` folder instead of `example` folder?

lispc:
1. we still need to add an example where it uses a std lib?  eg std::BTreeMap to do something? as a test case.
2. what is the high level summary on the magic how std supported is done?   seems like some building script magic, and main entry point asm magic?

ceno_rt/src/lib.rs
matthiasgoergens: Good idea!

ceno_host/tests/test_elf.rs
```diff
@@ -11,38 +11,62 @@ use itertools::enumerate;
 #[test]
 fn test_ceno_rt_mini() -> Result<()> {
     let program_elf = ceno_examples::ceno_rt_mini;
-    let mut state = VMState::new_from_elf(CENO_PLATFORM, program_elf)?;
+    let program = Program::load_elf(program_elf, u32::MAX)?;
```
matthiasgoergens, line 14: This snippet is repeated a lot.  TODO: deduplicate.

examples/.cargo/riscv32im-ceno-zkvm-elf.json
matthiasgoergens: We do not expect users to edit this file.  (And I need to add more documentation for how I made this file, too.  Because the format can change slightly over time as the Rust compiler and llvm evolve.  In fact, it changed already over the last year.)

Yes, something like the `ceno_rt` folder makes conceptual sense.  The main thing is that `examples/.cargo/config.toml` needs to refer to this file by location.

I moved it to where you suggested.

matthiasgoergens:
> 1. we still need to add an example where it uses a std lib?  eg std::BTreeMap to do something? as a test case.

I added `examples/examples/sorting.rs` to show that we are using the standard libary's `Vec` and its sorting function.

You can get `Vec` via `alloc` without the full standard library, but so can you get `alloc::collections:btree_map::BTreeMap`.

>     2. what is the high level summary on the magic how std supported is done?   seems like some building script magic, and main entry point asm magic?

Basically, yes.  Let me write up a bit, now that the diff against is a bit easier to read (even for myself).

The main goal of that write-up is to make sure other people can take this up and maintain it.

ceno_host/Cargo.toml
```diff
@@ -13,7 +13,7 @@ version.workspace = true
 anyhow.workspace = true
 ceno_emul = { path = "../ceno_emul" }
 itertools.workspace = true
-rkyv = { version = "0.8", default-features = false, features = ["alloc", "bytecheck"] }
+rkyv = { version = "0.8", features = ["pointer_width_32"] }
```
matthiasgoergens, line 16: We use `std` by default now.

ceno_host/tests/test_elf.rs
```diff
@@ -105,3 +129,39 @@ fn run(state: &mut VMState) -> Result<Vec<StepRecord>> {
     eprintln!("Emulator ran for {} steps.", steps.len());
     Ok(steps)
 }
+
+#[test]
+fn test_sorting() -> Result<()> {
```
matthiasgoergens, line 134: These two examples are pulled in from https://github.com/scroll-tech/ceno/pull/753 to show that we can actively use the standard library.

ceno_rt/riscv32im-ceno-zkvm-elf.json
```diff
@@ -0,0 +1,32 @@
+{
+  "arch": "riscv32",
+  "atomic-cas": true,
+  "cpu": "generic-rv32",
+  "crt-objects-fallback": "false",
+  "data-layout": "e-m:e-p:32:32-i64:64-n32-S128",
+  "disable-redzone": false,
```
matthiasgoergens, line 7: I'd love to add comments to explain the entries, but `json` doesn't let me.  Let me think of a way to overcome that.

rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "nightly-2024-12-06"
+components = ["rust-src"]
```
matthiasgoergens, line 3: We need the sources for `build-std`.

examples/.cargo/riscv32im-ceno-zkvm-elf.json
matthiasgoergens: The problem that I see is that I don't know how a standalone guest program would refer to `riscv32im-ceno-zkvm-elf.json` that lives in `ceno_rt`, instead of in its the guest's own repository.

examples/examples/hashing.rs
matthiasgoergens, line 1: @lispc Here's an example that really makes use of the standard library and couldn't be done without.

ceno_host/tests/test_elf.rs
lispc: indeed..

lispc (approved):

PR #765, opened 2024-12-16
Remove unnecessary `_stack_start` from Rust code

matthiasgoergens:
Just like `__global_pointer` or `_start`, we don't need to mention `_stack_start` in Rust to make use of it in the assembly.

### DIFF
--- a/ceno_rt/ceno_link.x
+++ b/ceno_rt/ceno_link.x
@@ -1,4 +1,5 @@
 
+/* The address of this variable is the start of the stack (growing downwards). */
 _stack_start = ORIGIN(REGION_STACK) + LENGTH(REGION_STACK);
 _hints_start = ORIGIN(REGION_HINTS);
 _hints_length = LENGTH(REGION_HINTS);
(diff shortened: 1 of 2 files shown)


lispc:
i am not sure whether it is a good idea.  Will the extern C code block make IDE happier?

matthiasgoergens:
Well, as mentioned in the PR description, there's a bunch of other symbols we only take from the linker, too. So `_stack_start` ain't special.

We should either define all of them here redundantly, or none.

(In general, this file should only be edited by people who have looked at least a bit into the linker scripts.)

lispc:
i prefer putting them all in extern C

matthiasgoergens:
> i prefer putting them all in extern C

OK, I just did some experiments.  For example, if you mistype the name of the variable in the `extern "C"` declaration like this:

```rust
extern "C" {
    // The address of this variable is the start of the stack (growing downwards).
    static _stack_start_this_is_a_typo: u8;
}
```

then you don't get an error message.  So it seems like having this declaration is no better than just adding a comment, and perhaps worse, because everyone knows that human readable comments are only there for humans and that the compiler doesn't check them.  But the `extern "C"` variable declaration looks like it might be checked, but ain't really.

So I suggest I'll add a comment to the assembly code instead.

PR #773, opened 2024-12-18
Remove ROM from platform

matthiasgoergens:
(Extracted from the work on the Rust standard library support for guests.)

Previously we had machinery that makes sure that executable code and data memory do not overlap.  But in a Harvard Archicture where code and data live in different address spaces anyway, that is not necessary.

(Our VM implements a Harvard Architecture.  Our circuits have always implemented that archicture, and the emulator has recently followed suit.)

Note how this PR does not remove any checks or constraints from the circuits.  Ie a sufficiently motivated and independent minded prover could have always done what this PR makes easier.

> The entry point is a property of the program and can be found in the ELF.

Btw, as far as I can tell, we do not currently check in the circuits whether the execution actually starts with the entry point.  Neither the entry point of the ELF nor the entry point of the Platform.  (And that's true before and after this PR.)


matthiasgoergens:
Converted to draft because of some issues around JAL and JALR and wrap around.

Issue #772, opened 2024-12-18
optimise wit_infer_by_expr util function


hero78119:
### Context
expression witness inferring util [wit_infer_by_expr](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/scheme/utils.rs#L236) both used in in [opcode](https://github.com/scroll-tech/ceno/blob/34a1ee16740f8edb34cd28f727c63b8c8fadbb73/ceno_zkvm/src/scheme/prover.rs#L248)/table proof 
https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/scheme/prover.rs#L248
occupied quite lot of time, e.g. 

with command `cargo run --release --package ceno_zkvm --bin e2e -- --profiling=3 --max-steps=1048576 --platform=sp1 ceno_zkvm/examples/fibonacci.elf`

and take "ADD" opcode as example
```
TRACE       ?? create_opcode_proof [ 574ms | 0.03% / 15.39% ] circuit_name: "ADD"
TRACE       ?  ?? wit_inference [ 273ms | 7.30% ] profiling_3: true
TRACE       ?  ?? SUMCHECK [ 186ms | 4.97% ] profiling_3: true
TRACE       ?  ?? witin::evals [ 12.9ms | 0.35% ] profiling_3: true
TRACE       ?  ?? pcs_open [ 102ms | 2.74% ] profiling_3: true
```
the wit_inference inferring occupied around 50%.

Preliminary review and the overhead probably caused by dy

### Idea of optimisation
- allocate just one mutable vector and do in-placement change during witness inferrence 
- leverage grid-stride-loop to traverse vector

### The (simulated) gain of idea
In this commit
https://github.com/scroll-tech/ceno/compare/master...hero78119:ceno:feat/wit_infer_opt
- we keep the whole traverse (but not updating mutable vector, as we need to figure out how to pass mutable vector during expression evaluation
- the traverse change to grid-stride-loop style

and the benchmark result
```
fibonacci_max_steps_1048576/prove_fibonacci/fibonacci_max_steps_1048576
                        time:   [4.1000 s 4.1310 s 4.1591 s]
                        change: [-16.733% -15.242% -13.712%] (p = 0.00 < 0.05)
                        Performance has improved.
```

So this optimisation show great potiential and bring up to 15% latency reduction

PR #775, opened 2024-12-19
accelerate wit_infer_by_expr via grid-strde-loop + vector pool

hero78119:
related #772 

This PR re-design `wit_infer_by_expr` to try grid-strde-loop traverse + reuse a simple vector pool to make vector reuse across closure across type boundary. The design is complex, since we need to deal with base/extension field and maximized code reuse.

### DIFF
--- a/ceno_zkvm/src/scheme/utils.rs
+++ b/ceno_zkvm/src/scheme/utils.rs
@@ -470,50 +470,60 @@ pub(crate) fn wit_infer_by_expr_in_place<'a, E: ExtensionField, const N: usize>(
             )
         },
         &|a, b, pool_e, pool_b| {
-            commutative_op_mle_pair!(|a, b| {
-                match (a.len(), b.len()) {
-                    (1, 1) => Arc::new(DenseMultilinearExtension::from_evaluation_vec_smart(
-                        0,
-                        vec![a[0] * b[0]],
-                    )),
-                    (1, _) => {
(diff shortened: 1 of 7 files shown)


hero78119:
There are many code redundant since I still do the experiment and see what the best design and avoid massive change

ceno_zkvm/src/scheme/utils.rs
```diff
@@ -233,118 +239,248 @@ pub(crate) fn infer_tower_product_witness<E: ExtensionField>(
...
+    pool_e: &mut SimpleVecPool<Vec<E>>,
+    pool_b: &mut SimpleVecPool<Vec<E::BaseField>>,
+    pool_expected_size_vec: usize,
+) {
+    fn downcast_arc<E: ExtensionField>(
+        arc: ArcMultilinearExtension<'_, E>,
+    ) -> DenseMultilinearExtension<E> {
+        unsafe {
+            // get the raw pointer from the Arc
+            assert_eq!(Arc::strong_count(&arc), 1);
```
hero78119, line 253: This is to
- follow `ArcMultilinearExtension<'_, E>` type across the whole interface
- however, there are some intermediate `ArcMultilinearExtension<'_, E>` which is hold some intermediate vector from vector pool, and we assure it can be return. However, I got `double free or corruption` error

hero78119:
It seems I am test on the wrong branch, will update it once benchmark result is confirmed

PR #776, opened 2024-12-19
benchmark fix: use iter_custom to measure routine time exactly

hero78119:
### context
the issue was identified when running sumcheck benchmark with setup but empty routine. The benchmark results shows there still ~60ms even routine is empty due to time was also measure something hidden out from desired routine

### solution
by criterion [docs](https://github.com/bheisler/criterion.rs/blob/0.5.0/src/bencher.rs#L24-L33). `iter_with_setup` will measure the time when setup object implementing `Drop`, which might be the root cause. The most accurate way I found is switch to use `iter_custom` so we can control elapsed time accurately.

### DIFF
--- a/ceno_zkvm/benches/fibonacci.rs
+++ b/ceno_zkvm/benches/fibonacci.rs
@@ -1,8 +1,4 @@
-use std::{
-    fs,
-    path::PathBuf,
-    time::{Duration, Instant},
-};
+use std::{fs, path::PathBuf, time::Duration};
 
 use ceno_emul::{Platform, Program};
 use ceno_zkvm::{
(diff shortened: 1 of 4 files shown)


mcalancea:
Useful information, thank you. I'm a bit confused at first glance: `iter_custom` seems to return the total time taken by all the samples. How does it retrieve the timing for each individual sample?

hero78119:
> Useful information, thank you. I'm a bit confused at first glance: `iter_custom` seems to return the total time taken by all the samples. How does it retrieve the timing for each individual sample?

The framework will measure individual sample by `total time of a sample` divided by `number of iteration` automatically

mcalancea (approved):

PR #785, opened 2024-12-23
docs: Corrected naming of RISC-V architecture Update README.md

0xwitty:
I noticed that the term "Risc-V" was used instead of the official "**RISC-V**" in several places, including the title and a section of the documentation.

I've updated these instances to ensure consistency with the official naming convention of the architecture.


matthiasgoergens:
Thanks for your PR.  We prefer the non-standard spelling.

Issue #788, opened 2024-12-26
experiment fanin > 2 in product argument  


hero78119:
### Context
Currently in ceno, product argument configure with fanin=2, e.g. [read set witness](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/scheme/prover.rs#L280). with respective sumcheck arguments $p[\vec{\gamma}] = \sum_{\vec{b}} eq(\vec{\gamma}, \vec{b}) p_{1}[\vec{b}] * p_{2}[\vec{b}]$. The reason is because to simplify, in sumcheck [prover_v2](https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/prover_v2.rs#L778) we only support degree <= 3 implementation so far.

We can extend product argument and experiment > 2 fanins on product and benchmark. To eliminate the limitation of degree sumcheck prover, we can use rust `macro` to generate the boilerplate code of arbitrary degree implementation.

### performance gain
Intuitively
1. compiler can do more optimisation https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/prover_v2.rs#L469-L473 if goes from `f1 * f2`  to `f1 * f2 * f3 * f4` 
2. layer of tower sumcheck product argument can be half, for fanin = 2 -> fanin = 4

### Suggested step
- [ ] use macro to generate sumcheck protocol boipolate code here https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/prover_v2.rs#L432-L533
- [ ] add benchmark follow [this](https://github.com/scroll-tech/ceno/blob/master/sumcheck/benches/devirgo_sumcheck.rs) to compare sumcheck degree = 2 vs degree = 4 
- [ ] refactor prover/verifier to support product argument with larger degree

Issue #702, opened 2024-12-05
Proposal features: 16 -> 32 bit range check via LogUp


hero78119:
### Background
Ceno heavily reply on LogUp to do range check, e.g. 16 bit range check
<img width="166" alt="image" src="https://github.com/user-attachments/assets/a966ff7c-0d32-4882-8939-ebd63d012334">
See circuits statistis here https://github.com/scroll-tech/ceno/pull/585#issuecomment-2477865876

Most of lookup operation are contribution via 16 bit range check. Take `add` opcode as example, the 9 lookups are all contributes via
- rs1, rs2, rd offline memory check timestamp check => (3 * 2) = 6 of 16 bit range check
- rd splitted into 2 of 16 bit range check
- program fetch: 1 lookup
So in overall there are 6 + 2 + 1 = 9 lookup

If we do 32 bit range check
- rs1, rs2, rd offline memory check timestamp check => 3 lookup
- rd range: 1 lookup
- program fetch: 1 lookup
overall is 5 lookup, which across 2^3 = 8 boundary, so the expected latency would be cut to half. 

As a side effect, we also save bunch of `witin` for they are there for holding 16 limb.  

### Design Rationale
On logup formula right hand side, we have m(x) and T(x). One of nice property for 32 bits range check table T(x) is we can skip its commitment & PCS, for verifier can evaluate T(r) succinctly via [tricks](https://github.com/scroll-tech/ceno/issues/378#issuecomment-2436580306) here. So another challenge is how to deal with huge & sparse polynomial m(x)

Via [spartan paper](https://eprint.iacr.org/2019/550.pdf) p29 7.2.2. sparse polynomial commitment `SPARK`, we can view sparse m(x) into  tuple of 3 dense [(i, j, M(i,j))] and commit 3 dense polynomial respectively. Giving original m(x) dense size is 2^32. The insight magic of the `SPARK` is via split into i, j polynomial, each size just 2^16. So we break down a 2^32 bit sparse poly into 2 poly each with 2^16 size. Prover need to commit `i`, `j`, `M`, along with `read_ts_row`, `write_ts_row`, `audit_ts_row`, `read_ts_col`, `write_ts_col`, `audit_ts_col` related to SPARK protocol.

With SPARK, the e2e table proving flow will be like this
[...]

PR #162, opened 2024-08-23
chores: code cleanup and lint error fix on more packages

hero78119:
Go after PR #106 to fix `ceno_zkvm` lint error
- replace sumcheck `prover.rs` with `prover_v2.rs`
- replace mle  `virtual_poly.rs` with `virtual_poly_v2.rs`
- fix most of package lint errors
- ceno-zkvm lint errors pending fix from #106

PR #791, opened 2024-12-27
refactor to replace v1 by v2 and cleanup suffix

hero78119:
Previously to quick verify idea and avoid massive change, there are new functionality with suffix `_v2`.  After experiment with good result, long time ago all logic already stick to v2 version and no longer use v1. This PR clean up all leftover v1 version, do renaming and file replacement without modify existing logic.

In summary
- `sumcheck/src/prover_v2.rs` ->  `sumcheck/src/prover.rs`
- `multilinear_extensions/src/virtual_poly_v2.rs` -> `multilinear_extensions/src/virtual_poly.rs`
- clean up all `V2` suffix

This addressed previous out-dated PR https://github.com/scroll-tech/ceno/pull/162, and as a preparation for https://github.com/scroll-tech/ceno/issues/788, https://github.com/scroll-tech/ceno/issues/702


multilinear_extensions/src/test.rs
```diff
@@ -31,31 +31,6 @@ fn test_virtual_polynomial_additions() {
     }
 }
 
-#[test]
-fn test_virtual_polynomial_mul_by_mle() {
```
zemse, line 35: Is there a reason why this test is removed? It seems that only `mul_by_mle` is also removed (as it's used just for this test) but it could be kept if we want to keep this test.

hero78119: yes we dont need this functionality anymore, which is just some leftover.

zemse (approved):

Issue #789, opened 2024-12-26
skip PCS of range table


hero78119:
An follow-up enhancement after #654 , for we can skip fixed_commit & PCS of range table due to its well-form via structural_witness feature introduced in PR #740

https://github.com/scroll-tech/ceno/blob/4d3948f78cfd17e481b2f54c510627e7e869fcd7/ceno_zkvm/src/tables/range/range_impl.rs#L31

hero78119:
completed via #797

PR #797, opened 2024-12-30
Feat/skip fixed commit of range table

10to4:
Work for #789


ceno_zkvm/src/circuit_builder.rs
```diff
@@ -94,6 +94,10 @@ pub struct ConstraintSystem<E: ExtensionField> {
     pub num_fixed: usize,
     pub fixed_namespace_map: Vec<String>,
 
+    pub num_structural_fixed: usize,
+    pub structural_fixed_namespace_map: Vec<String>,
+    pub structural_fixed_len: Vec<usize>,
```
hero78119, line 99: I think we dont need to define new `structural_fixed`. Instead we can just reuse current `structural_witin`?

hero78119, line 99: I think we dont need to define new `structural_fixed`. Instead we can just reuse current `structural_witin`?

hero78119 (commented):
Thanks for the PR :) Just would like to raise a preliminary question before goes into the detail

ceno_zkvm/src/circuit_builder.rs
10to4: ok😆

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -709,10 +709,12 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProver<E, PCS> {
                     .collect::<Vec<ArcMultilinearExtension<E>>>()
             })
             .unwrap_or_default();
+        let structural_fixed = fixed.split_off(cs.num_fixed);
```
naure, line 712: Related to the comment about the matrix. The content could be computed here instead.

ceno_zkvm/src/tables/range/range_impl.rs
```diff
@@ -41,10 +41,14 @@ impl RangeTableConfig {
...
+        num_structural_fixed: usize,
         content: Vec<u64>,
     ) -> RowMajorMatrix<F> {
-        let mut fixed =
-            RowMajorMatrix::<F>::new(content.len(), num_fixed, InstancePaddingStrategy::Default);
+        let mut fixed = RowMajorMatrix::<F>::new(
+            content.len(),
+            num_fixed + num_structural_fixed,
+            InstancePaddingStrategy::Default,
+        );
```
naure, line 51: `set_fixed_val!(row, self.fixed, F::from(i));`

There is some potential confusion between the indexes of `Fixed` and `StructuralFixed`. It seems that the idea is to concatenate first Fixed, then StructuralFixed?

- Technically the index should be: `num_fixed + self.fixed.0` to use in `set_fixed_val`
- Or assert `num_fixed` is 0 with a comment.
- Or as said in other comments, move the structural content to the prover code.

ceno_zkvm/src/expression.rs
```diff
@@ -791,6 +808,9 @@ pub struct StructuralWitIn {
 #[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub struct Fixed(pub usize);
 
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub struct StructuralFixed(pub usize, pub usize);
```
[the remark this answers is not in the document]
10to4: Ok, I have removed `StructuralFixed`.

ceno_zkvm/src/scheme/utils.rs
```diff
@@ -355,18 +357,20 @@ pub(crate) fn eval_by_expr<E: ExtensionField>(
     challenges: &[E],
     expr: &Expression<E>,
 ) -> E {
-    eval_by_expr_with_fixed(&[], witnesses, structural_witnesses, challenges, expr)
+    eval_by_expr_with_fixed(&[], &[], witnesses, structural_witnesses, challenges, expr)
 }
 
 pub(crate) fn eval_by_expr_with_fixed<E: ExtensionField>(
     fixed: &[E],
+    structural_fixed: &[E],
```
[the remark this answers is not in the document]
10to4: ok

ceno_zkvm/src/expression.rs
```diff
@@ -30,6 +30,8 @@ pub enum Expression<E: ExtensionField> {
     StructuralWitIn(WitnessId, usize, u32, usize),
     /// This multi-linear polynomial is known at the setup/keygen phase.
     Fixed(Fixed),
+    /// StructuralFixed is similar with Fixed, but it is structured.
+    StructuralFixed(StructuralFixed),
```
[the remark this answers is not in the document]
10to4: Ok, I have removed StructuralFixed.

ceno_zkvm/src/scheme/prover.rs
10to4: I have removed `StructuralFixed`.

ceno_zkvm/src/tables/range/range_impl.rs
10to4: I have removed `StructuralFixed`.

hero78119 (approved):
After latest change, overall LGTM now!

Also verified with benchmark, before/after no significant change on prover througput so it's good to go.

Just pointing out a further improvement

Issue #191, opened 2024-09-04
Sumcheck optimizations


dreamATD:
Implement optimizations the paper: [Some Improvements for the PIOP for ZeroCheck](https://eprint.iacr.org/2024/108).
- [ ] Support expression structure to allow more optimization.
  Proposed structure:
  ```rust
  pub enum PolyExpr<E: ExtensionField> {
      ConstBase(E::BaseField),
      ConstExt(E),
      MLEBase(usize),
      MLEExt(usize),
      Product(Vec<PolyExpr<E>>),
      Sum(Vec<PolyExpr<E>>),
  }
  ``` 
- [ ] Support zero-check special sumcheck.

hero78119:
Covered it #799

hero78119:
After discussion, more optimisation variants is tracked in #934

PR #799, opened 2024-12-31
Frontend and backend for building circuits

dreamATD:


naure (commented):
First pass on `gkr_iop`. It makes sense so far.

gkr_iop/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn gkr_witness(&self, phase1: &[Vec<E::BaseField>], challenges: &[E]) -> GKRCircuitWitness<E>
+    where
+        E: ExtensionField;
+}
+
+// TODO: the following trait consists of `commit_phase1`, `commit_phase2`, `gkr_phase` and `opening_phase`.
+pub trait ProtocolProver<E: ExtensionField, Trans: Transcript<E>, PCS> {}
+
+// TODO: the following trait consists of `commit_phase1`, `commit_phase2`, `gkr_phase` and `opening_phase`.
+pub trait ProtocolVerifier<E: ExtensionField, Trans: Transcript<E>, PCS> {}
```
dreamATD, line 59: @naure If I will implement those traits for `Chip` later, do you still think it's better to rename `Chip` to be `CircuitBuilder`?

naure: Or just add a `///` comment that explains all the responsibilities of the Chip type?

gkr_iop/examples/multi_layer_logup.rs
```diff
@@ -0,0 +1,283 @@
...
+struct TowerParams {
+    height: usize,
+}
+
+#[derive(Clone, Debug, Default)]
+struct TowerChipLayout<E> {
+    params: TowerParams,
+
+    // Commit poly indices.
+    committed_table: usize,
```
hero78119, line 32: we could just call `committed_table_id`, `committed_count_id`, more self-explanation

gkr_iop/examples/multi_layer_logup.rs
```diff
@@ -0,0 +1,283 @@
...
+        ));
+
+        chip.allocate_base_opening(self.committed_table, table.1);
+        chip.allocate_base_opening(self.committed_count, count);
+    }
+}
+
+pub struct TowerChipTrace {
+    pub table: Vec<u64>,
+    pub count: Vec<u64>,
```
hero78119, line 144: In ceno_zkvm codebase, we use `multiplicity` to represent `count`

gkr_iop/src/chip.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{evaluation::EvalExpression, gkr::layer::Layer};
+
+pub mod builder;
+pub mod protocol;
+
+/// Chip stores all information required in the GKR protocol, including the 
+/// commit phases, the GKR phase and the opening phase.
+#[derive(Clone, Debug, Default)]
+pub struct Chip {
```
hero78119, line 9: We can separate ChipBuilder vs ConstrainSystem.
Refer similar design in ceno_zkvm https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/circuit_builder.rs#L487-L488
All the information store in this chip struct should belong to `ConstrainSystem`.
So it can be Serializable.

similar as commment here https://github.com/scroll-tech/ceno/pull/799/files#r1904053980

subprotocols/src/expression.rs
```diff
@@ -0,0 +1,166 @@
...
+mod macros;
+
+pub type Point<E> = Arc<Vec<E>>;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum Constant {
+    /// Base field
+    Base(i64),
+    /// Challenge
+    Challenge(usize),
```
hero78119, line 17: It's a bit weird to put challenge in constant, but ok in this stage.
For later, I would propose to extend [expression](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/expression.rs#L668) in ceno_zkvm so they can be well support for pre-compile * ceno_zkvm. We did lots of optimisation on ceno_zkvm expression, e.g. we will optimize scalar & poly multiplication, e.g. `poly_1 * c1 * c2 * c3 ` into `poly * ((c1 * c2) * c3)`, for latter just O(N) with smaller constant factor

hero78119 (commented):
Awesome job! 
I leave few comments in separate section due to large PR so I did the review in segmented time.

Most of the utility of code reused can be done later, I think the most important point might be trying one pre-compile (e.g. keccak-f) first, and benchmark the preliminary performance. Once it meet the requirements, we proceed to more engineering polishing works :)

gkr_iop/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        (chip_spec, chip)
+    }
+
+    /// Specify the polynomials and challenges to be committed and generated in
+    /// Phase 1.
+    fn build_commit_phase1(&mut self, spec: &mut Chip);
+    /// Specify the polynomials and challenges to be committed and generated in
+    /// Phase 2.
+    fn build_commit_phase2(&mut self, _spec: &mut Chip) {}
```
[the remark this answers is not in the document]
dreamATD: Previously I created it just because Sp1 did it. I can remove it for now. If we need it in the future, we can add it back.

subprotocols/src/expression.rs
dreamATD: Sure, I agree. There is also an `Expression` in `mpcs` and it's better to unify the three structs :p.

subprotocols/src/expression/evaluate.rs
```diff
@@ -0,0 +1,458 @@
...
+                    base_mle_evals,
+                    out_points,
+                    in_point,
+                    challenges,
+                )
+                .pow([*d as u64]),
+        }
+    }
+
+    pub fn calc<E: ExtensionField>(
```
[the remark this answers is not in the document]
dreamATD: Are there any "parallel clone"? Here I only clone for single vector, of which the space will be reused when applying operations, such as follows:
```
define_commutative_op_mle2!(UniPolyVectorType, Mul, mul, |x, y| {
    zip_eq(&mut x, y).for_each(|(x, y)| zip_eq(x, y).for_each(|(x, y)| *x *= y));
    x
});
```

hero78119:
related to #191

hero78119 (commented):
Some quick question

gkr_iop/src/gkr/layer.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// expression corresponds to an output. While in sumcheck, there is only 1
+    /// expression, which corresponds to the sum of all outputs. This design is
+    /// for the convenience when building the following expression:
+    ///     `e_0 + beta * e_1 = sum_x (eq(p_0, x) + beta * eq(p_1, x)) expr(x)`.
+    /// where `vec![e_0, beta * e_1]` will be the output evaluation expressions.
+    pub exprs: Vec<Expression>,
+    /// Positions to place the evaluations of the base inputs of this layer.
+    pub in_bases: Vec<EvalExpression>,
+    /// Positions to place the evaluations of the ext inputs of this layer.
+    pub in_exts: Vec<EvalExpression>,
```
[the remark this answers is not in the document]
spherel: Originally, I saw there are many discussions in mpcs discussing whether each polynomial is base or ext. But actually when we batch commit polynomials, all polynomials in a batch are of the same type. I'm not sure how different the two methods are, but I'm okay with both way.

hero78119 (approved):
amazing work with many inspiring new designs :+1: !!

PR #802, opened 2025-01-03
Deduplicate magic syscall number

matthiasgoergens:
Also make examples visible to a plain `cargo clippy --workspace --all-targets`.  But that's now extracted into https://github.com/scroll-tech/ceno/pull/804 to be merged first.  Similar with https://github.com/scroll-tech/ceno/pull/803

We are using a technique inspired by the SP1 example already mentioned in `ceno_emul/src/syscalls.rs`.


ceno_rt/src/syscalls.rs
```diff
@@ -1,18 +1,18 @@
...
+
 /// Executes the Keccak256 permutation on the given state.
 ///
 /// ### Safety
 ///
 /// The caller must ensure that `state` is valid pointer to data that is aligned along a four
 /// byte boundary.
 #[allow(unused_variables)]
-#[no_mangle]
-pub extern "C" fn syscall_keccak_permute(state: &mut [u64; 25]) {
```
matthiasgoergens, line 15: We never use this in any context where we need the C calling convention.

In fact, making this a regular function makes the optimizers job a bit easier.

ceno_rt/src/syscalls.rs
```diff
@@ -1,18 +1,18 @@
...
+pub fn keccak_permute(state: &mut [u64; 25]) {
```
matthiasgoergens, line 14: No need to prefix the name of function with the name of module.  Just refer to the function as `syscalls::keccak_permute` to get the same effect.

ceno_emul/src/syscalls.rs
```diff
@@ -6,7 +6,7 @@ pub mod keccak_permute;
 // Using the same function codes as sp1:
 // https://github.com/succinctlabs/sp1/blob/013c24ea2fa15a0e7ed94f7d11a7ada4baa39ab9/crates/core/executor/src/syscalls/code.rs
 
-pub const KECCAK_PERMUTE: u32 = 0x00_01_01_09;
+pub use ceno_rt::syscalls::KECCAK_PERMUTE;
```
matthiasgoergens, line 9: No need to specify the magic number twice in our code, and risk it going out of sync in multiple places.

ceno_rt/src/syscalls.rs
```diff
@@ -1,18 +1,18 @@
 // Based on https://github.com/succinctlabs/sp1/blob/013c24ea2fa15a0e7ed94f7d11a7ada4baa39ab9/crates/zkvm/entrypoint/src/syscalls/keccak_permute.rs
-
-const KECCAK_PERMUTE: u32 = 0x00_01_01_09;
-
+#[cfg(target_os = "zkvm")]
 use core::arch::asm;
 
+pub const KECCAK_PERMUTE: u32 = 0x00_01_01_09;
```
matthiasgoergens, line 5: I just moved this one down, because we typically put `use` declarations above almost everything else in our files.

lispc (approved):
i prefer we merge this directly.  No need for so many tiny prs.

matthiasgoergens:
> i prefer we merge this directly. No need for so many tiny prs.

OK, can do.  I just wanted to make review easier.

PR #803, opened 2025-01-03
Remove obsolete `ceno_rt/Cargo.lock`

matthiasgoergens:
830b840b0db15c4e5ac2484acf19b038c1281d1f / PR#759 made `ceno_rt/Cargo.lock` obsolete, but we forgot deleting it back then.

Extracted from https://github.com/scroll-tech/ceno/pull/802


hero78119 (commented):
how about we also add `ceno_rt/Cargo.lock` into .gitignore explicitly

matthiasgoergens:
> how about we also add `ceno_rt/Cargo.lock` into .gitignore explicitly

No need, it will never be (re-) created, because `ceno_rt` is part of the overall workspace.

hero78119 (approved):

PR #805, opened 2025-01-06
migrate field from in-house goldilocks to plonky3

hero78119:
Related to #806

### Changes highlights
- compiles on base crates except for `mpcs`, `poseidon`. We will remove crate `poseidon` in the future, as right now only `mpcs` depends on it. 
- migrate `transcripts` to `p3::poseidon`. We can upgrade to `poseidon2` later.
- use `p3::Field` instead of `ff:Field`, and adjust respective api: `E::from` -> `from_canonical_u64`. `E::from -> from_base`

sumcheck/mle both all unittest passed.

PR #811, opened 2025-01-07
mpcs to plonky3 field

hero78119:
build on top of #805


hero78119:
With unittest test passed in below crate: mpcs/sumcheck/multilinear_extensions

`cargo test --package mpcs --package sumcheck --package multilinear_extensions  --lib --bins --tests --examples`

lispc (approved):
it seems good.
(btw the whir will be ready within this week. so i think then we will still a lot of things to do)

PR #813, opened 2025-01-07
`quadratic_sorting` and `is_prime` benchmarks

mcalancea:
This PR adds `benches/` files exercising the `quadratic_sorting` and `is_prime` examples (the latter added in this PR). The  benches target the `E2EProving` segment.

I've temporarily commented out a couple of things to make things work. They need to be addressed:
- Use of  `debug_memory_ranges` in `src/e2e.rs`.
- Use of `println!` in guests. The reason is that the contents of these prints are stored into a memory range which is not included in writable memory when using the default heap sizes.

Additionally, I've renamed `bubble_sorting` to `quadratic_sorting` which I think matches the guest code better. 

Later Edit:
- Confirmed with Aurel that `debug_memory_ranges` can be removed entirely.


matthiasgoergens:
Thanks!

We can completely remove the debug prints in the guests for now. Just make sure to have an assert or so inside the guest to make sure we got the right answers?

ceno_zkvm/benches/is_prime.rs
```diff
@@ -0,0 +1,76 @@
...
+                b.iter_custom(|iters| {
+                    let mut time = Duration::new(0, 0);
+
+                    for _ in 0..iters {
+                        let (_, prove) = run_e2e_with_checkpoint::<E, Pcs>(
+                            program.clone(),
+                            platform.clone(),
+                            hints.clone(),
+                            max_steps,
+                            Checkpoint::PrepE2EProving,
```
matthiasgoergens, line 61: Why do you stop here?

I was under the impression the goal is to benchmark the proving, not the running of the emulator.

matthiasgoergens (commented):
Let's see if GitHub works now..

PR #817, opened 2025-01-08
Log unused instructions, too

matthiasgoergens:
It's useful to see which Risc-V instructions aren't used at all when running eg sproll-evm.

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/rv32im.rs
+++ b/ceno_zkvm/src/instructions/riscv/rv32im.rs
@@ -29,12 +29,15 @@ use ceno_emul::{
 };
 use ecall::EcallDummy;
 use ff_ext::ExtensionField;
-use itertools::Itertools;
+use itertools::{Itertools, izip};
 use mul::{MulInstruction, MulhInstruction, MulhsuInstruction};
 use shift::SraInstruction;
 use slt::{SltInstruction, SltuInstruction};
 use slti::SltiuInstruction;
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -337,14 +340,10 @@ impl<E: ExtensionField> Rv32imConfig<E> {
             }
         });
 
-        for (insn_kind, (_, records)) in InsnKind::iter()
```
hero78119, line 340: (Not introduced by this PR) this block is only for statistics info, so can we make whole block only run when log level <= tracing?

matthiasgoergens: In principle, yes.  What are you trying to achieve with that suggested change?

The code here is only executed once, and not part of any inner loop etc; so performance is not an issue.

mcalancea (approved):

ceno_zkvm/src/instructions/riscv/rv32im.rs
hero78119: This code is on critical path, especially it go through all the instruction witness, it practice application it will up to 10M or more. We should avoid that and apparently it's just for debug info

hero78119: Ohoh, the result has been grouped by records, got it and there is no issue

hero78119 (approved):

PR #821, opened 2025-01-09
Clean up 'hints' for guests

matthiasgoergens:
Add the functionality to chain writes to the hints stream.


ceno_host/src/lib.rs
```diff
@@ -127,8 +127,10 @@ impl CenoStdin {
     pub fn write(
         &mut self,
         item: &impl for<'a> Serialize<HighSerializer<AlignedVec, ArenaHandle<'a>, Error>>,
-    ) -> Result<(), Error> {
-        to_bytes::<Error>(item).map(|bytes| self.write_slice(bytes))
+    ) -> Result<&mut Self, Error> {
+        to_bytes::<Error>(item)
+            .map(|bytes| self.write_slice(bytes))
```
lightsing, line 132: or maybe `write_slice` returns `&mut Self` too?

lightsing: also, although that Result::map is not lazy, this code looks really _SUSPICIOUS_ to me.

ceno_host/src/lib.rs
```diff
@@ -127,8 +127,10 @@ impl CenoStdin {
...
+            .map(|_| self)
```
lightsing, line 133:
```suggestion
        let bytes = to_bytes::<Error>(item)?;
        self.write_slice(bytes);
        Ok(self)
```

Let's be more explicit.

ceno_host/src/lib.rs
matthiasgoergens: > or maybe `write_slice` returns `&mut Self` too?

Good idea.  That also removes the need to `map` the `Result`.

lightsing (approved):

PR #822, opened 2025-01-14
Support for benchmarks

matthiasgoergens:
Extracted from https://github.com/scroll-tech/ceno/pull/809 because we need the non-controversial parts for other benchmarks.  That is anything but `pub const MAX_NUM_VARIABLES: usize = 27;`


ceno_emul/src/platform.rs
```diff
@@ -22,20 +22,14 @@ pub struct Platform {
...
             f,
-            "Platform {{ rom: {:?}, prog_data: {:?}, stack: {:?}, heap: {:?}, public_io: {:?}, hints: {:?}, unsafe_ecall_nop: {} }}",
-            self.rom,
-            prog_data,
-            self.stack,
-            self.heap,
-            self.public_io,
-            self.hints,
-            self.unsafe_ecall_nop
+            "Platform {{ rom: {:#x?}, prog_data: {prog_data}, stack: {:#x?}, heap: {:x?}, public_io: {:#x?}, hints: {:#x?}, unsafe_ecall_nop: {} }}",
```
matthiasgoergens, line 31: Hex is easier to read here (because we use round numbers, only when expressed in powers of 2), and it's also how we specify these numbers in our linker script etc.

ceno_zkvm/src/e2e.rs
```diff
@@ -45,6 +45,7 @@ pub struct EmulationResult {
     pi: PublicValues<u32>,
 }
 
+// TODO(Matthias): handle hints properly.
```
matthiasgoergens, line 48: This is something to fix soon.  But I don't know enough at the moment.

It's definitely broken in `master`, though.

ceno_zkvm/src/e2e.rs
```diff
@@ -574,8 +575,18 @@ fn debug_memory_ranges(vm: &VMState, mem_final: &[MemFinalRecord]) {
...
     );
 
-    for addr in &accessed_addrs {
-        assert!(handled_addrs.contains(addr), "unhandled addr: {:?}", addr);
+    let unhandled: BTreeSet<_> = accessed_addrs
+        .iter()
+        .filter(|addr| !handled_addrs.contains(addr))
+        .collect();
+
+    // TODO(Matthias): this should be an assert, but it's currently broken, because our caller doesn't handle hints.
```
hero78119, line 583: may you elaborate what the `caller` means in this context

matthiasgoergens: The function `emulate_program`, the only place that calls `debug_memory_ranges`.

matthiasgoergens: Do you want me to make that explicit in the comment in the code, too?

PR #833, opened 2025-02-20
upgrade to latest plonky3

hero78119:
Follow the new plonly3 [PR](https://github.com/Plonky3/Plonky3/pull/640) to use new api: "from_canonical_XXX" =>  "from_XXX"

Notes we can't update to latest p3 after this PR https://github.com/Plonky3/Plonky3/pull/653 because `rand~=0.9.x` version conflict with ark-std `rand~=0.8.x` https://github.com/arkworks-rs/std/blob/master/Cargo.toml#L16


hero78119:
Depends on this https://github.com/scroll-tech/ceno/pull/834 to fix test error: "filename too long"

PR #834, opened 2025-02-20
hash filename to make cache file shorter

hero78119:
This fix "Os file name too long" if challenge got to be large


lispc (approved):

lispc:
btw do you know how long does "load_tables" takes? if just several seconds, i think we don't need the complex save/load/cache mechanism...

hero78119:
> btw do you know how long does "load_tables" takes? if just several seconds, i think we don't need the complex save/load/cache mechanism...

yeah we have #205 #206 PR which shows that load_tables which occupied > 75% of unittest time, probably more than 5 secs. This mechanism in particular benefits when we run multiple unittests

lispc:
oh yes i realized the MockProver is run concurrently with dozens of unittests..

PR #835, opened 2025-02-20
upgrade to latest plonky3

hero78119:
Follow the new plonly3 [PR](https://github.com/Plonky3/Plonky3/pull/640) to use new api: "from_canonical_XXX" =>  "from_XXX", although the changes across multiple files but all are same purpose.

Notes we can't update to latest p3 after this PR https://github.com/Plonky3/Plonky3/pull/653 because the dependency `rand~=0.9.x` version are conflicted with ark-std `rand~=0.8.x` https://github.com/arkworks-rs/std/blob/master/Cargo.toml#L16 
So we need a separate task to remove ark-std from dependency, as it's just go from historical reason for we call it for tracking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,6 +771,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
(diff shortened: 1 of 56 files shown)


hero78119:
need to go after https://github.com/scroll-tech/ceno/pull/834

lispc (approved):

PR #854, opened 2025-03-06
bug fix: e2e missed proof validity check

hero78119:
previously end-to-end (e2e) tests were missing an assertion check for `verify_proof`, meaning that failed proofs due to the absence of a halt instruction were not properly detected.

Since the absence of a halt instruction can be a valid scenario (as execution can be controlled via max_step), this PR:
- Adds the missing verification check in e2e tests
- Makes the halt instruction optional

> Note: its doen't affect previous e2e testing result

### DIFF
--- a/ceno_zkvm/src/bin/e2e.rs
+++ b/ceno_zkvm/src/bin/e2e.rs
@@ -148,7 +148,11 @@ fn main() {
     // do statistics
     let stat_recorder = StatisticRecorder::default();
     let transcript = TranscriptWithStat::new(&stat_recorder, b"riscv");
-    verifier.verify_proof(zkvm_proof.clone(), transcript).ok();
+    assert!(
+        verifier
+            .verify_proof_halt(zkvm_proof.clone(), transcript, zkvm_proof.has_halt())
+            .is_ok()
+    );
     println!("e2e proof stat: {}", zkvm_proof);
     println!(
         "hashes count = {}",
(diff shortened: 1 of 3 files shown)


ceno_zkvm/src/bin/e2e.rs
```diff
@@ -148,7 +148,11 @@ fn main() {
     // do statistics
     let stat_recorder = StatisticRecorder::default();
     let transcript = TranscriptWithStat::new(&stat_recorder, b"riscv");
-    verifier.verify_proof(zkvm_proof.clone(), transcript).ok();
```
hero78119, line 151: this check was missing before fix

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -148,7 +148,11 @@ fn main() {
...
+    assert!(
+        verifier
+            .verify_proof_halt(zkvm_proof.clone(), transcript, zkvm_proof.has_halt())
```
kunxian-xia, line 153: If you set `expect_halt = zkvm_proof.has_halt()` then the new check that you added in this pr (i.e. `if has_halt != expect_halt`) will always pass, right?

ceno_zkvm/src/bin/e2e.rs
kunxian-xia: nice catch !

kunxian-xia (approved):

ceno_zkvm/src/bin/e2e.rs
hero78119: per offline discussion, `verify_proof_halt ` is internal api, thus here halt is optional.
while in `verify_proof` it will always set `true` and assert halt happenend

PR #864, opened 2025-03-18
Batched commitment of polys with different sizes on mpcs layer

Jiangkm3:
This PR enables ceno to batch commitment polynomials of different sizes on `mpcs` layer using any commitment scheme. The main contributions are:
- `pcs_batch_commit_diff_size`, `pcs_batch_commit_diff_size_and_write`, `pcs_batch_open_diff_size`, and `pcs_batch_verify_diff_size` in `mpcs/src/lib.rs` that implements the batching.
- `run_diff_size_batch_commit_open_verify` in `mpcs/src/lib.rs` and `batch_commit_diff_size_open_verify()` in `mpcs/src/basefold.rs` and `mpcs/src/whir.rs` for testing.

The idea is to use packing, which merges the polynomials into a sequence of large polynomials of a fixed size `packed_polys`. If the final few smallest polys cannot be fit into these packed polys, a final smaller polynomial `final_poly` is introduced. The process is then as follows:
1. Commit: the prover commits all `packed_polys` together (since they are of the same size) and `final_poly` separately.
2. Prove: the prover first uses a `unify_sumcheck` to reduce the different claim points on each poly down to the same point on each poly. It then performs packing to produce `packed_polys` and `final_poly` if exists. Finally it invokes the MPCS protocol to prove `packed_polys` and `final_poly` separately.
3. Verify: the verifier checks the correctness of `unify_sumcheck`, packing, and the MPCS.

### DIFF
--- a/mpcs/src/lib.rs
+++ b/mpcs/src/lib.rs
@@ -1,6 +1,6 @@
 #![deny(clippy::cargo)]
 use ff_ext::ExtensionField;
-use itertools::{interleave, Either, Itertools};
+use itertools::{Either, Itertools};
 use multilinear_extensions::{mle::{DenseMultilinearExtension, FieldType, MultilinearExtension}, virtual_poly::{build_eq_x_r, eq_eval, VPAuxInfo}};
 use serde::{Serialize, de::DeserializeOwned};
 use std::fmt::Debug;
@@ -10,6 +10,8 @@ use p3_field::PrimeCharacteristicRing;
 use multilinear_extensions::virtual_poly::VirtualPolynomial;
 use sumcheck::structs::{IOPProof, IOPProverState, IOPVerifierState};
 use witness::RowMajorMatrix;
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
 
 pub mod sum_check;
 pub mod util;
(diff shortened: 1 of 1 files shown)


Jiangkm3:
The current implementation is still rudimentary: it does not support parallel execution or has any optimization what so ever. I'm also awaiting for a PR on sumcheck to rewrite the sumcheck implementation to match with those in `ceno_zkvm`. However, this version is functioning and all tests passed.

hero78119:
Hi @Jiangkm3 could you elaborate primitive in your design from verifier perspective: giving 
- a unified pack point $r$
- 2 polynomial pair ($eval_1$, $r$[..num_var_1]), ($eval_2$, $r$[..num_var_2]), num_var_1 != num_var_2 and arbitrary.
Then how is formula to combine both into a single (eval_3, num_var_3)?

Jiangkm3:
Updated the approach to match with the new suffix-alignment of sumcheck. See the new interleaving approach (https://github.com/scroll-tech/ceno/blob/a0f20c013c233a52089c43463f89a6943fbd5fe4/mpcs/src/lib.rs#L80).

Jiangkm3:
New interleave function here:
https://github.com/scroll-tech/ceno/blob/1c504a522cd3d922152ee263f16dd1f563b98066/mpcs/src/lib.rs#L172-L175

This allows the prover to compute interleave in `O(num_polys) + O(interleave_size)` time, which should be optimal.
This approach is also potentially parallelizable, although it wouldn't be easy. Still, much easier than the original approach above.

Jiangkm3:
Parallel version of interleave now available: https://github.com/scroll-tech/ceno/blob/3d8b8b93c65e11221e15b7cbfd0cd91c7f029f80/mpcs/src/lib.rs#L229-L231

PR #870, opened 2025-03-20
batched sumcheck suffix alignment

### LINKED ISSUE #867: [sumcheck] modify batched sumcheck final opening point alignment
This issue was initiative along with reviewing PR #864 , the first scenario we kick of the batched of sumcheck in different number of variables.

Basically due to the invesigation for why num_thread > 1 not work in this test case https://github.com/scroll-tech/ceno/blob/master/sumcheck/src/test.rs#L29-L30

In current design, batched sumcheck with different num_vars are implemented by times a scalar, see docs in code comment
https://github.com/scroll-tech/ceno/blob/a0b719d6d0fc323574c4ddc2bb68b7ba3bcfb8c1/sumcheck/src/prover.rs#L414-L425 
The methodology works by starting binding variable on all polys, so some poly with smaller num_var will be ended earlier.
In another word, giving the final opening point $\vec{r}$ , the final reduce opening point for each polynomial will be $\vec{r}$[..num_var_i], so are aligned on "prefix"

However, batch different num of variable in prefix alignment, doesn't fit in devirgo style sumcheck, due to we break down sumcheck into 2 phase. In first phase, we break down evaluation into #threads section, and all polynomial bind as "constant" right at the last round of phase 1. This prefix-alignment works well when all poly are in same num_var. However, when poly are in different number of variable, some poly with less num_var in prefix alignment will be end much earlier, which break the rule for all polynomial should be end together in phase 1.

To address the issue, we need to modify the sumcheck design to be  right alignment. Thus, giving same opening point $\vec{r}$, each poly will be $\vec{r}$[r.len() - num_var_i..], where we call right alignment. With that, poly with less num_var will start to bind lately, and they can end together at the last round of phase 1 sumcheck.

cc @Jiangkm3

hero78119:
To close #867 

New design only apply on prover, as verifier logic remains unchanged

### benchmark 
original e2e benchmark result remain no change, which is expected
```
                        time:   [3.9061 s 3.9327 s 3.9608 s]
                        change: [-2.0470% -1.2697% -0.4292%] (p = 0.01 < 0.05)
```


sumcheck_macro/src/lib.rs
```diff
@@ -219,33 +219,69 @@ pub fn sumcheck_code_gen(input: proc_macro::TokenStream) -> proc_macro::TokenStr
...
-                res
-            };
-            let num_vars_multiplicity = self.poly.aux_info.max_num_variables - (ceil_log2(v1.len()).max(1) + self.round - 1);
-            if num_vars_multiplicity > 0 {
-                AdditiveArray(res.0.map(|e| e * E::BaseField::from_u64(1 << num_vars_multiplicity)))
+            // To deal with different num_vars, we exploit a fact that for each product which num_vars < max_num_vars
+            // we actually need to have a full sum, times 2^(bh_num_vars - num_vars) to accumulate into univariate computation
+            // E.g. Giving multivariate poly f(X) = f_1(X1) + f_2(X), X1 \in {F}^{n'}, X \in {F}^{n}, |X1| := n', |X| = n, n' <= n
+            // For i < n - n', to compute univariate poly, f^i(x), b is i-th round boolean hypercube
+            // f^i[0] = \sum_b f(r, 0, b), b \in {0, 1}^{n-i-1}, r \in {F}^{n-i-1} chanllenge get from prev rounds
```
spherel, line 246: `chanllenge` => `challenge`?

sumcheck_macro/src/lib.rs
```diff
@@ -219,33 +219,69 @@ pub fn sumcheck_code_gen(input: proc_macro::TokenStream) -> proc_macro::TokenStr
...
+                }
+                AdditiveArray::<_, #degree_plus_one>([sum; #degree_plus_one])
             } else {
-                res
+                let res = (0..largest_even_below(v1.len()))
+                    #iter
+                    .map(|b| {
+                        #additive_array_items
+                    })
+                    .sum::<AdditiveArray<_, #degree_plus_one>>();
```
spherel, line 277: This only occurs in the `else` branch. Could it be moved there?

spherel (approved):
LGTM!

Issue #632, opened 2024-11-26
[debug] support guest program "println!" 


lispc:
host(emulator) will handle the syscall, and print the content to stdin.

matthiasgoergens:
I would also suggest making this a no-op in the circuits.

naure:
There is a demo of this as a starting point:

https://github.com/scroll-tech/ceno/blob/1edbf65010274acee38d445beba236e7293c66b4/examples/examples/ceno_rt_io.rs#L10

https://github.com/scroll-tech/ceno/blob/1edbf65010274acee38d445beba236e7293c66b4/ceno_rt/src/io.rs#L72

https://github.com/scroll-tech/ceno/blob/1edbf65010274acee38d445beba236e7293c66b4/ceno_emul/tests/test_elf.rs#L70

From there, to avoid the proving cost of even generating the message, we can use a a `debug!` macro (conditional prints or conditional compilation), and run the program in debug or production mode.

SamarthJ2004:
Hi,  I would like to work on this.

SamarthJ2004:
Hi, I opened a pull request trying to do this but I am facing a problem. In the debug mode I am unable to capture the messages. I tried changing the opt-level in .toml to 0 as well but couldn't solve it.

hero78119:
> Hi, I opened a pull request trying to do this but I am facing a problem. In the debug mode I am unable to capture the messages. I tried changing the opt-level in .toml to 0 as well but couldn't solve it.
> 
> Kindly guide me.

Hi @SamarthJ2004 may you elaborate more detail about the question you have, such as there is no log even before the change? or the assembly code doesn't being preserved by compiler even for debug build?

hero78119:
> > Hi, I opened a pull request trying to do this but I am facing a problem. In the debug mode I am unable to capture the messages. I tried changing the opt-level in .toml to 0 as well but couldn't solve it.
> > Kindly guide me.
> 
> Hi [@SamarthJ2004](https://github.com/SamarthJ2004) may you elaborate more detail about the question you have, such as there is no log even before the change? or the assembly code doesn't being preserved by compiler even for debug build?

Hi @SamarthJ2004 sorry to come back to your question lately. Previously it doen't work because guest program compilation doesn't respect profile, see here
https://github.com/scroll-tech/ceno/pull/929/files#r2075093153

Will make this issue as completed, as #929 cover full features along with other stuff.

PR #871, opened 2025-03-22
Issue632

SamarthJ2004:
Added the debug_print marco and used debug_assertions for conditional message printing.

#632


hero78119:
Thanks for the PR.
Will close it due to feature covered in PR #929 :)

PR #873, opened 2025-03-24
[Experiment] ceno + babybear

hero78119:
to generete mock proof for recursion.

PR #858, opened 2025-03-13
WHIR switch to Plonky3

yczhangsjtu:
Compilation and tests passed.

Current benchmark results:
- 64 Polys with 20 variables
  - Commit: 5.9639 s
  - Open: 199.75 ms
  - Verify: 55.735 ms
  - Proof size: 3395312 Bytes
- 1 Poly with 20 variables
  - Commit: 87.098 ms
  - Open: 165.82 ms
  - Verify: 14.413
  - Proof size: 169712 Bytes

PR #843, opened 2025-03-04
WIP mpcs commit & opening code refactor cleanup

hero78119:
Take basefold mpcs as pilot, then apply to WHIR once plonky3 migration is ready

PR #875, opened 2025-03-25
field agnostic refactor (part 1)

hero78119:
extract from #873

### DIFF
--- a/ff_ext/src/babybear.rs
+++ b/ff_ext/src/babybear.rs
@@ -92,8 +92,6 @@ pub mod impl_babybear {
     }
 
     impl PoseidonField for BabyBear {
-        const PERM_WIDTH: usize = POSEIDON2_BABYBEAR_WIDTH;
-        const RATE: usize = POSEIDON2_BABYBEAR_RATE;
         type P = Poseidon2BabyBear<POSEIDON2_BABYBEAR_WIDTH>;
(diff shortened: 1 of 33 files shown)


ff_ext/src/babybear.rs
```diff
@@ -0,0 +1,184 @@
...
+    impl_from_uniform_bytes_for_binomial_extension!(p3::babybear::BabyBear, 4);
+
+    impl ExtensionField for BabyBearExt4 {
+        const DEGREE: usize = 4;
+        const MULTIPLICATIVE_GENERATOR: Self = <BabyBearExt4 as Field>::GENERATOR;
+        const TWO_ADICITY: usize = BabyBear::TWO_ADICITY;
+        // Passing two-adacity itself to this function will get the root of unity
+        // with the largest order, i.e., order = 2^two-adacity.
+        const BASE_TWO_ADIC_ROOT_OF_UNITY: Self::BaseField = BabyBear::new(0x78000000);
+        const TWO_ADIC_ROOT_OF_UNITY: Self =
```
hero78119, line 155: we can remove this after  #858 #843, as this is only for field wrapper to serve ark lib

lispc (approved):

ff_ext/src/babybear.rs
kunxian-xia: This is removed in #966.

Issue #876, opened 2025-03-25
feat: ceno cli


lispc:
like openvm, we add commands like

```
cargo ceno build
cargo ceno keygen
cargo ceno run
cargo ceno prove
cargo ceno verify
```

hero78119:
cli functionality also related to #709

cyberbono3:
Hi. Can I start working on this?

lispc:
> Hi. [@hero78119](https://github.com/hero78119) Can I start working on this?

Hi @cyberbono3 !   Thanks for the reply! But we decided to implement this feature ourselves now.

PR #877, opened 2025-03-25
feat: add serde/deserde to vk and proof

lispc:
a small step towards https://github.com/scroll-tech/ceno/issues/876


ceno_rt/memory.x
```diff
@@ -2,7 +2,7 @@ MEMORY
 {
   RAM : ORIGIN = 0x80000000, LENGTH = 1024M
   ROM : ORIGIN = 0x20000000, LENGTH = 16M
-  HINTS: ORIGIN = 0x40000000, LENGTH = 1024M
+  HINTS : ORIGIN = 0x40000000, LENGTH = 1024M
```
lispc, line 5: newer rust toolchain needs this fix.  just bring it in.

multilinear_extensions/src/mle.rs
```diff
@@ -123,14 +123,13 @@ impl<F: Field, E: ExtensionField<BaseField = F>> IntoMLEs<DenseMultilinearExtens
...
 #[derive(Clone, PartialEq, Eq, Hash, Default, Debug, Serialize, Deserialize)]
-#[serde(untagged)]
 #[serde(bound(
     serialize = "E::BaseField: Serialize",
     deserialize = "E::BaseField: DeserializeOwned"
 ))]
 /// Differentiate inner vector on base/extension field.
 pub enum FieldType<E: ExtensionField> {
-    Base(#[serde(skip)] Vec<E::BaseField>),
+    Base(Vec<E::BaseField>),
```
hero78119, line 132: understood this is to serve trivial proof for small polynomial 👍

ceno_zkvm/src/bin/e2e.rs
```diff
@@ -142,9 +151,25 @@ fn main() {
         Checkpoint::PrepSanityCheck,
     );
 
-    let (mut zkvm_proof, verifier) = state.expect("PrepSanityCheck should yield state.");
+    let (zkvm_proof, vk) = state.expect("PrepSanityCheck should yield state.");
+
+    let proof_bytes = bincode::serialize(&zkvm_proof).unwrap();
```
hero78119, line 156: Aware we have two serialization framework: bincode/rykv both serve similar purpose
I guess this is to align with openvm as we need to load proof in recursion?

hero78119 (approved):
LGTM 👍

multilinear_extensions/src/mle.rs
lispc: yes

ceno_zkvm/src/bin/e2e.rs
lispc: bincode is compatible with serde traits.  While rykv you need to derive its trait on every structs.. Unless we want extreme performance, we usually dont want to use rykv, the API is much more unfriendly.

Issue #878, opened 2025-03-25
implement public-io in SDK to serve guest program


hero78119:
This issue was related to #[353](https://github.com/scroll-tech/ceno/issues/353) to complete functionality of client program

PR #881, opened 2025-03-26
support public io/hint in guest program

hero78119:
To address issue #878

### change scope
- [x] clean up hardcode fibonacci elf from repo
- [x] support hint/public io 
- [x] add fibonacci source with ceno sdk to read from hint and commit with public io


hero78119:
Benchmark fibonacci 2^20 before/after on AMD EPYC 32 cores
```
                        time:   [3.9471 s 3.9732 s 4.0000 s]
                        change: [-1.3421% -0.5420% +0.2730%] (p = 0.23 > 0.05)
                        No change in performance detected.
```
shows no different which is expected

lispc (approved):

lispc:
i feel now finally we can add a section in top level README.md to show how to run ceno prove & verify e2e?

previously, either '--platform sp1' and the binary elf file seems a bit wired.

hero78119:
> i feel now finally we can add a section in top level README.md to show how to run ceno prove & verify e2e?
> 
> previously, either '--platform sp1' and the binary elf file seems a bit wired.

It make sense
I added in README to show how to run e2e example https://github.com/scroll-tech/ceno/blob/09f43c6912e245812b71f697fac9149d8b0574da/README.md

ceno_rt/src/mmio.rs
```diff
@@ -51,3 +51,51 @@ where
...
+    static mut NEXT_PUBIO_LEN_AT: *const usize = &raw const _lengths_of_pubio_start;
+    unsafe {
+        let len: usize = core::ptr::read(NEXT_PUBIO_LEN_AT);
+        NEXT_PUBIO_LEN_AT = NEXT_PUBIO_LEN_AT.add(1);
+        len
+    }
+}
+
+pub fn pubio_read_slice<'a>() -> &'a [u8] {
+    &pubio_region()[..pubio_len()]
```
naure, line 90: How is this mechanism intended to work?

As it stands each call to `pubio_read_slice` returns a slice starting at the same address, but with a different size, following the sequence in `_lengths_of_pubio_start` above.

hero78119: This follow by hints IO. I found it's very tricky here due to the right alignment of `rkyv`, and better having a clear documentation.

So the memory layout for pub io will be like
```
PubIOMem = [end_addr1, end_addr2, end_addr3, ....end_addrN, <rkyv serialized data 1>,  <rkyv serialized data 2>, ... <rkyv serialized data N>]
```
Where each end_addr point to the end address of each serialized data in one to one manner.

`_lengths_of_pubio_start` is a mutable pointer point, per public io read it will return next end_addr, then rkyv::access(PubIOMem[0....END]) will serialize those part of data from the end to start

naure: Got it, thanks.

PR #892, opened 2025-04-01
fix dependency chain of var.rs error

hero78119:
Addressed the root cause in #887, for previously I thought it's due to example not build thus add a new ci step for it.

Integration build still occasionally failed. Finally figure out the root cause - invalid generated file, which probably due to CI interrupt after PR update.

"build.rs" generate "var.rs", but if "var.rs" exist but invalid, "build.rs" will failed BEFORE compile example and generate "var.rs", which is a dead loop.

The fix is simple, via clean up "var.rs" before generate "var.rs".


lispc (approved):
ok...

PR #893, opened 2025-04-01
Mihai/keccakf integration

mcalancea:


ceno_zkvm/src/structs.rs
```diff
@@ -130,6 +134,13 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> VerifyingKey<E, PCS>
     }
 }
 
+#[derive(Clone)]
+pub struct KeccakGKRIOP<E> {
+    pub chip: gkr_iop::chip::Chip,
+    pub layout: KeccakLayout<E>,
+    pub circuit: gkr_iop::gkr::GKRCircuit,
```
spherel, line 141: If `GKRCircuit` can be derived from `Chip`? Could we remove this member?

ceno_zkvm/src/structs.rs
```diff
@@ -244,6 +284,7 @@ impl<E: ExtensionField> ZKVMFixedTraces<E> {
 
 #[derive(Default, Clone)]
 pub struct ZKVMWitnesses<E: ExtensionField> {
+    keccak_trace: <KeccakLayout<E> as gkr_iop::ProtocolWitnessGenerator<E>>::Trace,
```
spherel, line 287: It seems like other members are `E::BaseField` elements. Does it make sense that we also make it `Vec<Vec<E::BaseField>>`, which is the result after calling `phase1_witness()`?

ceno_zkvm/src/structs.rs
```diff
@@ -263,6 +304,44 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
...
+            .get_cs(&LargeEcallDummy::<E, KeccakSpec>::name())
+            .unwrap();
+        let (witness, logup_multiplicity) = LargeEcallDummy::<E, KeccakSpec>::assign_instances(
+            config,
+            cs.num_witin as usize,
+            records,
+        )?;
+
+        // GKR-IOP-specific trace from row major witness
+        self.keccak_trace = KeccakTrace::from(witness.clone());
```
spherel, line 324: Is it possible and better to just call `phase1_witness()` here?

spherel (approved):

PR #894, opened 2025-04-02
mpcs e2e batch in one proof

hero78119:
Build on top of https://github.com/scroll-tech/ceno/pull/901.

### Design rationale
zkvm + mpcs api: use `batch_commit/batch_opening` to commit/open **ALL** opcode together. In more detail, we commit `fixed_commit` and `witin_commit` separately with just one opening. 
basefold: batch opcodes with rmm with different "height", in other words, different "num_vars".

For basefold to batch opcodes with different variables, for sumcheck part we apply suffix alignment techniques based on 
 #870. For FRI part, we also apply suffix alignment techniques. So the smaller codeword will "involved" into the folding process when current length match.

During implementation, there is key principle in mind: verifier only rely on one untrusted information "num_instances" from prover,  and other information should derived from verifier key.

### Working items
- [x] mpcs batch api + batch basefold prover
- [x] batch basefold verifier

### Fibonacci Benchmark Results

| Size     | Branch               | Proof Size | E2E Latency Change | Prover Speed (kHz) |
|----------|----------------------|------------|--------------------|---------------------|
| 2²⁰      | Master               | 14.68 MB   | -                  | 363                 |
| 2²⁰      | Batched + conjecture        | 1.16 MB    | -14.195%           | 422                 |
| 2²⁰      | Batched  | 2.02 MB    | -11.599%                  | 411                 |
| 2²¹      | Master               | 15.53 MB   | -                  | 411                 |
| 2²¹      | Batched  + conjecture         | 1.24 MB    | -11.178%           | 451                 |
| 2²¹      | Batched   | 2.16 MB    | -6.4525%                  | 428                 |
| 2²²      | Master               | 16.42 MB   | -                  | 406                 |
| 2²²      | Batched + conjecture         | 1.31 MB    | -11.178%           | 433                 |
| 2²²      | Batched  | 2.31 MB    | -13.691%                 | 448                 |


mpcs/src/basefold.rs
```diff
@@ -213,91 +206,90 @@ where
...
-        let num_polys = polys.len();
+        let mmcs = poseidon2_merkle_tree::<E>();
 
-        let is_base = match polys[0].evaluations() {
-            FieldType::Ext(_) => false,
-            FieldType::Base(_) => true,
-            _ => unreachable!(),
-        };
+        let span = entered_span!("to_mles", profiling_3 = true);
+        let (polys, rmm_to_batch_commit, trivial_proofdata, circuit_codeword_index): (
```
kunxian-xia, line 220: nitpick: shall we rename `trivial_proofdata` to `rmms_without_encoding` or add some comment to make it clear?

mpcs/src/basefold/structure.rs
```diff
@@ -61,52 +64,73 @@ pub struct BasefoldCommitmentWithWitness<E: ExtensionField>
...
 {
+    pub fn new(
+        commit: Digest<E>,
+        codeword: MerkleTree<E::BaseField>,
+        polys: BTreeMap<usize, Vec<ArcMultilinearExtension<'static, E>>>,
+        trivial_proofdata: BTreeMap<usize, (Digest<E>, MerkleTree<E::BaseField>)>,
+        circuit_codeword_index: BTreeMap<usize, usize>,
+    ) -> Self {
+        let mmcs = poseidon2_merkle_tree::<E>();
+        // size = height * 2 because we concat pi[left]/pi[right] under same row index
```
kunxian-xia, line 90: nitpick: we need to clarify the meaning of `pi` here?

mpcs/src/basefold.rs
```diff
@@ -320,154 +312,379 @@ where
...
-        comm: &Self::CommitmentWithWitness,
-        point: &[E],
-        evals: &[E],
+        num_instances: &[(usize, usize)],
+        fixed_comms: Option<&Self::CommitmentWithWitness>,
+        witin_comms: &Self::CommitmentWithWitness,
+        points: &[Point<E>],
+        // TODO this is only for debug purpose
+        evals: &[Vec<E>],
+        circuit_num_polys: &[(usize, usize)],
```
kunxian-xia, line 323: why not change the type of `circuit_num_polys` to `BTreeMap<usize, (usize, usize)>` as the matrices/circuit's witness inside `witin_comms` are indexed by the circuit index?

hero78119: because `circuit_num_polys` deterministic and static information came from verifier key. Thus I feel a static information should avoid using `BTreeMap` which make room to include some invalid key. Actually I feel in the end we should change this to fixed size array, and so as other part.

mpcs/src/basefold.rs
```diff
@@ -320,154 +312,379 @@ where
...
-    fn simple_batch_open(
         pp: &Self::ProverParam,
-        polys: &[ArcMultilinearExtension<E>],
-        comm: &Self::CommitmentWithWitness,
-        point: &[E],
-        evals: &[E],
+        num_instances: &[(usize, usize)],
+        fixed_comms: Option<&Self::CommitmentWithWitness>,
+        witin_comms: &Self::CommitmentWithWitness,
+        points: &[Point<E>],
```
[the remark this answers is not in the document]
hero78119: Have different opinion: as golden rule, we should avoid using BTreeMap since the dynamism always make some room to include invalid key, thus in the end we need some strictly check. For `points`, in here we confirmed it's length will exactly one-to-one mapping with `witin_comms`. So we should iterate them with zip_eq() and always assure they are one to one mapping without rely on btreemap access.

mpcs/src/basefold/structure.rs
hero78119: Right and thanks for the check.

I removed the comment here besides switch to use `rmm.values.len()` so we dont need to explain why heightx2 here :)

kunxian-xia:
The batch open feature is very flexible, as we open multilinear polynomials of different number of variable (`nv`) at different points. 
We may want to divide this procedure into 3 steps:
1. for each nv, run a sumcheck to reduce the evaluations of MLPs (stored in matrices) on different points $z_i$ to the evaluation on single point $\alpha$ (check the proposal introduced in HyperPlonk paper #904).
2. for each nv, randomly combine MLPs into one MLP, and then open it on single point $\alpha$, we get an iterator of tuple $(p(x), \alpha)$;
3. In the commit phase, in each step, pop one tuple from the above iterator, do the folding around challenge $r$, and add the next item in the iterator if its nv + 1 equals to p(x)'s nv.

mpcs/src/basefold.rs
hero78119: yeah it make sense. comment added in BatchCommitmentWithWitness structure https://github.com/scroll-tech/ceno/pull/894/commits/d957832a01215de6cf8e57b02623e66ae18dcc47

hero78119:
Updated: to make verifier more robust, in latest 2 commits https://github.com/scroll-tech/ceno/pull/894/commits/f1c99776789fb69ab32e7273761442f4356e76be 
https://github.com/scroll-tech/ceno/pull/894/commits/50e390e7aca193e0a6b917e63bc5e5ae50069182 all `BTreeMap` related are removed and convert to pure vector, as all ordering can deterministic derived from global information `num_instances`

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +325,142 @@ where
...
+                .unwrap()
+                .values
+                .par_chunks_exact(2)
+                .zip(folding_coeffs)
+                .map(|(ys, coeff)| {
+                    let (left, right) = (ys[0], ys[1]);
+                    // original (left, right) = (lo + hi*x, lo - hi*x), lo, hi are codeword, but after times x it's not codeword
+                    // recover left & right codeword via (lo, hi) = ((left + right) / 2, (left - right) / 2x)
+                    let (lo, hi) = ((left + right) * inv_2, (left - right) * *coeff); // e.g. coeff = (2 * dit_butterfly)^(-1) in rs code
+                    // we do fold on folded = (1-r) * left_codeword + r * right_codeword, as it match perfectly with raw message in lagrange domain fixed variable
```
kunxian-xia, line 401: nitpick: should be "we do fold on (lo, hi) to get folded = (1-r) * lo + r* hi (with lo, hi are two codewords) ...."

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +325,142 @@ where
...
+                        .map(|codeword| {
+                            let (left, right) =
+                                (codeword.values[index], codeword.values[index + 1]);
+                            // original (left, right) = (lo + hi*x, lo - hi*x), lo, hi are codeword, but after times x it's not codeword
+                            // recover left & right codeword via (lo, hi) = ((left + right) / 2, (left - right) / 2x)
+                            let (lo, hi) = ((left + right) * inv_2, (left - right) * *coeff); // e.g. coeff = (2 * dit_butterfly)^(-1) in rs code
+                            // we do fold on folded = (1-r) * left_codeword + r * right_codeword, as it match perfectly with raw message in lagrange domain fixed variable
+                            lo + (hi - lo) * challenge
+                        })
+                        .sum::<E>();
```
kunxian-xia, line 428: Performance: this is slower than "1. do multiple codewords summation; 2. do fold with challenge"

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +325,142 @@ where
...
+                        .into_iter()
+                        .chain(codewords_matched.iter().map(|m| m.as_view()))
+                        .map(|codeword| {
+                            let (left, right) =
+                                (codeword.values[index], codeword.values[index + 1]);
+                            // original (left, right) = (lo + hi*x, lo - hi*x), lo, hi are codeword, but after times x it's not codeword
+                            // recover left & right codeword via (lo, hi) = ((left + right) / 2, (left - right) / 2x)
+                            let (lo, hi) = ((left + right) * inv_2, (left - right) * *coeff); // e.g. coeff = (2 * dit_butterfly)^(-1) in rs code
+                            // we do fold on folded = (1-r) * left_codeword + r * right_codeword, as it match perfectly with raw message in lagrange domain fixed variable
+                            lo + (hi - lo) * challenge
```
kunxian-xia, line 426: nitpick: we should abstract these lines of codes in a closure, as this can allow us to avoid the duplication.

mpcs/src/basefold/structure.rs
```diff
@@ -242,30 +274,11 @@ where
 {
     pub(crate) sumcheck_messages: Vec<Vec<E>>,
     pub(crate) commits: Vec<Digest<E>>,
-    pub(crate) final_message: Vec<E>,
+    pub(crate) final_message: Vec<Vec<E>>,
     pub(crate) query_opening_proof: QueryOpeningProofs<E>,
     pub(crate) sumcheck_proof: Option<Vec<IOPProverMessage<E>>>,
```
kunxian-xia, line 279: It seems `sumcheck_proof` is identical to `sumcheck_messages`, shall we remove `sumcheck_messages` field and keep `sumcheck_proof` only?

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +325,142 @@ where
...
+    });
+
+    // optimize for single codeword match
+    let folded_codeword = if (running_codeword_opt.is_some() as usize + codewords_matched.len())
+        == 1
+        && codewords_next_level_matched.is_empty()
+    {
+        RowMajorMatrix::new(
+            running_codeword_opt
+                .or_else(|| codewords_matched.first().map(|m| m.as_view()))
```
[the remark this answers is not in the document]
hero78119: At the first round running_codeword_opt is `None` so we can't just unwrap here. That's why here or_else deal with this case

mpcs/src/basefold/commit_phase.rs
hero78119: Good idea. Before widely refactor to define it in codeword, I define a util function and eliminate all of the duplicated logic in mpcs prover/verifier impl

mpcs/src/basefold/commit_phase.rs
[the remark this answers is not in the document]
kunxian-xia: yeah, you're right.

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +325,142 @@ where
...
+                .unwrap()
+                .values
+                .par_chunks_exact(2)
+                .zip(folding_coeffs)
+                .map(|(ys, coeff)| {
+                    let (left, right) = (ys[0], ys[1]);
+                    // original (left, right) = (lo + hi*x, lo - hi*x), lo, hi are codeword, but after times x it's not codeword
+                    // recover left & right codeword via (lo, hi) = ((left + right) / 2, (left - right) / 2x)
```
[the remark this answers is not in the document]
hero78119: fixed, also extract those duplicated logic into a util function

mpcs/src/basefold/commit_phase.rs
hero78119: Nice catch 👍  updated in commit https://github.com/scroll-tech/ceno/pull/894/commits/88ebfbbeb30ac122e72f4db9db59675f3e48e8a1 via aggregate first then fold

mpcs/src/basefold/structure.rs
hero78119: Yes, add new commit https://github.com/scroll-tech/ceno/pull/894/commits/1e403e80544a644cee56d9d73340c053dcd01cc1 to clean up old `sumcheck_messages` and only keep `sumcheck_proof `

mpcs/src/basefold/structure.rs
```diff
@@ -211,7 +233,7 @@ impl<E: ExtensionField, Spec: BasefoldSpec<E>> Clone for Basefold<E, Spec> {
     }
 }
 
-pub type MKProofNTo1<F1, P> = (Vec<F1>, P);
+pub type MKProofNTo1<F1, P> = (Vec<Vec<F1>>, P);
 // for 2 to 1, leaf layer just need one value, as the other can be interpolated from previous layer
 pub type MKProof2To1<F1, P> = (F1, P);
 pub type QueryOpeningProofs<E> = Vec<(
```
kunxian-xia, line 239: maintainability: shall we refactor this big type definition by assigning names to these inner fields
```rust
// borrowed from https://github.com/Plonky3/Plonky3/blob/main/fri/src/two_adic_pcs.rs#L51
pub struct BatchOpening<F: Field, InputMmcs: Mmcs<F>> {
    pub opened_values: Vec<Vec<F>>,
    pub opening_proof: InputMmcs::Proof,
}

// type definitions borrowed from https://github.com/Plonky3/Plonky3/blob/main/fri/src/proof.rs#L24
// M is ExtensionMmcs
pub struct CommitPhaseStep<E, M: Mmcs<E>> {
     pub sibling_value: E, 
     pub opening_proof: M::Proof,
}

// InputProof here could be `struct BatchOpening`
pub struct QueryOpeningProof<E, InputProof> {
   pub witin_input_proof: InputProof,
   pub fixed_input_proof: Option<InputProof>,
   pub commit_phase_openings: Vec<CommitPhaseStep>
}
```

hero78119: 👍 good idea and it make easier to speak in same language
Add new commit https://github.com/scroll-tech/ceno/pull/894/commits/da7d1422018acd5df131d646009b57edf9e6246a switch to p3 struct

mpcs/src/basefold/query_phase.rs
```diff
@@ -61,173 +64,408 @@ where
...
+                .collect_vec();
+            debug_assert_eq!(folding_sorted_order.len(), base_codeword_lo_hi.len());
+            debug_assert!(fixed_commit_leafs_iter.next().is_none());
+            debug_assert!(batch_coeffs_iter.next().is_none());
+
+            // fold and query
+            let mut cur_num_var = max_num_var;
+            let rounds = cur_num_var
+                - <Spec::EncodingScheme as EncodingScheme<E>>::get_basecode_msg_size_log()
+                - 1;
```
kunxian-xia, line 285: why do we need to minus 1?

mpcs/src/basefold/commit_phase.rs
```diff
@@ -201,54 +326,139 @@ where
...
+                })
+                .collect::<Vec<_>>(),
+            2,
+        )
+    };
+
+    if cfg!(feature = "sanity-check") && is_last_round {
+        let (commitment, merkle_tree) = mmcs_ext.commit_matrix(folded_codeword.clone());
+        commits.push(commitment);
+        trees.push(merkle_tree);
```
kunxian-xia, line 443: why did we skip the commit of last round's codeword?

kunxian-xia (approved):
Great work!

mpcs/src/basefold/commit_phase.rs
hero78119: add explanation in https://github.com/scroll-tech/ceno/pull/894/commits/3c20a048416b5965f92a4ff88a66ebeca79fbe3e

mpcs/src/basefold/query_phase.rs
hero78119: add explanation in https://github.com/scroll-tech/ceno/pull/894/commits/3c20a048416b5965f92a4ff88a66ebeca79fbe3e

Issue #861, opened 2025-03-17
Use plonky3 `maybe-rayon` crate for optional parallelism mode


hero78119:
Plonky3 [maybe-rayon](https://github.com/Plonky3/Plonky3/tree/main/maybe-rayon) provide a good wrapper to control parallism via "parallel" flag, thus within debug mode we can turn off the flag to avoid messy stack tracing info due to parallel `rayon`.

Do achieve that, we need to clean up all `rayon` import in current codebase and unifying to `use p3::maybe_rayon::prelude::*` instead

SamarthJ2004:
Hi, I would like to work on this issue.

Had a doubt: does it require to change even specific imports to ::*
eg: 
use rayon::prelude::{ParallelIterator, ParallelSlice, ParallelSliceMut};

to use p3::maybe_rayon::prelude::*

and are following imports also required to be changed?

use rayon::iter::{IntoParallelRefIterator, ParallelIterator};

hero78119:
> Hi, I would like to work on this issue.
> 
> Had a doubt: does it require to change even specific imports to ::* eg: use rayon::prelude::{ParallelIterator, ParallelSlice, ParallelSliceMut};
> 
> to use p3::maybe_rayon::prelude::*
> 
> and are following imports also required to be changed?
> 
> use rayon::iter::{IntoParallelRefIterator, ParallelIterator};

Hi, thanks for interesting in this issue and please go ahead if you are interested :)

> Had a doubt: does it require to change even specific imports to ::* eg: use rayon::prelude::{ParallelIterator, ParallelSlice, ParallelSliceMut};
> use rayon::iter::{IntoParallelRefIterator, ParallelIterator};

I think to simpler we just need to had wildcard import. Checked other similar project e.g. [openvm](https://github.com/search?q=repo%3Aopenvm-org%2Fopenvm%20%22p3_maybe_rayon%22&type=code) it seems to have same strategy. 

We also need to unify some serial/non-serial version e.g. https://github.com/scroll-tech/ceno/blob/master/witness/src/lib.rs#L131-L144 into same api and avoid concrete type so we can control all function in same place.

A side note: for your first PR, please probably on hold for the rayon change in within sumcheck crate https://github.com/scroll-tech/ceno/tree/master/sumcheck and goes for other crate first

hero78119:
> Hey [@hero78119](https://github.com/hero78119) I am new here and would like to contribute . Is this issue still open to be worked upon ?

[...]

PR #896, opened 2025-04-04
fix: conditional variable definition in `num_threads`

mdqst:
I fixed a problem in the `num_threads` function. The original code used `#[cfg(feature = "parallel")] let nt = rayon::current_num_threads();`, which caused an issue because the variable `nt` would only be defined if the `parallel` feature was active. This led to a compilation error when the feature was disabled, as the compiler couldn't find `nt`.  
Now, the function properly returns the correct value based on whether the `parallel` feature is enabled or not, avoiding the issue where the variable `nt` wasn't always available.

### DIFF
--- a/mpcs/src/util/parallel.rs
+++ b/mpcs/src/util/parallel.rs
@@ -1,7 +1,13 @@
 pub fn num_threads() -> usize {
     #[cfg(feature = "parallel")]
-    let nt = rayon::current_num_threads();
-    if cfg!(feature = "parallel") { nt } else { 1 }
+    {
+        return rayon::current_num_threads();
+    }
+
+    #[cfg(not(feature = "parallel"))]
+    {
+        return 1;
+    }
 }
 
 pub fn parallelize_iter<I, T, F>(iter: I, f: F)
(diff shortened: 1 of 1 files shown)


mpcs/src/util/parallel.rs
```diff
@@ -1,7 +1,13 @@
 pub fn num_threads() -> usize {
```
hero78119, line 1: Giving another similar TODO issue https://github.com/scroll-tech/ceno/issues/861
I would suggest make this PR with slightly wider scope: remove this `parallel.rs` file and leverage plonky3 `maybe-rayon` to achieve the same functionality

hero78119 (commented):
Hi, thanks for the PR!

mpcs/src/util/parallel.rs
hero78119: we will in favour of this solution instead https://github.com/scroll-tech/ceno/pull/897, for `parallel.rs` is some previous left-over and we will clean it up eventually

hero78119:
This issue was addressed, feel free to re-open if issue still there 😄

PR #897, opened 2025-04-04
mpcs parallel feature refactor

hero78119:
partially addressed #861 and an alternative for #896


mpcs/Cargo.toml
```diff
@@ -43,7 +43,7 @@ criterion.workspace = true
 [features]
 benchmark = ["parallel"]
 default = ["parallel"] # Add "sanity-check" to debug
-parallel = ["dep:rayon"]
```
hero78119, line 46: we will remove "dep:rayon" control and only keep "p3/parallel" after #861

PR #899, opened 2025-04-07
simplify design without transcript fork

hero78119:
Extract from PR #895 

Remove forked transcript to simplify design. 

### benchmark
To assure no impact for e2e
Fibonacci 2^20
```
                        time:   [2.8696 s 2.8901 s 2.9127 s]
                        change: [-1.0773% -0.1480% +0.8449%] (p = 0.79 > 0.05)
                        No change in performance detected.

```

Fibonacci 2^21
```
fibonacci_max_steps_2097152/prove_fibonacci/fibonacci_max_steps_2097152
                        time:   [5.2648 s 5.2801 s 5.2959 s]
                        change: [+0.4221% +0.9989% +1.5705%] (p = 0.00 < 0.05)
```

### DIFF
--- a/ceno_zkvm/src/scheme/prover.rs
+++ b/ceno_zkvm/src/scheme/prover.rs
@@ -19,7 +19,7 @@ use sumcheck::{
     structs::{IOPProverMessage, IOPProverState},
     util::optimal_sumcheck_threads,
 };
-use transcript::{ForkableTranscript, Transcript};
+use transcript::Transcript;
 use witness::{RowMajorMatrix, next_pow2_instance_padding};
 
 use crate::{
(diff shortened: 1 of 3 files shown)


lispc (approved):
the new codes seems good.  While i dont know the initial reason for forkable transript.

For parallel proving, or for better handling that some circuit may not exist (non proof)?

Now since we are using uniq name to index, then we can use a normal transcript?

hero78119:
> the new codes seems good. While i dont know the initial reason for forkable transript.
> 
> For parallel proving, or for better handling that some circuit may not exist (non proof)?
> 
> Now since we are using uniq name to index, then we can use a normal transcript?

I think it's for parallel proving purpose at that moment. Right now design is more simple since we plan to batched mpcs for all opcode in one machine, do it make sense one base vm opcode/table proof also be in one machine

PR #901, opened 2025-04-09
control `num_instances` in one place to improve soundness check

hero78119:
This features is preliminary support to batched all opcode/table in one mpcs proof https://github.com/scroll-tech/ceno/pull/894

Previously we got few issues
1. `num_instance` are specify in each opcode proof
2. `rw_hint_var` are specify in each table proof

while both are describe the same thing in different place, which make thing hard to control and make room for malicious prover to crack based on inconsistency.

In this PR, we unify num_instance in one central place, and do a strictly check of it's consistency.
This also addressed feature in #648

### DIFF
--- a/ceno_zkvm/src/scheme.rs
+++ b/ceno_zkvm/src/scheme.rs
@@ -30,9 +30,6 @@ mod tests;
     deserialize = "E::BaseField: DeserializeOwned"
 ))]
 pub struct ZKVMOpcodeProof<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> {
-    // TODO support >1 opcodes
-    pub num_instances: usize,
-
     // product constraints
     pub record_r_out_evals: Vec<E>,
     pub record_w_out_evals: Vec<E>,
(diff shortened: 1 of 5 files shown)


ceno_zkvm/src/scheme.rs
```diff
@@ -144,8 +138,10 @@ pub struct ZKVMProof<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> {
     pub raw_pi: Vec<Vec<E::BaseField>>,
     // the evaluation of raw_pi.
     pub pi_evals: Vec<E>,
-    opcode_proofs: BTreeMap<String, (usize, ZKVMOpcodeProof<E, PCS>)>,
```
hero78119, line 147: here we have 
- String: circuit name
- usize: circuit index
to carry information in proof. However a malicious prover can make both inconsistent.

ceno_zkvm/src/scheme.rs
```diff
@@ -208,10 +213,10 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E> + Serialize> fmt::Dis
         let mpcs_opcode_commitment = self
             .opcode_proofs
             .iter()
-            .map(|(circuit_name, (_, proof))| {
```
hero78119, line 211: These change make less readibility since now we lost circuit_name and instead only circuit index. But it's ok we fix it later, for now mpcs still be the major bottleneck of proof size

hero78119:
covered via #894

PR #908, opened 2025-04-17
Fix table proof failed if there is only single instance

### LINKED ISSUE #906: make table proof works when instance = 1
This issue is found when upgrade toolchain to newer nightly version, e.g. https://github.com/scroll-tech/ceno/blob/master/rust-toolchain.toml#L2 from `2025-01-06` -> `2025-03-25`, and failed when there is only one record for stack/heap result, reproduced via command

```
cargo run  --features sanity-check --package ceno_zkvm --bin e2e  -- --platform=ceno --hints=10 --public-io=4191 examples/target/riscv32im-ceno-zkvm-elf/release/examples/fibonacci
```

The root cause is previously for table proof they are all static table so size are pre-defined during key setup phase.
After features #889, stack/heap of zkVM can be non-uniform, so the number of instances in this case is not necessary power of 2
In particularly, they can be 1, and in this case it will cause assertion failed.

The same code works fine in old toolchain. So one of the major reason might due to in latest tool, assemply code of `fibonacci` was optimized due to riscv32 toolchain upgrade

hero78119:
To closed #906 

- [x] one line change in `ceno_zkvm/src/scheme/verifier.rs` to make minimal padding size as 2
- [x] upgrade to `nightly-2025-03-25`, some file re-format due to newer `rustfmt`.

In `nightly-2025-01-06` + fibonacci 2^21, max accessed `stack` got `17` records.
In `nightly-2025-03-25` +  fibonacci 2^21,  `stack` only got `1` records.

Giving Fibonacci source

https://github.com/scroll-tech/ceno/blob/aaeaaf7e11f5b2e791d3b7df163c0f802166b83c/examples/examples/fibonacci.rs#L1-L19

It make sense for much less stack access after toolchain upgrade, which I believe it's due to newer riscv compiler optimization


hero78119:
Soundness notes: the number of instances could be witnessed from prover freely. In stack/heap circuit all initial values are constrained to be 0,  and read_set/write_set are constrained via opcode circuits. The prover have no choice but only can supply honest heap/stack "final" values to make offline-memory-check protocol pass

kunxian-xia (approved):

PR #913, opened 2025-04-22
fix: fs::canonicalize return Err when nx

lightsing:
fix cli bug

### DIFF
--- a/ceno_cli/src/utils.rs
+++ b/ceno_cli/src/utils.rs
@@ -126,3 +126,35 @@ pub fn parse_size(s: &str) -> Result<u32, parse_size::Error> {
         .parse_size(s)
         .map(|size| size as u32)
 }
+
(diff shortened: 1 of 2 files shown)


ceno_cli/src/utils.rs
```diff
@@ -126,3 +126,35 @@ pub fn parse_size(s: &str) -> Result<u32, parse_size::Error> {
         .parse_size(s)
         .map(|size| size as u32)
 }
+
+/// Canonicalize a path allowing for non-existent paths.
+pub fn canonicalize_allow_nx<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
```
hero78119, line 131: nitpick: `try_canonicalize_full_path` might be `self-explained`

hero78119 (approved):

PR #921, opened 2025-04-24
fix: temporary skip verify

lightsing:


ceno_zkvm/src/bin/e2e.rs
```diff
@@ -282,9 +288,11 @@ fn run_inner<
     let vk_bytes = bincode::serialize(&vk).unwrap();
     fs::write(&vk_file, vk_bytes).unwrap();
 
-    let verifier = ZKVMVerifier::new(vk);
-    verify(&zkvm_proof, &verifier).expect("Verification failed");
-    soundness_test(zkvm_proof, &verifier);
+    if !skip_verify {
```
hero78119, line 291: How about we define a new e2e stage "PrepProof"? I think this design fit more natural in original e2e flow.

Take `run_elf_inner` as example
```rust
run_elf_inner::<E, PCS, P>(args, elf_path, Checkpoint::PrepSanityCheck, skip_verify)?;
```  

here 3rd and 4th got some subset relation, somehow indicated one of them is redundant

lightsing: oh! good point!

ceno_zkvm/src/e2e.rs
```diff
@@ -546,10 +588,13 @@ pub fn run_e2e_with_checkpoint<
     let exit_code = emul_result.exit_code;
 
     if let Checkpoint::PrepWitnessGen = checkpoint {
-        return (
-            (None, None),
-            Box::new(move || _ = generate_witness(&system_config, emul_result, &program, false)),
```
lightsing, line 551: why previously this didn't respect the `is_mock_proving ` flag? is that intend?

hero78119: yes it is on purpose, but I feel the naming is confused. 

**what's `is_mock_proving` do**
this is like halo2 mock_prover functionality, which is checking satisfying of al constrains on traces directly without generate proof.

In here, when we run e2e and halt **before** `generate_witness`, this imply we are going to benchmark generate_witness performance. So we skip mock proving check on `generate_witness`

But I feel we should have its control via "sanity_check" feature, wdyt? (we can do it in separate PR)

PR #928, opened 2025-04-28
Fix an example code in doc comments (cargo test tries to run it)

yczhangsjtu:
Cargo test fails on my laptop in master because it tries to run an example snippet in the documentation comment of the function `split_by_sizes`. Fixed that snippet.


hero78119 (approved):
May I know what thee command to trigger docs test?
probably we can add it in CI (+ cargo.make script)

PR #929, opened 2025-04-29
support guest program "println!" in e2e flow

hero78119:
To close issue #632
named io as `debug_println` in guest program debug build, assuming no "println!" use case in guest program.

### Design rationales
In debug build, we extend stack address a bit to cover a reserved 256k for io. This extra reserved space also reflect in linker script, so the write to this region wont got any complaints from either elf or riscv emulator

Besides, this PR also fix a previous problem where meaningful symbol in bss/sbss section will be skip due to their value are 0. We need to reserve and padding to cover them, since those might be some static variables initialized with 0 or uninitialized. Without do it, emulator will also complain regions is not writable.

### Change scope
- cleanup previois workaround in guest program for io
- extend stack address for io consistency check during debug build
- refactor `load_elf` bss/sbss padding issue.
- e2e command also shows io result.
- respect profile in guest program examples compilation.

### Example testing command:

An guest program with IO

```bash
# release run
cargo run --release --features sanity-check --package ceno_zkvm --bin e2e  -- --platform=ceno --hints=10 --public-io=4191 examples/target/riscv32im-ceno-zkvm-elf/release/examples/ceno_rt_io

# debug run
cargo run --features sanity-check --package ceno_zkvm --bin e2e  -- --platform=ceno --hints=10 --public-io=4191 examples/target/riscv32im-ceno-zkvm-elf/debug/examples/ceno_rt_io
```


lightsing (approved):
I see, so the stdout still not supported. native println! or eprintln! won't be collected.
We may need this later.

hero78119:
> I see, so the stdout still not supported. native println! or eprintln! won't be collected. We may need this later.

Yeah I intent to not support `println!` at this moment since I feel io should only for debug purpose. In release mode guest should never incurr io. Application constraint `assert` still work as expected

examples-builder/build.rs
```diff
@@ -23,21 +23,26 @@ fn build_elfs() {
     let _ = remove_file(&dest_path);
     let mut dest = File::create(&dest_path).expect("failed to create vars.rs");
 
-    // TODO(Matthias): skip building the elfs if we are in clippy or check mode.
-    // See git history for an attempt to do this.
+    let is_release = std::env::var("PROFILE").unwrap() == "release";
+    let mut args = vec!["build", "--examples", "--target-dir", "target"];
+    if is_release {
```
hero78119, line 28: This also resolved asked question in original issue https://github.com/scroll-tech/ceno/issues/632#issuecomment-2745315257

Issue #934, opened 2025-05-05
more enhancement of sumcheck protocol


hero78119:
To support features and enhancement
- #799 for precompile more expressive expression-based circuit style
- applied [Some Improvements for the PIOP for ZeroCheck](https://eprint.iacr.org/2024/108)
- skip round tricks [GKR for Boolean Circuits with Sub-linear RAM Operations](https://eprint.iacr.org/2025/717)

Will extend current sumcheck protocol with
- [ ] add [expression](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/expression.rs) in virtual poly syntax for sumcheck protocol.
       > internally we convert expression into monomial terms thus computation-intensive tasks in expression based protocol still remain inline
- [ ] applied zerocheck optimisation in sumcheck protocol
- [ ] add skip rounds mechanism

PR #935, opened 2025-05-05
refactor expression to right place.

hero78119:
Previously we put expression in zkvm crate for fast development. As times goes by, there are more place which require expression functionality, e.g. #934 

This PR 
- refactor expression into `multilinear_extensions` crate.
- cleanup unused expression in mpcs crate


spherel (approved):
LGTM!

PR #943, opened 2025-05-12
support debug mode in cargo cli run

hero78119:
Previously cargo ceno cli only run on release mode, while there is some "debug_assertion" optional compilation in e2e which couldn't be switch on/off. 

This PR 
1. refactor debug flag into `Platform` so both cli/e2e can control it properly
2. print io message even vm execution failed

### DIFF
--- a/ceno_cli/src/commands/common_args/cargo.rs
+++ b/ceno_cli/src/commands/common_args/cargo.rs
@@ -38,11 +38,11 @@ pub struct CargoOptions {
 }
 
 /// Package Selection:
-///   --package [<SPEC>]  Package with the target to run
+///   -p, --package [<SPEC>]  Package with the target to run
 #[derive(Clone, Args)]
 pub struct PackageSelection {
     /// Package with the target to run
-    #[arg(long)]
+    #[arg(short = 'p', long)]
     pub package: Option<String>,
 }
 
--- a/ceno_cli/src/commands/common_args/cargo.rs
+++ b/ceno_cli/src/commands/common_args/cargo.rs
@@ -38,11 +38,11 @@ pub struct CargoOptions {
 }
 
 /// Package Selection:
-///   -p, --package [<SPEC>]  Package with the target to run
+///   --package [<SPEC>]  Package with the target to run
 #[derive(Clone, Args)]
 pub struct PackageSelection {
     /// Package with the target to run
-    #[arg(short = 'p', long)]
+    #[arg(long)]
     pub package: Option<String>,
 }
 
--- a/ceno_cli/src/commands/run.rs
+++ b/ceno_cli/src/commands/run.rs
@@ -91,7 +91,7 @@ impl ProveCmd {
 }
 
 impl CmdInner {
-    fn run(self, toolchain: Option<String>, kind: RunKind) -> anyhow::Result<()> {
+    fn run(mut self, toolchain: Option<String>, kind: RunKind) -> anyhow::Result<()> {
         let manifest_path = match self.manifest_options.manifest_path.clone() {
             Some(path) => path,
             None => search_cargo_manifest_path(current_dir()?)?,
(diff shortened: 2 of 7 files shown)


ceno_cli/src/commands/common_args/cargo.rs
```diff
@@ -38,11 +38,11 @@ pub struct CargoOptions {
 }
 
 /// Package Selection:
-///   -p, --package [<SPEC>]  Package with the target to run
+///   --package [<SPEC>]  Package with the target to run
 #[derive(Clone, Args)]
 pub struct PackageSelection {
     /// Package with the target to run
-    #[arg(short = 'p', long)]
+    #[arg(long)]
```
hero78119, line 45: this "short" conflict with another "platform". 

https://github.com/scroll-tech/ceno/blob/75d75ea49c5a0c9a9f4383561eaa95cf003b7b19/ceno_cli/src/commands/common_args/ceno.rs#L27

lightsing: prefer to make platform flag only available at `--platform` to keep compatible with cargo cli.

ceno_cli/src/commands/run.rs
```diff
@@ -100,6 +100,9 @@ impl CmdInner {
             .target_selection
             .canonicalize(manifest_path, &self.package_selection)?;
 
+        // XXX: custom handling: set release mode from compilation_options
+        self.ceno_options.release = self.compilation_options.release;
```
lightsing, line 104: passing compilation_options where it needs instead add a new field `release` in ceno_options

ceno_zkvm/src/e2e.rs
```diff
@@ -289,14 +291,38 @@ pub fn setup_platform(
     stack_size: u32,
     heap_size: u32,
     pub_io_size: u32,
+) -> Platform {
+    setup_platform_inner(preset, program, stack_size, heap_size, pub_io_size, false)
+}
+
+pub fn setup_platform_debug(
```
lightsing, line 298: why adding a new function instead of add a new arg

lightsing (changes requested):

ceno_cli/src/commands/run.rs
hero78119: Do you mean add `compilation_options` as an argument to `compilation_options.run()`

https://github.com/scroll-tech/ceno/blob/75d75ea49c5a0c9a9f4383561eaa95cf003b7b19/ceno_cli/src/commands/run.rs#L117-L125

I feel it break the symmetric of other api. For now ceno_options.release mark as `#[clap(skip)]` so I think it's self-explained

ceno_zkvm/src/e2e.rs
hero78119: Usually we prefer function to be self-explain which avoid "Boolean Parameter Smell" and possess "Self-documenting Code"

For example.

```
setup_platform_inner(..., true)   // what does `true` mean?
setup_platform_inner(..., false)  // is this debug mode or something else
```

ceno_cli/src/commands/common_args/cargo.rs
hero78119: could you elaborate more on the comment?
originally both "--package", "--platform" conflict on short "-p" 
I think we should let "-p" works on platform, while leave "--package" will full name

lightsing: I prefer to let `-p` used by `--package`, which makes the behavior same as `cargo` itself.

ceno_cli/src/commands/run.rs
lightsing: but this make me feels like "multiple source of the knowledge"

ceno_zkvm/src/e2e.rs
lightsing: ok, that make sense, I might do some refactor later

PR #944, opened 2025-05-12
fix: find the correct target when using cargo workspace

lightsing:
found bug when ceno cli is used in a cargo workspace, this pr fixes how ceno cli finds workspace root and correct manifest/target to run with.


hero78119 (approved):
LGTM (although I didnt fully verified locally and assume it has been well tested)

PR #945, opened 2025-05-13
feat: separate e2e artifacts

lightsing:
- Adds a context struct `E2EProgramCtx` that remains the same for a given program and platform. 
- Makes several functions and struct fields public.
- `run_e2e_with_checkpoint` do not need `hints` and `public_io` as Vec.


hero78119 (approved):
LGTM!

PR #946, opened 2025-05-13
add jemallocator as optional global allocator

hero78119:
benchmark shows there are quite of time spending on glibc free (drop) when object end of its scopes.

Follow openvm using [jemalloc](https://github.com/openvm-org/openvm/blob/c771a213f5e7f0732e0ddbafb273e15d99c5049d/crates/vm/Cargo.toml#L56) as global allocators.
and set jemalloc parameter follows https://github.com/openvm-org/openvm/blob/c771a213f5e7f0732e0ddbafb273e15d99c5049d/.github/workflows/benchmark-call.yml#L218
> I do not use jemalloc "background_thread: true" as I thought thread in background might occupied other schedule which affect cpu intensive program

### change scope
- enable jemalloc by default when compiling ceno_cli
- support `cargo make cli` to install ceno_cli
- introduce "jemalloc" features

### benchmark

benchmark on AMD EPYC 32 cores with command `JEMALLOC_SYS_WITH_MALLOC_CONF="retain:true,metadata_thp:always,thp:always,dirty_decay_ms:-1,muzzy_decay_ms:-1,abort_conf:true" cargo bench --bench fibonacci --features jemalloc --package ceno_zkvm -- --baseline opt-baseline`
 

| Benchmark       | Average Time | Improvement | Throughput (instructions/sec) |
|-----------------|--------------|-------------|---------------------------|
| fibonacci 2^20  | 2.0020 s     | -14.74%     | 523.76k                    |
| fibonacci 2^21  | 3.5903 s     | -18.89%     | 584.34k                    |
| fibonacci 2^22  | 6.6531 s     | -24.69%     | 630.28k                    |


lightsing:
OpenVM enables `jemallocator` as a feature even when used as a library, which is quite annoying.

In my opinion, the choice of global allocator should be entirely left to the user. Specifically, `jemallocator` being silently enabled as a default feature of OpenVM is something we should avoid.

I suggest making `jemallocator` an optional feature (possibly enabled by default) for the `cargo-ceno/benchmark` binary only, and not affecting the library part at all.

lispc:
i agree with "let top level app manage allocator" is good.  But i think this PR can be merged.

Now the e2e.rs in zkvm crate needs jemalloc.   If we want to avoid jemalloc in zkvm the lib crate, we need to refactor to move the e2e.rs elsewhere.   It would bring lot of code changes.  Current code anyway is useful.   When it is used as a lib, "if I rely on a library's bundled jemallocator, I lose the ability to manage its version explicitly.",  but anyway it is doable.

We can consider move e2e.rs elsewhere in future, then reconsider whether to remove the jemalloc feature.

lispc (approved):

PR #950, opened 2025-05-14
support u64 cycle

### LINKED ISSUE #949: end_cycle could exceeds u32::MAX
Current we use `PublicValues<u32>` holds those values, but end_cycle could easily exceeds `u32::MAX`.
https://github.com/scroll-tech/ceno/blob/9dca294e89ad463ad246f3e7cc561feba743eba2/ceno_zkvm/src/scheme.rs#L73-L117

While the `Cycle` type is an alias to `u64`, e2e would panic when cast it to fit in `u32`.
https://github.com/scroll-tech/ceno/blob/9dca294e89ad463ad246f3e7cc561feba743eba2/ceno_emul/src/addr.rs#L31
https://github.com/scroll-tech/ceno/blob/9dca294e89ad463ad246f3e7cc561feba743eba2/ceno_zkvm/src/e2e.rs#L147

hero78119:
To closed #949

### DIFF
--- a/ceno_zkvm/src/scheme.rs
+++ b/ceno_zkvm/src/scheme.rs
@@ -71,22 +71,22 @@ pub struct ZKVMTableProof<E: ExtensionField> {
 
 /// each field will be interpret to (constant) polynomial
 #[derive(Default, Clone, Debug)]
-pub struct PublicValues<T: Default + Clone + Debug> {
-    exit_code: T,
-    init_pc: T,
-    init_cycle: T,
-    end_pc: T,
-    end_cycle: T,
-    public_io: Vec<T>,
+pub struct PublicValues {
+    exit_code: u32,
+    init_pc: u32,
+    init_cycle: u64,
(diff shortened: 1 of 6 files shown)


ceno_zkvm/src/scheme.rs
```diff
@@ -71,22 +71,22 @@ pub struct ZKVMTableProof<E: ExtensionField> {
 
 /// each field will be interpret to (constant) polynomial
 #[derive(Default, Clone, Debug)]
-pub struct PublicValues<T: Default + Clone + Debug> {
+pub struct PublicValues<T: Default + Clone + Debug, V: Default + Clone + Debug> {
```
lightsing: why we need generic here?

lightsing (approved):

hero78119:
Also fix toolchain version mismatched from previous upgrade https://github.com/scroll-tech/ceno/blob/master/rust-toolchain.toml#L2

PR #952, opened 2025-05-19
WIP refactor #799

hero78119:
### Change scope
- [x] xunify `Expression` with ceno
- [x] unify sumcheck with ceno
- [ ] WIP GKR witness generation, take bit benchmark as example

PR #956, opened 2025-05-27
optimize extrapolation with zero field inverse during runtime

hero78119:
Extracted from #952.

Observe a bottleneck on previous interpolation which contribute to most of time due to `vector.extend` operation and bunch of allocations.
This PR rewrite univariate extrapolation
1. as the point to be interpolate are fixed set, we can pre-compute all stuff require field inverse
2. in-place change to avoid allocation 

### benchmark
In Ceno opcode main sumcheck part we batch different degree > 1 into one batch so this function will be used.
It shows a slightly improvement (~3%) on Fibonacci 2^24 e2e

| Benchmark                        | Median Time (s) | Median Change (%) |
|----------------------------------|------------------|--------------------|
| fibonacci_max_steps_1048576      | 2.3978           | +0.9805%   (No significant change    )        |
| fibonacci_max_steps_2097152      | 4.2579           | +1.7587%   (Change within noise)        |
| fibonacci_max_steps_4194304      | 7.7561           | -3.5338%           |

### DIFF
--- a/sumcheck/src/structs.rs
+++ b/sumcheck/src/structs.rs
@@ -1,5 +1,7 @@
 use ff_ext::ExtensionField;
-use multilinear_extensions::{virtual_poly::VirtualPolynomial, virtual_polys::PolyMeta};
+use multilinear_extensions::{
+    mle::Point, virtual_poly::VirtualPolynomial, virtual_polys::PolyMeta,
+};
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use transcript::Challenge;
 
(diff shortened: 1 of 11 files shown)


sumcheck/src/structs.rs
```diff
@@ -12,7 +14,7 @@ use transcript::Challenge;
     deserialize = "E::BaseField: DeserializeOwned"
 ))]
 pub struct IOPProof<E: ExtensionField> {
-    pub point: Vec<E>,
+    pub point: Point<E>,
```
spherel, line 17: I just found `Point<E>` shouldn't be included in `IOPProof<E>`. What do you think about it?

hero78119: yes I agree we should remove it and I will submit another PR to do that since in involved more change outside of this scope

spherel (approved):
LGTM!

PR #957, opened 2025-05-28
loop unroll univariate extrapolate

hero78119:
refactor univariate interpolation in barycentric and unroll version.

cross refer to issue https://github.com/scroll-tech/ceno-recursion-verifier/issues/6

### DIFF
--- a/sumcheck/src/util.rs
+++ b/sumcheck/src/util.rs
@@ -140,15 +140,24 @@ fn inner_extrapolate<F: Field, const IS_PARALLEL: bool>(
 ///   evaluation at degree 1:
 ///   - Two inverses for the distances `d0 = x - x0`, `d1 = x - x1`
 ///   - One inverse for the final normalization term
+///
+/// barycentric weights `w` are for polynomial interpolation.
+/// for a fixed set of interpolation points {x_0, x_1, ..., x_n}, the barycentric weight w_j is defined as:
+/// w_j = 1 / ∏_{k ≠ j} (x_j - x_k)
+/// these weights are used in the barycentric form of Lagrange interpolation, which allows
+/// for efficient evaluation of the interpolating polynomial at any other point
+/// The weights depend only on the interpolation nodes and can be precomputed once
 fn extrapolate_uni_poly_deg_1<F: Field>(p_i: &[F; 2], eval_at: F) -> F {
     let x0 = F::ZERO;
     let x1 = F::ONE;
 
-    let w0 = F::ONE;
-    let w1 = -F::ONE;
+    // w0 = 1 / (0−1) = -1
+    // w1 = 1 / (1−0) =  1
+    let w0 = -F::ONE;
+    let w1 = F::ONE;
 
-    let d0 = eval_at - x0; // = eval_at
-    let d1 = eval_at - x1; // = eval_at - 1
+    let d0 = eval_at - x0;
+    let d1 = eval_at - x1;
 
     let inv_d0 = d0.inverse();
     let inv_d1 = d1.inverse();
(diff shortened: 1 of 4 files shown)


sumcheck/src/util.rs
```diff
@@ -140,73 +136,154 @@ fn inner_extrapolate<F: Field, const IS_PARALLEL: bool>(
...
+    let x1 = F::ONE;
+
+    let w0 = F::ONE;
+    let w1 = -F::ONE;
+
+    let d0 = eval_at - x0; // = eval_at
+    let d1 = eval_at - x1; // = eval_at - 1
+
+    let inv_d0 = d0.inverse();
+    let inv_d1 = d1.inverse();
```
kunxian-xia, line 154: what `d0` or `d1` is zero?

hero78119: 1. the probability of d0 or d1 to be zero is only 2/|E| which is negligible
2. it will panic on rust version verifier https://github.com/scroll-tech/Plonky3/blob/6db45dd9da0e4ef496f98c96f780a4eaf0bccd0f/field/src/field.rs#L483 but we should deal with error properly in dsl

sumcheck/src/util.rs
```diff
@@ -140,73 +136,154 @@ fn inner_extrapolate<F: Field, const IS_PARALLEL: bool>(
...
+fn extrapolate_uni_poly_deg_3<F: Field>(p_i: &[F; 4], eval_at: F) -> F {
+    let x0 = F::from_u64(0);
+    let x1 = F::from_u64(1);
+    let x2 = F::from_u64(2);
+    let x3 = F::from_u64(3);
+
+    let w0 = -F::from_u64(1).div(F::from_u64(6)); // -1/6
+    let w1 = F::from_u64(1).div(F::from_u64(2)); // 1/2
+    let w2 = -F::from_u64(1).div(F::from_u64(2)); // -1/2
+    let w3 = F::from_u64(1).div(F::from_u64(6)); // 1/6
```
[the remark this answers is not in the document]
hero78119: sure, added more comment https://github.com/scroll-tech/ceno/pull/957/commits/724f2069fbddf06d08efdf9ed6ad3ff11e48ad3d

In recursive verifier we should evaluate these weight and hardcode as constant

sumcheck/src/util.rs
kunxian-xia: Since `eval_at` is the outcome of poseidon2 permutation, it's possible that hacker can take advantage of this to get the verifier (either in rust or circuit) panic. This will hurt our liveness guarantee.

sumcheck/src/util.rs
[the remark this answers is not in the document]
kunxian-xia: >  fixed a bug where in degree 1

sumcheck/src/util.rs
hero78119: > Since `eval_at` is the outcome of poseidon2 permutation, it's possible that hacker can take advantage of this to get the verifier (either in rust or circuit) panic. This will hurt our liveness guarantee.

Do you mean $0 = \text{Poseidon}(\text{pre-image})$ ? this account for preimage-resistant cryptographic permutation and Poseidon should got the same property which make this operation infeasible

sumcheck/src/util.rs
```diff
@@ -140,73 +136,175 @@ fn inner_extrapolate<F: Field, const IS_PARALLEL: bool>(
...
+    let t1 = w1 * p_i[1] * inv_d1;
+    let t2 = w2 * p_i[2] * inv_d2;
+    let t3 = w3 * p_i[3] * inv_d3;
+
+    let b0 = w0 * inv_d0;
+    let b1 = w1 * inv_d1;
+    let b2 = w2 * inv_d2;
+    let b3 = w3 * inv_d3;
+
+    (t0 + t1 + t2 + t3) * (b0 + b1 + b2 + b3).inverse()
```
kunxian-xia, line 240: According to barycentric formula, `p(r) = \sum_i p[i] * w[i] * prod / eval_at - i)` where `prod = ∏ (eval_at - i)`;

Since `t[i] = w[i] * p[i] * 1 / (eval_at - i)`, then does it mean `1 / (b0 + b1 + b2 + b3) = prod`???

kunxian-xia: If we take constant poly `f(x) = 1`, then `1 = f(eval_at) = \sum_i w[i] * 1 * prod / (eval_at - i)`. 
If we take `prod` out from the right hand in above equation, we get `prod = 1 / (\sum_i w[i] * b[i])`.

kunxian-xia: @hero78119 Can you add the above explanation to the comments?

kunxian-xia (approved):

PR #964, opened 2025-06-05
generalize main sumcheck as one gkr layer

hero78119:
This PR build on top of #799  with one extra https://github.com/scroll-tech/ceno/pull/964/commits/48ded1a7d6ab40f20336dd03bf65779a78007e51 to introduce backend expression and cached in constrain system. This align the design with pre-compile so its easier for next step refactor to introduce precompile chip in main flow.
Main sumcheck read/write lookup expression was simplified, as post `evaluate()` was also removed. 

### Expression

Expression will be simplified into 2 kind: frontend and backend expression
- frontend expression: expression with Witin/StructuralWitin/Fixed, in recursive/nested style
- backend expression: expression with Witin only, in monomial style.

After circuit setup, both expression content are all known and freezed. During runtime, we can take backend expression and evaluate its scalar with "challenge/instance" then the final expression can be put into sumcheck.

### benchmark

The nice thing is before/after change, there is no performance difference.

| Benchmark                        | Median Time (s) | Median Change (%)                     |
|----------------------------------|------------------|----------------------------------------|
| fibonacci_max_steps_1048576      | 2.0641           | -0.9869% (No change in performance detected) |
| fibonacci_max_steps_2097152      | 3.5514           | -1.0748% (Change within noise threshold)     |
| fibonacci_max_steps_1048576      | 2.0641           | -0.9869% (No change in performance detected) |

### DIFF
--- a/ceno_zkvm/src/chip_handler/general.rs
+++ b/ceno_zkvm/src/chip_handler/general.rs
@@ -469,4 +469,8 @@ impl<'a, E: ExtensionField> CircuitBuilder<'a, E> {
 
         Ok((is_eq, diff_inverse))
     }
+
+    pub(crate) fn finalize(&mut self) {
+        self.cs.finalize_backend_monomial_expression();
+    }
 }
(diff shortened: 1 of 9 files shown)


dreamATD:
The newest benchmark compared with the master branch:
| Benchmark                   | Median Change (%) |
|-----------------------------|-------------------|
| fibonacci_max_steps_1048576 | -2.9942%           |
| fibonacci_max_steps_2097152 | +1.7704%          |
| fibonacci_max_steps_4194304 | +1.9446%          |

PR #965, opened 2025-06-17
Add rotation verifier

dreamATD:
Here is a summary of the updates in this PR:
1. Correct the previous typo in the rotation prover, and add a rotation verifier.
2. Change the selector formular:
	0(rt) = \sum_{b \in \{0, 1\}^n} sel(rt, b) * (\sum_i alpha^i * rotated[i](b) - target[i](b)))
   where sel(rt, b) = eq(rt[5..], b[5..]) * sel'(rt[..5], b[..5])
   where sel'(rt', b') = b' in rotated[0..24] ? eq(rt', b') : 0.
3. Refine the default method `gkr_witness`. Previously, it assumes the `phase1_witnesses` are connected to the input layer in the same order. In a more general case, the `phase1` witnesses can be connected to the wires in any layer, which is set by `allocate_opening` method. I modify `gkr_witness` in consider of this case.

| Benchmark                  | My PC Running Time | Median Change (%) |
|----------------------------|--------------------|-------------------|
| prove_keccak_lookup_f_4096 | 4.7770 s           | -10.773%          |
| prove_keccak_lookup_f_8192 | 11.178 s           | -1.9963%          |

### DIFF
--- a/gkr_iop/src/gkr/booleanhypercube.rs
+++ b/gkr_iop/src/gkr/booleanhypercube.rs
@@ -83,9 +83,9 @@ impl BooleanHypercube {
         point: &Point<E>,
     ) -> E {
         match self.num_vars {
-            // rotated_eval = (1-r4) * left_eval - r4 * right_eval
-            // right_eval = ((1-r4) * left_eval - rotated_eval) / r4
-            5 => ((E::ONE - point[4]) * left_eval - rotated_eval) / point[4],
+            // rotated_eval = (1-r4) * left_eval + r4 * right_eval
+            // right_eval = (rotated_eval - (1-r4) * left_eval) / r4
+            5 => (rotated_eval - (E::ONE - point[4]) * left_eval) / point[4],
             num_vars => unimplemented!("not support {num_vars}"),
         }
     }
--- a/gkr_iop/src/gkr/layer.rs
+++ b/gkr_iop/src/gkr/layer.rs
@@ -7,7 +7,7 @@ use multilinear_extensions::{
     mle::{ArcMultilinearExtension, Point, PointAndEval},
 };
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
-use sumcheck_layer::SumcheckLayerProof;
+use sumcheck_layer::LayerProof;
 use transcript::Transcript;
 use zerocheck_layer::ZerocheckLayer;
 
@@ -74,7 +74,7 @@ pub struct Layer<E: ExtensionField> {
 
 #[derive(Clone, Debug, Default)]
 pub struct LayerWitness<'a, E: ExtensionField> {
-    pub bases: Vec<ArcMultilinearExtension<'a, E>>,
+    pub wits: Vec<ArcMultilinearExtension<'a, E>>,
     pub num_vars: usize,
 }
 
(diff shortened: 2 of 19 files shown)


gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -1185,7 +1164,7 @@ mod tests {
             states.push(std::array::from_fn(|_| rng.gen()));
         }
         // TODO enable check
-        let _ = run_faster_keccakf(setup_gkr_circuit::<E>(), states, false, false);
```
hero78119, line 1188: 👍 really nice to see functionality check pass

gkr_iop/src/gkr/layer.rs
```diff
@@ -100,7 +100,7 @@ impl<E: ExtensionField> Layer<E> {
             expr_names.len() == exprs.len(),
             "there are expr without name"
         );
-        let max_expr_degree = exprs.iter().map(|expr| expr.degree()).max().unwrap();
```
hero78119, line 103: Does the purpose of this change is for layer with empty expressions?

gkr_iop/src/gkr/layer.rs
```diff
@@ -261,6 +261,9 @@ impl<'a, E: ExtensionField> LayerWitness<'a, E> {
         assert!(!bases.is_empty() || !bases.is_empty());
         let num_vars = log2(bases[0].evaluations().len()) as usize;
         assert!(bases.iter().all(|b| b.evaluations().len() == 1 << num_vars));
-        Self { bases, num_vars }
+        Self {
+            wits: bases,
```
hero78119: nitpick: here is one missing `bases`-> `wits`

gkr_iop/src/gkr/booleanhypercube.rs
```diff
@@ -83,9 +83,9 @@ impl BooleanHypercube {
         point: &Point<E>,
     ) -> E {
         match self.num_vars {
-            // rotated_eval = (1-r4) * left_eval - r4 * right_eval
```
hero78119, line 86: nice catch (and thanks for update [hackmd](https://hackmd.io/HAAj1JTQQiKfu0SIwOJDRw?view=#))

hero78119 (commented):
Major functionality looks good to me! This bring a step milestone for next e2e integration + PCS opening

A question: do you have more insights for which key change bring "prove_keccak_lookup_f" performance boost 🔥 ?

gkr_iop/src/gkr/booleanhypercube.rs
dreamATD: My mistake. I wrote a typo previously...

gkr_iop/src/gkr/layer.rs
dreamATD: Yes, although it seems unlikely. I changed this when I was debugging and commented out all expressions :P

gkr_iop/src/gkr/layer.rs
dreamATD: Thanks for pointing it out! I have fixed it in https://github.com/scroll-tech/ceno/pull/965/commits/de3806ea63ebdeb1f7e626d596f5b02f720a1851.

hero78119 (approved):
:+1:

dreamATD:
> Major functionality looks good to me! This bring a step milestone for next e2e integration + PCS opening
> 
> A question: do you have more insights for which key change bring "prove_keccak_lookup_f" performance boost 🔥 ?

I think it's mainly because I change the fomular from
0(rt) = \sum_{b \in \{0, 1\}^n} eq(rt, b) sel(rt, b) * (\sum_i alpha^i * rotated[i](b) - target[i](b)))
to
0(rt) = \sum_{b \in \{0, 1\}^n} sel(rt, b) * (\sum_i alpha^i * rotated[i](b) - target[i](b)))
which is one degree smaller than before.

I also bench the witness generation time locally, the performance seems regressed.

PR #971, opened 2025-06-30
[precompile] aggregate lookup expression within keccak precompile

hero78119:
aggregate lookup expression in keccak precompile instead of pass flatten out_evals of raw lookup value.
This make later gkr-iop flow integration easier when integrate into ceno main proving flow

### benchmark

with command on `research` bench server
```
JEMALLOC_SYS_WITH_MALLOC_CONF=retain:true,metadata_thp:always,thp:always,dirty_decay_ms:-1,muzzy_decay_ms:-1,abort_conf:true cargo bench -p gkr_iop --features jemalloc --bench lookup_keccakf -- --baseline baseline
```
There shows prover performance keccak precompile regressed which is expected. Because the cost to aggregate them always there, just we do it within chip or outside chip.

| Benchmark                         | Median Time (s) | Median Change (%)                     |
|----------------------------------|------------------|----------------------------------------|
| keccak_lookup_f_4096             | 0.98618          | +29.015% (Performance has regressed)   |
| keccak_lookup_f_8192             | 1.9001           | +23.847% (Performance has regressed)   |


gkr_iop/src/gkr/mock.rs
```diff
@@ -63,7 +63,7 @@ impl<E: ExtensionField> MockProver<E> {
         let field_type_default = FieldType::default();
         let mut evaluations = evaluations; // Change evaluations' lifetime.
         evaluations.resize_with(circuit.n_evaluations, || &field_type_default);
-        challenges.resize_with(circuit.n_challenges, || E::random(&mut rng));
+        challenges.resize_with(2 + circuit.n_challenges, || E::random(&mut rng));
```
spherel, line 66: Why is it `+2`?

hero78119: index 0, 1 is for alpha, beta global challenge. Here `circuit.n_challenges` only counting local challenges, which is 0.
we can make this converation open without Resolve 👍 , so it can be refer from future task

spherel (approved):
LGTM!

PR #973, opened 2025-07-01
mem read/write rlc within precompile circuit

hero78119:
This PR add the missing witness: current ts, mem read ts, mem start addr in keccak precompile.
Those information are needed for consistent check.
Beside, mem read/write are rlc with those record

> This PR is not fully sound, as we need to revamp current eq with selector logic


ceno_zkvm/src/instructions/riscv/dummy/dummy_ecall.rs
```diff
@@ -208,8 +210,14 @@ impl<E: ExtensionField> GKRIOPInstruction<E> for LargeEcallDummy<E, KeccakSpec>
                     .unwrap()
             })
             .collect_vec();
+        let num_instances = instances.len();
 
-        layout.phase1_witness_group(KeccakTrace { instances })
+        layout.phase1_witness_group(KeccakTrace {
+            instances,
+            ram_start_addr: ceno_emul::ByteAddr::from(0),
```
hero78119, line 217: value here is just mock. Will use correct value during integration

gkr_iop/src/gkr/layer_constraint_system.rs
```diff
@@ -270,8 +350,27 @@ impl<E: ExtensionField> LayerConstraintSystem<E> {
...
+        assert_eq!(ram_write_evals.len(), self.ram_write.len(),);
+        assert_eq!(ram_read_evals.len(), self.ram_read.len(),);
+
+        for (idx, ram_expr, ram_eval) in izip!(
+            0..,
+            chain!(self.ram_write.clone(), self.ram_read.clone(),),
+            ram_write_evals.chain(ram_read_evals)
+        ) {
+            self.add_non_zero_constraint(
+                ram_expr - E::BaseField::ONE.expr(), // ONE is for padding
```
hero78119, line 367: original formula 

$$ alpha  \times out(x) = \sum_s alpha \times (sel(s) \times input(s) + (1-sel(s)) \times padding) $$


rearrange to 

$$ alpha  \times (out(x) - padding) = \sum_s alpha \times sel(s) \times (input(s) - padding) $$

This follow current ceno main prover logic

rhs evaluation will be evaluated from verifier. Thus LHS expression just need expr - padding

gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -129,6 +129,9 @@ pub struct KeccakWitCols<T> {
     pub nonlinear: [T; 200],
     pub chi_output: [T; 8],
     pub iota_output: [T; 200],
+    pub input_ram_start_addr: [T; 1],
+    pub cur_ts: [T; 1],
+    pub input_ram_read_ts: [T; KECCAK_INPUT32_SIZE],
```
hero78119, line 134: 2 extra field `cur_ts`, `input_ram_read_ts` is for mem consistency check for offline-memory-check protocol

gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -137,6 +140,8 @@ pub struct KeccakLayer<WitT, FixedT, EqT> {
     pub wits: KeccakWitCols<WitT>,
     pub fixed: KeccakFixedCols<FixedT>,
     pub eq_zero: EqT,
+    pub eq_mem_read: EqT,
+    pub eq_mem_write: EqT,
```
hero78119, line 144: ultimately, we need to tune eq_zero, eq_mem_read, eq_mem_write into right version

- eq_zero => deal with non-pow2 instance
- eq_mem_read => as a selector, only enable on first rotation index per instance, also deal with non-pow2 instance
- eq_mem_write => as a selector, only enable on last rotation index per instance, also deal with non-pow2 instance

ceno_zkvm/src/instructions/riscv/dummy/dummy_ecall.rs
```diff
@@ -208,8 +210,14 @@ impl<E: ExtensionField> GKRIOPInstruction<E> for LargeEcallDummy<E, KeccakSpec>
                     .unwrap()
             })
             .collect_vec();
+        let num_instances = instances.len();
 
-        layout.phase1_witness_group(KeccakTrace { instances })
+        layout.phase1_witness_group(KeccakTrace {
```
hero78119, line 215: value here is just mock. Will use correct value during integration in another task

ceno_zkvm/src/instructions/riscv/dummy/dummy_ecall.rs
```diff
@@ -208,8 +210,14 @@ impl<E: ExtensionField> GKRIOPInstruction<E> for LargeEcallDummy<E, KeccakSpec>
...
+            instances,
+            ram_start_addr: vec![ByteAddr::from(0); num_instances],
+            cur_ts: vec![0; num_instances],
+            read_ts: vec![[0; KECCAK_INPUT32_SIZE]; num_instances],
```
spherel, line 219: If we write precompile events at the beggining (cur_ts = 0), does it mean we don't need read_ts?

hero78119: Here is just mock data, and we will fill correct data in next PR.

- other opcode load keccak state data into mem in different global ts, it equivalently append `W` set with different ts
- so in keccak opcode, we need prover to witness read_ts to `R` set to cancel out respective entry in `W` set
- then after state mutated, we write cur_ts to `W` set for all state entry in mem

So we need cur_ts and read_ts here. Actually we need more field for register 1 (ecall id) read, register 2 read (mem start address). Those value i will added in future incremental PR.  And consider wrap them into nested `KeccakStateCols` to have better witness management

PR #975, opened 2025-07-02
[precompile] Add `fixed_witness_group()` to officially generate fixed witnesses.

dreamATD:


gkr_iop/src/precompiles/bitwise_keccakf.rs
```diff
@@ -795,6 +795,11 @@ where
     E: ExtensionField,
 {
     type Trace = KeccakTrace<E>;
+
+    fn fixed_witness_group(&self) -> Vec<Vec<E::BaseField>> {
```
hero78119: fixed witness return type should be aligned with https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/tables/mod.rs#L31-L35 which is `RowMajorMatrix<E::BaseField>` to be commit into merkle tree

gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -558,6 +558,20 @@ where
 {
     type Trace = KeccakTrace;
 
+    fn fixed_witness_group(&self) -> RowMajorMatrix<E::BaseField> {
```
hero78119, line 574: 👍  

Not this line, but per our offline discussion, to complete functionality, we need expand (i.e. `iter().cycle()` in rust) fixed poly dynamically via num_instances. I will come back to this part when I make the e2e integration

dreamATD: Do you mean the returned matrix should be the padded version?

hero78119: return matrix should be original size for commit into mpcs.
the place of padding should be `gkr_witness`, or another define another stage when know num_instance at that moment.

hero78119 (approved):
LGTM!

PR #976, opened 2025-07-02
[precompile] part1 integrate keccak precompile into e2e flow

hero78119:
This raised PR is not fully sound yet, however it serve as a step milestone for easier to review

### change scope
- define new `ComposedConstrainSystem` to wrap `original constrain system + gkr-iop circuit`. I decide for our first version, tower argument will remain in original constrain system because its dynamism layers not fit well into gkr-iop design. So I tend to keep both constrain system separately, then invoke css in proving flow accordingly. 
- define `Keccak` as a new (ecall-)opcode followed our conventions, and clean up previous workaround as `LargeDummy<KccakSpec>` which is not used for this purpose
- refactor lookup tables utilization from zkvm crate to gkr-iop so both crate could be shared.

### TODOs in followup PR(s)
- build correct circuit of `keccak` ecall opcode to deal with state(memory) consistency => existing ceno zkvm circuit builder has few good utilization to wrap boilerplate code. But gkr-iop define it's own constrain system so it's hard to leverage existing utilization. 
- assign witness mechanism
- integrate into main prover flow + mpcs

### benchmarks

On fibonacci e2e against master branch

| Benchmark                        | Median Time (s) | Median Change (%)                           |
|----------------------------------|------------------|----------------------------------------------|
| fibonacci_max_steps_1048576      | 2.0902           | -1.1437% (No change in performance detected) |
| fibonacci_max_steps_2097152      | 3.5378           | -0.5016% (No change in performance detected) |
| fibonacci_max_steps_4194304      | 6.8585           | -1.9079% (No change in performance detected) |

and lookup keccak

| Benchmark                        | Median Time (s) | Median Change (%)                           |
|----------------------------------|------------------|----------------------------------------------|
| keccak_lookup_f_4096             | 0.88786          | +1.8583% (No change in performance detected) |
| keccak_lookup_f_8192             | 1.7728           | +0.9121% (No change in performance detected) |

both benchmark remain unchanged

PR #978, opened 2025-07-04
circuit builder and constrain system share between ceno_zkvm + gkr_iop

hero78119:
PR build on top of #973 #976

This PR refactor `ConstrainSystem`, `CircuitBuilder` to `gkr-iop`, thus gkr-iop can leverage few existing utilization implemented in `ConstrainSystem`, e.g. register_read/write, assert_lt etc.

`CircuitBuilder` some util dedicate to zkvm logic are stayed in original crate

### benchmark

For fibonacci remain the same (4194304 improvement just due to noise I guess)
| fibonacci_max_steps_1048576      | 2.0913           | +0.6455% (No change in performance detected) |
| fibonacci_max_steps_2097152      | 3.5427           | -0.3998% (No change in performance detected) |
| fibonacci_max_steps_4194304      | 6.4078           | -10.032% (Performance has improved)          |

For lookup keccak
| keccak_lookup_f_4096             | 0.88398          | -1.5324% (No change in performance detected) |
| keccak_lookup_f_8192             | 1.7557           | -1.2695% (No change in performance detected) |


witness/src/lib.rs
```diff
@@ -293,3 +293,17 @@ impl<F: Sync + Send + Copy + FieldAlgebra> Index<usize> for RowMajorMatrix<F> {
...
+        $ins[$field.id as usize] = $val.into_f();
+    };
+}
+
+#[macro_export]
+macro_rules! set_fixed_val {
+    ($ins:ident, $field:expr, $val:expr) => {
+        $ins[$field.0] = $val;
+    };
+}
```
spherel, line 309: Those macros seem to be defined multiple times all over the workspace.

gkr_iop/src/gkr/layer_constraint_system.rs
```diff
@@ -120,17 +128,90 @@ impl<E: ExtensionField> LayerConstraintSystem<E> {
...
+            let rlc_record = rlc_chip_record(
+                vec![
+                    E::BaseField::from_canonical_u64(LookupTable::U16 as u64).expr(),
+                    value * E::BaseField::from_canonical_u64(1 << (16 - size)).expr(),
+                ],
+                self.alpha.clone(),
+                self.beta.clone(),
+            );
+            self.range_lookups.push(rlc_record)
+        }
```
spherel: Why does it generate two records? Is that to avoid overflow?

gkr_iop/src/gkr/layer_constraint_system.rs
```diff
@@ -270,8 +351,27 @@ impl<E: ExtensionField> LayerConstraintSystem<E> {
...
+        assert_eq!(ram_write_evals.len(), self.ram_write.len(),);
+        assert_eq!(ram_read_evals.len(), self.ram_read.len(),);
+
+        for (idx, ram_expr, ram_eval) in izip!(
+            0..,
+            chain!(self.ram_write.clone(), self.ram_read.clone(),),
+            ram_write_evals.chain(ram_read_evals)
+        ) {
+            self.add_non_zero_constraint(
+                ram_expr - E::BaseField::ONE.expr(), // ONE is for padding
```
spherel: Why -1 here?

gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -129,6 +132,16 @@ pub struct KeccakWitCols<T> {
...
+    pub input_ram_start_addr: [T; 1],
+    pub input_ram_read_ts: [T; KECCAK_INPUT32_SIZE],
+
+    // transition state
+    pub cur_ts: [T; 1],
+    pub pc: [T; 1],
+
+    // VM ecall specs
+    pub read_rs1: [T; 1],
+    pub read_rs2: [T; 1],
```
spherel: I think we can directly write `T` instead of `[T; 1]`.

gkr_iop/src/precompiles/lookup_keccakf.rs
```diff
@@ -527,6 +560,9 @@ impl<E: ExtensionField> ProtocolBuilder<E> for KeccakLayout<E> {
 #[derive(Clone, Default)]
 pub struct KeccakTrace {
     pub instances: Vec<[u32; KECCAK_INPUT32_SIZE]>,
+    pub ram_start_addr: Vec<ByteAddr>,
+    pub cur_ts: Vec<Cycle>,
+    pub read_ts: Vec<[Cycle; KECCAK_INPUT32_SIZE]>,
 }
```
spherel: Is this better to define
```rust
pub struct KeccakInstance {
    pub instance: [u32; KECCAK_INPUT32_SIZE],
    pub ram_start_addr: ByteAddr.
    pub cur_ts: Cycle,
    pub read_ts: [Cycle; KECCAK_INPUT32_SIZE],
}
```
and `type KeccakTrace = Vec<KeccakInstance>;`?

gkr_iop/src/circuit_builder.rs
```diff
@@ -0,0 +1,1019 @@
+use itertools::{Itertools, chain};
```
spherel, line 1: Is this to be merged with `LayerConstraintBuilder`?

gkr_iop/src/gkr/layer_constraint_system.rs
hero78119: it was explained in previous PR https://github.com/scroll-tech/ceno/pull/973#discussion_r2177075904

gkr_iop/src/gkr/layer_constraint_system.rs
hero78119: It was explained previously https://scrollco.slack.com/archives/C064WRCBMHU/p1726064658084059

1. it was previous logic in lookup gkr-iop, not new added.
2. elaboration for instance: when doing u8 range check on u16. let say `expr := 2.inverse()`, apparently it's not in u8. however,
```rust
fn assert_u8(…) {
    assert_u16(expr * Expression::from(1 << 8)); // 2.inverse() will pass the check, so it's sound
    assert_u16(expr);  //  2.inverse() failed on this
}
```
We need 2 constrains to assure the soundness, only any of this is not sound

gkr_iop/src/precompiles/lookup_keccakf.rs
hero78119: thanks for reminding. I will do those in more organized way in next PR https://github.com/scroll-tech/ceno/pull/980/files#diff-52d83d296448263d04812e908937f2a3c37a8a1cc6b78c9c9112308127460e2e so let move the discussion here. In short, I figure out those column should be wrapped into a new column layout`KeccakStateCols` structure to manage column related to global state, instead of mixed them within `KeccakWitsCols`

see screenshot
image

gkr_iop/src/precompiles/lookup_keccakf.rs
hero78119: It's a good idea! I will do it in next ongoing PR as mentioned above https://github.com/scroll-tech/ceno/pull/978#discussion_r2188381908, so let move the discussion and finalized the discussion here.

gkr_iop/src/circuit_builder.rs
hero78119: yes I plan to merge `LayerConstraintBuilder`  with this circuit builder, so 
1. gkr-iop could leverage state read/write (pc, ts), registers read/write logic, and lookup logic from previous well-defined `ConstrainBuilder`.
2. the more formal purpose of `ConstrainSystem` is to serve gkr circuit **frontend**, and finally we convert it to **backend** constrain system should be `Layer`. So layer is more like `BackendConstrainSystem`

gkr_iop/src/gkr/layer_constraint_system.rs
spherel: Got you! Sorry, I forgot..

witness/src/lib.rs
hero78119: yes I had clean up all of duplicated in other place in latest commit. Right now only `witness/src/lib.rs` got this definition

spherel (approved):
LGTM!

PR #980, opened 2025-07-06
[precompile] part2 integrate keccak precompile into e2e flow

hero78119:
### gkr-iop circuit 
This PR complete keccak precompile circuit construction, and try to separate clearly: for zkvm related circuit are construct in "keccak" opcode, while precompile only handle dedicate logic. For witness assignment also same principle, phase 1 witness are also assign separately for better modulariztion

### circuit builder
lookup keccak circuit switch to new circuit builder and not depends on `layer_constrain_system` dependency.

### e2e integration
keccak opcode are integrated into prover flow. It can be invoke via guest program: `cargo run --package ceno_zkvm --release --bin e2e -- --platform=ceno --hints=1,6  examples/target/riscv32im-ceno-zkvm-elf/release/examples/ceno_rt_keccak`

### fixed RC processing
In this PR RC was refactor to use witness first, because I realized the design of expanding fixed poly (and commitment) according to num_instance is very complex because fixed poly in our codebase are assume immutable everywhere. Mutability involve quite of change which also seems unnessesary. In particular, fixed poly immutable should be common sense. Thus, I will try to figure out how to deal with RC (and fixed poly) properly later 

### Next steps (in next PR)
Integrate gkr-iop verifier into verifler flow

### benchmark

```
cargo bench --features jemalloc --bench keccak --package ceno_zkvm -- --baseline baseline
```

| mode                 | # instructions | # cycles  | proof size (mb) | median time (ms) | speedup (↑ better) | proof size ratio (↓ better) |
|----------------------|----------------|-----------|------------------|------------------|--------------------|-----------------------------|
| (baseline) w/o precompile | 1,755,337      | 7,021,352 | 1.95             | 3.5385           | 1.00× (baseline)   | 1.00× (baseline)            |
| precompile           | 732            | 2,932     | 4.51             | 2.18             | 1.62×              | 2.31×                       |

on fibonacci against master remain no change

```bash
cargo bench --features jemalloc --bench fibonacci --package ceno_zkvm -- --baseline baseline
```

| fibonacci_max_steps_1048576      | 2.1139           | +0.5573% (No change in performance detected) |
| fibonacci_max_steps_2097152      | 3.6055           | +1.6858% (Change within noise threshold)     |

### DIFF
--- a/ceno_zkvm/src/tables/ops/ops_impl.rs
+++ b/ceno_zkvm/src/tables/ops/ops_impl.rs
@@ -18,7 +18,6 @@ use multilinear_extensions::{Expression, Fixed, ToExpr, WitIn};
 pub struct OpTableConfig {
     abc: [Fixed; 3],
     mlt: WitIn,
-    rom_type: ROMType,
 }
 
 impl OpTableConfig {
@@ -47,7 +46,7 @@ impl OpTableConfig {
             mlt.expr(),
         )?;
 
-        Ok(Self { abc, mlt, rom_type })
+        Ok(Self { abc, mlt })
     }
 
     pub fn generate_fixed_traces<F: SmallField>(
@@ -78,7 +77,6 @@ impl OpTableConfig {
         let mut witness =
             RowMajorMatrix::<F>::new(length, num_witin, InstancePaddingStrategy::Default);
 
-        println!("rom_type {:?}", self.rom_type);
         let mut mlts = vec![0; length];
         for (idx, mlt) in multiplicity {
             mlts[*idx as usize] = *mlt;


gkr_iop/src/gkr/mock.rs
```diff
@@ -115,12 +96,10 @@ impl<E: ExtensionField> MockProver<E> {
                         &layer.expr_names,
                         &layer.out_eq_and_eval_exprs
                     ) {
-                        if !expect.is_equal(&got) {
+                        if expect != got {
```
spherel, line 99: Is there a case that `expect` and `got` are equal, but one is saved as `FieldType::Base(..)` but the other is saved as `FieldType::Ext(..)`?

gkr_iop/src/utils.rs
```diff
@@ -84,7 +85,11 @@ pub fn extend_exprs_with_rotation<E: ExtensionField>(
             .zip_eq(alpha_pows_iter.by_ref().take(group_length))
             .map(|(expr, alpha)| alpha * expr)
             .sum::<Expression<E>>();
-        zero_check_exprs.push(eq_expr.clone().unwrap() * zero_check_expr);
+        let eq_expr = match eq_expr {
+            Some(Expression::StructuralWitIn(id, ..)) => Expression::WitIn(offset_eq_id + *id),
+            invalid => panic!("invalid eq format {:?}", invalid),
+        };
+        zero_check_exprs.push(eq_expr * zero_check_expr);
```
spherel, line 92: Is there any reason to change it from `Expression::WitIn(..)` to `Expression::StructuralWitIn(..)`?

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -131,14 +134,18 @@ impl<
                     }
                 })
             {
-                num_instances.push((index, num_instance));
+                num_instances.push((
+                    index,
+                    num_instance >> vk.get_cs().rotation_vars().unwrap_or(0),
+                ));
+                num_instances_with_rotation.push((index, num_instance))
```
spherel, line 137: Why compute `num_instances` from `num_instance >> vk.get_cs().rotation_vars().unwrap_or(0),))`?

ceno_zkvm/src/e2e.rs
```diff
@@ -879,8 +881,7 @@ pub fn verify<E: ExtensionField, PCS: PolynomialCommitmentScheme<E> + serde::Ser
         transcript,
         zkvm_proof.has_halt(&verifier.vk),
     )?;
-    // print verification statistics like proof size and hash count
-    tracing::info!("e2e proof stat: {}", zkvm_proof);
```
hero78119, line 883: this was move to `proof ready before verify`, so even verify failed we still can see proof size data

gkr_iop/src/gkr/mock.rs
hero78119: yes you are right and I verified it's syntatic equal.

I added this [commit](https://github.com/scroll-tech/ceno/pull/980/commits/75ebf0b0a611795c028664df0ec0f6fb959f4e84), moving is_equal into `PartialEq` trait definition

gkr_iop/src/utils.rs
hero78119: Do you mean reverse? what it actually do here is convert `StructuralWitIn` to `WitIn`, because gkr-iop backend only recognize `Witin`

ceno_zkvm/src/scheme/prover.rs
hero78119: because what `rmm.num_instances()` return is height, or number of rows. With rotation, the relation between number of rows and real instance are got rotation factor. So to derive real num instances, we need to divide rotation size, thus we do right shit toward rotation num_vars

ceno_zkvm/src/precompiles/lookup_keccakf.rs
```diff
@@ -159,26 +158,20 @@ pub struct KeccakLayer<WitT, FixedT, EqT> {
 
 #[derive(Clone, Debug)]
 pub struct KeccakLayout<E: ExtensionField> {
-    pub layer_exprs: KeccakLayer<WitIn, Fixed, WitIn>,
-    layer_in_evals: [usize; KECCAK_LAYER_SIZE],
-    final_out_evals: KeccakOutEvals<usize>,
-    alpha: Expression<E>,
```
spherel, line 165: `layer_in_evals` and `final_out_evals` both are part of the circuit structure. Shall we keep this?

spherel (commented):
More comments.

ceno_zkvm/src/scheme/prover.rs
spherel: To be honest, I felt confusing when I saw this. I think it's not very natural and currently I only see this holding in the Keccak circuit. Maybe in the future, we will use rotation in other ways (across instances, for example).

ceno_zkvm/src/precompiles/lookup_keccakf.rs
hero78119: per offline discussion, current we dont use it, and this was not removed entirely, as just move to other place where need it

hero78119:
Example command to generate proof with gkr-iop
`RUST_LOG=info cargo run --package ceno_zkvm --release --bin e2e -- --field=baby-bear --platform=ceno  examples/target/riscv32im-ceno-zkvm-elf/release/examples/keccak_syscall`

PR #982, opened 2025-07-13
Feat: commit to all matrices for BaseFold

kunxian-xia:
## Summary

In this pr, we decide to encode all matrices using Reed Solomon code no matter their height. This decision will make it easier for us to write the **recursion** program for verifying BaseFold opening proof. 

In summary, it's **easier** for us to write recursion program (😊 ✅) but it **increases** the proof size(😭 ❌).

### Advantage
1. If we don't encode matrix with small height, then we need to put the full matrix itself in the opening proof (which is called `trivial_proof` in the code). And then we need to verify the correctness of that matrix by computing the merkle root  from matrix then compare the root with `trivial_commits[i]`. It's quite cumbersome for us to compute merkle root in our current DSL.

### Disadvantage
This comes at some price. Even for a matrix with 8 rows, we need to encode it using RS code. And the query phase requires us to sample at least 100 queries (each query includes 1 row). Therefore the proof itself will instead include around 100 rows which is larger than the original matrix.

## Performance (proof size)

|     field             | before          | after          |
|---------------|-------------|-----------|
| final message |       46k        |    1.13k      |
| opening proof.input opening |      432k       |    1024k (137% ⬆️)  |
| opening proof.commit phase |        307k     |     411k   (33% ⬆️)   |
| sumcheck proof |         0.5k    |       0.88k  |
| total                  |      1.06M      |      1.65M    (55% ⬆️) |

## Future work

It's easy to find that the input opening proofs increase a lot, we can reduce this increase by half. Here is the idea.

Currently we pack two rows (`i`th and `i+n/2`th row) in a matrix A into one row in matrix B. Therefore the matrix's width is increased by twice with its height is decreased by half. We then use these two rows to derive the `i`th value of the 1st matrix in the commit phase. 

In fact, we can commit to the original matrix A. And then computes the random linear combination of rows of A as the 1st matrix C1 in the commit phase. This will be the first codeword. In order to derive the 2nd codeword, we just need to  open the `i+n/2`th value in the matrix C1 (aka. sibling value). That is, we add one more merkle tree in the commit phase. 

This will allow us to reduce the number of values opened for the input matrix. Since we have almost 600 columns to open, this should save us about 470k.

This is tracked in https://github.com/scroll-tech/ceno/issues/983.

### DIFF
--- a/ceno_zkvm/benches/riscv_add.rs
+++ b/ceno_zkvm/benches/riscv_add.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, time::Duration};
+use std::time::Duration;
 
 use ceno_zkvm::{
     self,
@@ -77,10 +77,7 @@ fn bench_add(c: &mut Criterion) {
                     for _ in 0..iters {
                         // generate mock witness
                         let num_instances = 1 << instance_num_vars;
-                        let rmms = BTreeMap::from([(
-                            0,
-                            RowMajorMatrix::rand(&mut OsRng, num_instances, num_witin),
-                        )]);
+                        let rmms = vec![RowMajorMatrix::rand(&mut OsRng, num_instances, num_witin)];
 
                         let instant = std::time::Instant::now();
                         let num_instances = 1 << instance_num_vars;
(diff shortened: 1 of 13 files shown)


hero78119 (approved):
LGTM!

PR #984, opened 2025-07-15
Chore: field access adjustment

kunxian-xia:
## Summary

We aims to make fields in data type that used by verifier to be public. These data type are needed by the recursion verifier program.


hero78119 (approved):
👍 LGTM

PR #986, opened 2025-07-18
Support selector in GKR IOP

dreamATD:


gkr_iop/src/utils.rs
```diff
@@ -307,6 +267,56 @@ pub const fn wits_fixed_and_eqs<const N: usize, const M: usize, const Q: usize>(
     (wits, fixed, eqs)
 }
 
+/// TODO this is copy from gkr crate
+/// including gkr crate after gkr clippy fix
```
hero78119: this comment out-of-date and could be clean up 😺

gkr_iop/src/cpu/mod.rs
```diff
@@ -166,12 +165,12 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>>
             // infer current layer output
             let current_layer_output: Vec<
                 Arc<multilinear_extensions::mle::MultilinearExtension<'_, E>>,
-            > = infer_layer_witness(layer, &current_layer_wits, challenges);
+            > = layer.layer_witness(&current_layer_wits, challenges, num_instances);
```
hero78119: Here need to fix

```rust
let num_instances = num_instances >> layer.rotation_cyclic_group_log2;
```
because we need instances, while original num_instances including rotation

gkr_iop/src/gkr/layer.rs
```diff
@@ -266,6 +285,241 @@ impl<E: ExtensionField> Layer<E> {
...
+        challenges: &[E],
+        num_instances: usize,
+    ) -> Vec<ArcMultilinearExtension<'a, E>>
+    where
+        E: ExtensionField,
+    {
+        let out_evals: Vec<_> = self
+            .out_sel_and_eval_exprs
+            .iter()
+            .flat_map(|(sel_type, out_eval)| izip!(iter::once(sel_type), out_eval.iter()))
```
hero78119: Based on the logic, here should be 

```diff
- .flat_map(|(sel_type, out_eval)| izip!(iter::once(sel_type), out_eval.iter()))
+ .flat_map(|(sel_type, out_eval)| izip!(iter::repeat(sel_type), out_eval.iter())
```

gkr_iop/src/selector.rs
```diff
@@ -0,0 +1,142 @@
...
+        evals[wit_id] = eval;
+    }
+}
+
+pub(crate) fn select_from_expression_result<'a, E: ExtensionField>(
+    sel_type: &SelectorType<E>,
+    out_mle: ArcMultilinearExtension<'a, E>,
+    num_instances: usize,
+) -> ArcMultilinearExtension<'a, E> {
+    match sel_type {
```
hero78119, line 136: Per design, our read/write expression already include padding constant. Thus in selector part we just need transparent bypass it, which equivalent to times it with 0 or 1, and without need of padding logic

gkr_iop/src/gkr/layer.rs
```diff
@@ -266,6 +285,241 @@ impl<E: ExtensionField> Layer<E> {
...
+                    // sanity check: zero mle
+                    if cfg!(debug_assertions) {
+                        assert!(
+                            out_mle.evaluations().is_zero(),
+                            "layer name: {}, expr name: \"{expr_name}\" got non_zero mle",
+                            self.name
+                        );
+                    }
+                };
+                select_from_expression_result(sel_type, out_mle, num_instances)
```
hero78119: Another place per my previous proposal, make selector mle construction in `structural_witness` so we can greatfully simplify the logic here, so it's also be more organized for GPU as gkr-witness will happened in gpu instead

gkr_iop/src/gkr/layer.rs
```diff
@@ -266,6 +285,241 @@ impl<E: ExtensionField> Layer<E> {
             };
         }
     }
+
+    pub fn layer_witness<'a>(
+        &self,
+        layer_wits: &[ArcMultilinearExtension<'a, E>],
```
hero78119: `layer_witness` not proper in layer.rs because layer are general for cpu/gpu, for example, `ArcMultilinearExtension<'a, E>` are only for cpu.

hero78119 (commented):
Just left few comments, while I am working on fixing them, along with make test pass in `precompile/lookup_keccak.rs`

hero78119 (approved):
LGTM!!

#### Benchmark 
On fibonacci e2e against master branch => no impact 

| fibonacci_max_steps_1048576      | 2.7596           | -1.1465% (No change in performance detected) |
| fibonacci_max_steps_2097152      | 5.1991           | -0.3432% (No change in performance detected) |
| fibonacci_max_steps_4194304      | 10.303           | -1.5076% (Change within noise threshold)     |

lookup gkr benchmark -> slow 10% due to new selector logic

| Benchmark                        | Median Time (s) | Median Change (%)               |
|----------------------------------|------------------|----------------------------------|
| keccak_lookup_f_4096             | 5.2914           | +11.384% (Performance has regressed) |

guest program with 100 keccak for precompile vs non-precompile

| mode                      | # instructions | # cycles  | proof size (mb) | median time (ms) | speedup (↑ better) | proof size ratio (↓ better) |
|---------------------------|----------------|-----------|------------------|------------------|--------------------|-----------------------------|
| (baseline) w/o precompile | 1,755,337      | 7,021,352 | 2.14             | 4,832.4          | 1.00× (baseline)   | 1.00× (baseline)            |
| precompile                | 732            | 2,932     | 4.97             | 2,140.4          | 2.26×              | 2.32×                       |

PR #989, opened 2025-07-23
unify proving/verifying flow for all opcode

hero78119:
### Change scope
- prover flow: riscv opcode + keccak precompiles share same gkr-iop proving/verify flow, and cleanup bunch of code
- remove `RepeatLast` in opcode witness, because right now we make selector for it and skip padding row zero constraint check
- refactor monomial expression from proving phase to circuit setup phase, which greatfully save runtime and memory cost

### DIFF
--- a/ceno_zkvm/src/scheme/cpu/mod.rs
+++ b/ceno_zkvm/src/scheme/cpu/mod.rs
@@ -538,10 +538,6 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> MainSumcheckProver<C
                 &input.fixed,
                 challenges,
             );
-            dbg!(
-                "gkr_circuit_out.0.0[0] num_var {}",
-                gkr_circuit_out.0.0[0].num_vars()
-            );
             gkr_circuit_out.0.0
         } else {
             // opcode must have at least one read/write/lookup
--- a/ceno_zkvm/src/scheme/cpu/mod.rs
+++ b/ceno_zkvm/src/scheme/cpu/mod.rs
@@ -58,7 +58,7 @@ impl CpuTowerProver {
         }
 
         // XXX to sumcheck batched product argument with logup, we limit num_product_fanin to 2
-        // TODO mayber give a better naming?
+        // TODO maybe give a better naming?
         assert_eq!(num_fanin, 2);
 
         let (prod_specs_len, logup_specs_len) = (prod_specs.len(), logup_specs.len());
@@ -80,6 +80,7 @@ impl CpuTowerProver {
                 logup_specs_len * 2,
             transcript,
         );
+        dbg!(alpha_pows[1].clone());
         let initial_rt: Point<E> = transcript.sample_and_append_vec(b"product_sum", log_num_fanin);
         let (mut out_rt, mut alpha_pows) = (initial_rt, alpha_pows);
 
@@ -527,6 +528,10 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> MainSumcheckProver<C
                 &input.witness,
                 &input.fixed,
                 challenges,
+            );
+            dbg!(
+                "gkr_circuit_out.0.0[0] num_var {}",
+                gkr_circuit_out.0.0[0].num_vars()
             );
             gkr_circuit_out.0.0
         } else {
--- a/ceno_zkvm/src/scheme/cpu/mod.rs
+++ b/ceno_zkvm/src/scheme/cpu/mod.rs
@@ -251,22 +251,18 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TraceCommitter<CpuBa
         PCS::CommitmentWithWitness,
         PCS::Commitment,
     ) {
-        let largest_poly_size = traces
+        let max_poly_size_log2 = traces
             .values()
-            .map(|trace| next_pow2_instance_padding(trace.num_instances()) << 1)
+            .map(|trace| ceil_log2(next_pow2_instance_padding(trace.num_instances())))
             .max()
             .unwrap();
-        let prover_param = if let Some(s) = self.largest_poly_size
-            && s >= largest_poly_size
-        {
-            self.pp.as_ref().unwrap()
-        } else {
-            let (prover_param, _) =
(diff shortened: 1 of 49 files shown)


ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -401,15 +330,17 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<CpuBacke
...
-        exit_span!(record_span);
-
         // infer all tower witness after last layer
         let span = entered_span!("tower_witness_last_layer");
         let mut r_set_last_layer = r_set_wit
             .iter()
             .chain(w_set_wit.iter())
             .map(|wit| {
+                // if is_padded {
+                //     wit.as_view_chunks(NUM_FANIN)
```
hero78119: Here got lifetime annotation error thus temporarily comment out this optimisation, which doesn't break soundness

ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -294,95 +291,27 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<CpuBacke
...
         Vec<TowerProverSpec<'b, CpuBackend<E, PCS>>>,
-    ) {
+    )
+    where
+        'a: 'b,
+    {
         let ComposedConstrainSystem {
             zkvm_v1_css: cs, ..
         } = composed_cs;
-        let num_instances = input.num_instances;
```
hero78119, line 304: All below logic move into `build_main_witness`, and gkr_iop do not need this, as the logic is the same as `gkr_witness`

ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -583,6 +517,124 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<CpuBacke
...
+            let (_, gkr_circuit_out) = Self::gkr_witness(
+                gkr_circuit,
+                num_instances_with_rotation,
+                &input.witness,
+                &input.fixed,
+                &input.public_input,
+                challenges,
+            );
+            (gkr_circuit_out.0.0, true)
+        } else {
```
hero78119, line 578: `else` condition only needed for table circuit, because table circuit do not need main-sumcheck, thus I didn't refactor it into gkr_iop framework at this moment

gkr_iop/src/cpu/mod.rs
```diff
@@ -121,36 +120,42 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>>
...
-                                .into_mle()
-                                .into(),
-                        ),
-                    )
-                })
-                .collect::<HashMap<_, _>>()
-                .into_iter()
-                .for_each(|(witin, optional_mle)| witness_mle_flatten[witin] = optional_mle);
+
+            // first_layer
```
hero78119, line 138: We dont have fixed mle in gkr-iop at this moment, thus temporarily comment out and fixed its compile error later

gkr_iop/src/gkr/layer/zerocheck_layer.rs
```diff
@@ -41,36 +43,90 @@ pub(crate) struct RotationClaims<E: ExtensionField> {
...
+            Some(rotation_selector_expr * rotation_expr)
+        } else {
+            None
+        };
+
+        // build main sumcheck expression
+        let alpha_pows_expr = (2..)
+            .take(self.exprs.len() + num_rotations * ROTATION_OPENING_COUNT)
+            .map(|id| Expression::Challenge(id as ChallengeId, 1, E::ONE, E::ZERO))
+            .collect_vec();
```
spherel, line 103: Is `Expression::Challenge(id as ChallengeId, 1, E::ONE, E::ZERO)` or `Expression::Challenge(1, id, E::ONE, E::ZERO)`?

spherel (approved):
LGTM!

gkr_iop/src/gkr/layer/zerocheck_layer.rs
hero78119: format: `Expression::Challenge(challenge_id, power, scalar, offset)`, thus `Expression::Challenge(id as ChallengeId, 1, E::ONE, E::ZERO)` mean `1 * c^1 + 0 = c` which is expected result :)

hero78119:
Update benchmark result, run with fibonacci e2e

```
| fibonacci_max_steps_1048576      | 2.7085           | +3.6336% (Performance has regressed)         |
| fibonacci_max_steps_2097152      | 5.0492           | +3.5620% (Performance has regressed)         |
| fibonacci_max_steps_4194304      | 9.8169           | +3.0887% (Performance has regressed)         |
```

I verified the regressed due to below gkr_witness implementation

https://github.com/scroll-tech/ceno/blob/19b755d5172d98406e18e908956aa5f9fdba6489/gkr_iop/src/cpu/mod.rs#L202-L212

As we compute $$(out[y] - pad) = \sum_x sel * (input[x] - pad)$$ in 2 steps
1. compute $$\sum_x sel * (input[x] - pad)$$, store in `out_mle`
2. take `out_mle`, add with padding `pad`

Which cause 2 traverse. We can optimise this part by one shot, as we only need `gkr_out`, and can skip `layer_witness`.
We can left it as next step optimisation

ceno_zkvm/src/scheme/cpu/mod.rs
hero78119: This was addressed in https://github.com/scroll-tech/ceno/pull/989/commits/19b755d5172d98406e18e908956aa5f9fdba6489

gkr_iop/src/gkr/layer/zerocheck_layer.rs
hero78119: Future work: consider build rlc challenge with same challenge id and leverage power.

Relevant work: optimize these part to cache challenge results and avoid compute incremental power from scratch
https://github.com/scroll-tech/ceno/blob/09bcac43f06c3803fd5c37928916dda0cf239ede/multilinear_extensions/src/expression/utils.rs#L54-L55

PR #923, opened 2025-04-24
feat: unify cpu / gpu provers under same framework

kunxian-xia:
# Summary

We want to unify cpu prover and gpu prover under same framework, therefore we introduce two traits 
- `trait ProverBackend`: this trait is mainly used for defining types that are common across all devices and types that are specific to device.
      
- `trait ProverDevice` which itself is composed from 5 smaller traits.
     - `trait TraceCommitter<PB>`: commit to the main trace matrices.
     - `trait TowerProver<PB>`: run GKR iop for product argument and logup argument.
     - `trait MainSumcheckProver<PB>`: prove the relation between last layer of tower tree with the trace matrices.
     - `trait OpeningProver<PB>`: generate the opening proof for trace matrices at multiple points.
     - `trait DeviceTransporter<PB>`: transport the proving key to device.

## Performance

We noted there are about 2% performance **regression** after this PR.

```
fibonacci_max_steps_2097152/prove_fibonacci/fibonacci_max_steps_2097152
                        time:   [4.3981 s 4.4283 s 4.4620 s]
                        change: [+1.8861% +2.6677% +3.4125%] (p = 0.00 < 0.05)
                        Performance has regressed.

fibonacci_max_steps_4194304/prove_fibonacci/fibonacci_max_steps_4194304
                        time:   [8.1119 s 8.1328 s 8.1534 s]
                        change: [+1.5970% +2.3411% +3.0799%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

PR #993, opened 2025-07-29
refactor proving backend

hero78119:
Fixed the latency regressed reported in https://github.com/scroll-tech/ceno/pull/923#issue-3016452781 
The reason is because PCS::trim was executed every time during prover-backend initialization.

This PR refactor proving device as reusable across whole program live cycle.

### benchmark

with benchmark command of fibonacci e2e
`JEMALLOC_SYS_WITH_MALLOC_CONF=retain:true,metadata_thp:always,thp:always,dirty_decay_ms:-1,muzzy_decay_ms:-1,abort_conf:true cargo bench --features jemalloc --bench fibonacci --package ceno_zkvm -- --baseline baseline`

| fibonacci_max_steps_1048576      | 2.6131           | -3.29% (Performance has improved)            |
| fibonacci_max_steps_2097152      | 4.8859           | -2.85% (Performance has improved)            |
| fibonacci_max_steps_4194304      | 9.5519           | -4.56% (Performance has improved)            |

### DIFF
--- a/ceno_zkvm/src/e2e.rs
+++ b/ceno_zkvm/src/e2e.rs
@@ -600,20 +600,20 @@ impl<E: ExtensionField> E2EProgramCtx<E> {
         (pk, vk)
     }
 
-    pub fn keygen_with_pd<
+    pub fn keygen_with_pb<
         PCS: PolynomialCommitmentScheme<E> + 'static,
         PB: ProverBackend<E = E, Pcs = PCS> + 'static,
     >(
         &self,
-        pd: &PB,
+        pb: &PB,
     ) -> (ZKVMProvingKey<E, PCS>, ZKVMVerifyingKey<E, PCS>) {
         let pk = self
             .system_config
             .zkvm_cs
             .clone()
             .key_gen::<PCS>(
-                pd.get_pp().clone(),
-                pd.get_vp().clone(),
+                pb.get_pp().clone(),
+                pb.get_vp().clone(),
                 self.zkvm_fixed_traces.clone(),
             )
             .expect("keygen failed");
@@ -675,7 +675,7 @@ pub fn run_e2e_with_checkpoint<
 
     // Keygen
     let start = std::time::Instant::now();
-    let (pk, vk) = ctx.keygen_with_pd(device.get_pb());
+    let (pk, vk) = ctx.keygen_with_pb(device.get_pb());
     tracing::debug!("keygen done in {:?}", start.elapsed());
 
     let start = std::time::Instant::now();
--- a/ceno_zkvm/src/scheme/cpu/mod.rs
+++ b/ceno_zkvm/src/scheme/cpu/mod.rs
@@ -256,7 +256,7 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TraceCommitter<CpuBa
     ) {
         let max_poly_size_log2 = traces
             .values()
-            .map(|trace| ceil_log2(trace.num_instances()) + 1)
+            .map(|trace| ceil_log2(next_pow2_instance_padding(trace.num_instances())))
             .max()
             .unwrap();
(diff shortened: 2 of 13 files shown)


gkr_iop/src/cpu/mod.rs
```diff
@@ -59,22 +69,24 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ProverBackend for Cp
...
 }
 
 impl<PB: ProverBackend> CpuProver<PB> {
-    pub fn new(backend: PB) -> Self {
-        Self {
-            backend,
-            pp: None,
-            largest_poly_size: None,
-        }
+    pub fn new(backend: &'static PB) -> Self {
```
kunxian-xia: why not pass ownership?

ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -254,22 +254,18 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TraceCommitter<CpuBa
         PCS::CommitmentWithWitness,
         PCS::Commitment,
     ) {
-        let largest_poly_size = traces
+        let max_poly_size_log2 = traces
             .values()
-            .map(|trace| next_pow2_instance_padding(trace.num_instances()) << 1)
+            .map(|trace| ceil_log2(trace.num_instances()) + 1)
```
kunxian-xia: If `trace.num_instances() = 1`, then `ceil_log2()` will return 0, but we expect it to return 1.

ceno_zkvm/src/e2e.rs
```diff
@@ -599,6 +600,27 @@ impl<E: ExtensionField> E2EProgramCtx<E> {
         (pk, vk)
     }
 
+    pub fn keygen_with_pd<
```
kunxian-xia: typo? Should it be `keygen_with_pb`?

ceno_zkvm/src/e2e.rs
```diff
@@ -599,6 +600,27 @@ impl<E: ExtensionField> E2EProgramCtx<E> {
...
+        PCS: PolynomialCommitmentScheme<E> + 'static,
+        PB: ProverBackend<E = E, Pcs = PCS> + 'static,
+    >(
+        &self,
+        pd: &PB,
```
kunxian-xia: ditto.

ceno_zkvm/src/scheme/cpu/mod.rs
kunxian-xia: I think the old code should not `+1` as the trim function always expect it's the message's size.

gkr_iop/src/cpu/mod.rs
hero78119: because we dont need to mutate proving-backend, and it only need to be initialized once, then pass to proving-device.
proving device will be instantiated per proof. Thus with this function signature we can avoid instantiate proving backend.

hero78119: However thanks for reminding 👍  I just add https://github.com/scroll-tech/ceno/pull/993/commits/5afd4f44632df6237f2d08a2e032d2b133fc62b0 with Rc<PB> is more proper fix for rust style and let rust maintain lifetime for us

ceno_zkvm/src/scheme/cpu/mod.rs
hero78119: Nice catch and added in https://github.com/scroll-tech/ceno/pull/993/commits/5afd4f44632df6237f2d08a2e032d2b133fc62b0

hero78119: you are right and we should remove +1, Done in https://github.com/scroll-tech/ceno/pull/993/commits/5afd4f44632df6237f2d08a2e032d2b133fc62b0

kunxian-xia (approved):
LGTM 🎉

PR #995, opened 2025-08-01
refactor all recursive based witness inference to index based 

hero78119:
Followed up on #994, which migrate all recursive based witness inference to evaluation based.
It also fixed some issue where public io convert to monomial form.

As public io could be 2 form
- scalar => introduce `Expression::InstanceScalar` => used in `halt` opcode to constrain `end_pc`. 
- vector => reuse existing `Expression::Instance`, and we use this vector in RAM and define a specific range for it.

As 2 form are used in different purpose, we need to differentiate them when convert to monomial form.
Previous monomial form didnt cover it properly. I found this issue when migrate table circuit into new `wit_infer_by_expr`, thus also fixed in this PR

### benchmark 
e2e shows no impact 

| fibonacci_max_steps_1048576      | 2.4354           | -1.62% (Change within noise threshold)       |
| fibonacci_max_steps_2097152      | 4.5743           | -1.11% (No change in performance detected)   |
| fibonacci_max_steps_4194304      | 8.7778           | -2.66% (Performance has improved)            |


multilinear_extensions/src/expression/monomial.rs
```diff
@@ -28,14 +28,14 @@ impl<E: ExtensionField> Expression<E> {
     fn distribute(&self) -> Vec<Term<Expression<E>, Expression<E>>> {
         match self {
             // only contribute to scalar terms
-            Constant(_) | Challenge(..) | Instance(_) => {
+            Constant(_) | Challenge(..) | InstanceScalar(_) => {
```
hero78119, line 31: Found an improper handling for Instance (public io). For we have 2 type
- vec length = 1
- vec length > 1

we should put vec length = 1 into scalar, while make > 1 length put in product part.
This issue was only trigger in this PR, after I try to use monomial term for `table circuit` lookup witness generation

gkr_iop/src/cpu/mod.rs
```diff
@@ -241,26 +242,24 @@ where
...
-                EvalExpression::Linear(_, _, _) | EvalExpression::Single(_) => {
-                    eval_by_monomial_expr(expr, layer_wits, pub_io_evals, challenges)
-                }
-                EvalExpression::Zero => MultilinearExtension::default().into(),
-                EvalExpression::Partition(_, _) => unimplemented!(),
-            };
-            if let EvalExpression::Zero = out_eval {
-                // sanity check: zero mle
-                if cfg!(debug_assertions) {
+            if cfg!(debug_assertions) {
```
hero78119, line 245: Here only swap the procedure without modify logic, to make thing more organized

multilinear_extensions/src/expression.rs
```diff
@@ -890,14 +909,16 @@ impl<E: ExtensionField> Mul for Expression<E> {
             ),
             // instance * witin
             // instance * fixed
-            (c @ Expression::Instance(..), w @ Expression::WitIn(..))
```
hero78119, line 893: `Instance` now only serve vector pub-io used in RAM. For  here it just scalar, thus replace with `InstanceScalar`

multilinear_extensions/src/expression.rs
```diff
@@ -1066,6 +1087,12 @@ impl<E: ExtensionField> ToExpr<E> for &Fixed {
 impl<E: ExtensionField> ToExpr<E> for Instance {
     type Output = Expression<E>;
     fn expr(&self) -> Expression<E> {
```
hero78119, line 1089: make default `expr()` of Instance  return `Expression::InstanceScalar`, and define `Expression::Instance` with a separate `expr_as_instance()`

spherel (approved):
LGTM!

PR #1001, opened 2025-08-05
fix lk-multiplicity deep_clone bug

hero78119:
Current master branch e2e failed when `MOCK_PROVING=1` was set.

The root cause is arised from `Multiplicity::deep_clone` which only clone current thread `ThreadLocal` and discard all the other thread result. When test on `mock_prover`  with small data + single thread, everything works fine, however it got failed when run in multi-thread mode. 

To fixed it, we can convert Multiplicity to single thread at the end of `assign_instances`, then we use pure clone.

This PR also fix several issues of mock_prover, and add `MOCK_PROVING=1` to ci for debug build.

### benchmark 
e2e fibonacci remain the same
| Benchmark                        | Median Time (s) | Median Change (%)                |
|----------------------------------|------------------|-----------------------------------|
| fibonacci_max_steps_1048576      | 2.4446           | -1.25% (Change within noise threshold) |
| fibonacci_max_steps_2097152      | 4.6252           | -0.01% (No change in performance detected) |
| fibonacci_max_steps_4194304      | 8.9366           | -0.90% (No change in performance detected) |

### DIFF
--- a/ceno_zkvm/src/structs.rs
+++ b/ceno_zkvm/src/structs.rs
@@ -322,19 +322,14 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
     }
 
     // merge the multiplicities in each opcode circuit into one
-    pub fn finalize_lk_multiplicities(&mut self, is_keep_raw_lk_mlts: bool) {
+    pub fn finalize_lk_multiplicities(&mut self) {
         assert!(self.combined_lk_mlt.is_none());
         assert!(!self.lk_mlts.is_empty());
 
         let mut combined_lk_mlt = vec![];
         let keys = self.lk_mlts.keys().cloned().collect_vec();
         for name in keys {
-            let lk_mlt = if is_keep_raw_lk_mlts {
-                // mock prover needs the lk_mlt for processing, so we do not remove it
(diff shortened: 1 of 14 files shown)


ceno_zkvm/src/structs.rs
```diff
@@ -332,13 +331,9 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
         for name in keys {
             let lk_mlt = if is_keep_raw_lk_mlts {
                 // mock prover needs the lk_mlt for processing, so we do not remove it
-                self.lk_mlts
-                    .get(&name)
-                    .unwrap()
-                    .deep_clone()
-                    .into_finalize_result()
+                self.lk_mlts.get(&name).cloned().unwrap()
```
kunxian-xia: I think there's no need to call `.cloned()` since we will call `to_vec()` or `iter()` later which only need a immutable reference.

hero78119: good idea and thanks for remind. Update PR with commit https://github.com/scroll-tech/ceno/pull/1001/commits/8aeb178ec742b30d1bf8d8b4f152964f2ab2e9b1

PR #1011, opened 2025-08-12
use par_iter_mut instead of scoped spawns in phase1 of sumcheck

montekki:
Hi all, I have encountered a racy behaviour of workers in phase1 of sumcheck prover that essentially forbids running two sumcheck proving processes in a parallel because of a potential deadlock. The nature of the deadlock is that the rayon tasks are not that well designed for communicating between each other using channels, as discussed [here.](https://github.com/rayon-rs/rayon/issues/1258).

`ceno` starts racing if two or more provers are being run in the following situation:
* consider you have a machine such as a 12-core Mac
* you want to launch two proving processes each with 8 workers in phase1
* rayon pool by default will likely contain 12 worker threads
* first ceno prover manages to spawn 7 worker tasks
* second ceno worker manages to spawn another 5 tasks
* however the main worker in both of them waits for evaluations of the first iteration from _all_ workers and _all_ workers do not exist in eather of the provers.
* both processes are now hanging deadlocked.

A repro can be found [here](https://github.com/montekki/ceno-sum-rayon-test).

This patch uses `par_iter_mut` instead of the existing workload architecture and does not use blocking channels. It also simplifies code somewhat unifying it for main and non-main workers and removing the syncronised transcript structure.


hero78119:
Hi, thanks and it's really excited to see an great potential improvement on this 👍 
It's an known issue that existing workload architecture doesn't work on platform which is not pow2 cores, e.g. 12-core Mac which will be deadlock.

We explore the similar architecture before with below pesudo flow
```rust
let mut next_challaneg = None
for 0 to n {  // sumcheck with n variables
   // fix variable for round > 0, then compute unit poly for each round
   let unit_poly_all_workers = (0..n).par_iter_mut().map(|worker| worker.compute_uni_poly(next_challaneg)).collect_vec();
   let unit_polys = unit_poly_all_workers.iter().sum(); // sum all unit poly 
   // then derive challenge from transcript 
   next_challaneg = Some(trannscript.observe(unit_polys).get_challenge())
}
```
With that the codebase could be greatly simplify, however from benchmark I found this model is slower. I didn't dig into much detail for why it's slow, just guess this might have more hard-syncronized point so some worker might goes to idle state easily.

Could you also kindly help to attach benchmark https://github.com/scroll-tech/ceno/blob/master/sumcheck/benches/devirgo_sumcheck.rs on this change against master branch implementation :) ?
Thanks!!

montekki:
for a better repr of differences here is the `critcmp` between `master` and this PR
Screenshot 2025-08-13 at 11 56 53

transcript/src/syncronized.rs
```diff
@@ -1,122 +0,0 @@
-use std::array;
-
-use crossbeam_channel::{Receiver, Sender, bounded};
```
hero78119, line 3: please also help to remove `crossbeam-channel` from `Cargo.toml`

hero78119 (approved):
The polishing sumcheck implementation in particular for the good readibility!

Yes you are right :) besides I verified in another AMD 5900XT cores on e2e fibonacci it shows no impact :)

| Benchmark                        | Median Time (s) | Median Change (%)                |
|----------------------------------|------------------|-----------------------------------|
| fibonacci_max_steps_1048576      | 2.5775           | +2.54% (No change in performance) |
| fibonacci_max_steps_2097152      | 4.7062           | +0.96% (No change in performance) |
| fibonacci_max_steps_4194304      | 9.1287           | +0.42% (No change in performance) |
 
just left a nitpick and need to fix lint error, overall LGTM!

transcript/src/syncronized.rs
montekki: Done

PR #1017, opened 2025-08-13
fix smart slice serialization implementation

montekki:
`SmartSlice` has a custom `Deserialize` implementation that deserializes `Vec<T>`. `Serialize` is simply derived on the enum, so the implementations are not compatible. This patch adds `#[serde(untagged)]` to the enum to fix it without breaking changes.


hero78119 (approved):
LGTM!

PR #1020, opened 2025-08-15
Remove unnecessary heap allocations

c410-f3r:
It is not necessary to collect into `Vec` to perform a simple `sum` operation

### DIFF
--- a/mpcs/src/util.rs
+++ b/mpcs/src/util.rs
@@ -103,19 +103,22 @@ pub(crate) fn codeword_fold_with_challenge<E: ExtensionField>(
 
 #[cfg(any(test, feature = "benchmark"))]
 pub mod test {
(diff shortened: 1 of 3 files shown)


mpcs/src/util.rs
```diff
@@ -103,19 +103,22 @@ pub(crate) fn codeword_fold_with_challenge<E: ExtensionField>(
 
 #[cfg(any(test, feature = "benchmark"))]
 pub mod test {
-    #[cfg(test)]
-    use crate::util::{base_to_usize, u32_to_field};
     use ff_ext::FromUniformBytes;
-    use p3::field::FieldAlgebra;
```
c410-f3r, line 109: `FieldAlgebra` is only used by a test and now is gated as such.

PR #1031, opened 2025-08-22
Remove unused dependencies

c410-f3r:
Most dependencies are not used anywhere


ceno_zkvm/benches/riscv_add.rs
```diff
@@ -18,20 +18,18 @@ use rand::rngs::OsRng;
...
-      targets = bench_add
-    }
-  } else {
-    criterion_group! {
-      name = op_add;
-      config = Criterion::default().warm_up_time(Duration::from_millis(3000));
-      targets = bench_add
-    }
-  }
+#[cfg(feature = "flamegraph")]
```
c410-f3r, line 21: `cfg-if` is only used here and the scenario looks simple enough to apply `cfg` manually

PR #1033, opened 2025-08-24
Docs: explain prover's design for precompiles

kunxian-xia:


kunxian-xia:
trigger the integration test for gkr keccak

gkr_iop/src/gkr/layer/zerocheck_layer.rs
```diff
@@ -283,6 +286,8 @@ impl<E: ExtensionField> ZerocheckLayer<E> for Layer<E> {
         );
         let in_point = in_point.into_iter().map(|c| c.elements).collect_vec();
 
+        // TODO: is there any check on eq_left, eq_right, eq for rotation argument?
```
kunxian-xia, line 291: @hero78119 It seems to me that `eq_left / eq_right / eq_target` is not checked at all.

hero78119: rotation eq will be re-construct when the random point was derived from previous layer

https://github.com/scroll-tech/ceno/blob/ca7b038cd5d308423114c978e34354a54a6291f1/gkr_iop/src/gkr/layer/cpu/mod.rs#L178-L198

in summary, eq will transform from A to B format
- format A: each eq will be assign as selector with only 0, 1 during witness generation to serve layer by layer witness inference
- format B: once random point $$r$$ derive from previous layer, each eq will be constructed via `build_eq(r)` then proceed sumcheck

kunxian-xia: I think the array `self.out_sel_and_eval_exprs` does not include `eq_left / eq_right / eq`.

https://github.com/scroll-tech/ceno/blob/ca7b038cd5d308423114c978e34354a54a6291f1/gkr_iop/src/gkr/layer/zerocheck_layer.rs#L287-L297

PR #1034, opened 2025-08-25
fix lookup keccak rotation to use max 16 limb

hero78119:


.github/workflows/tests.yml
```diff
@@ -43,6 +43,10 @@ jobs:
         run: |
           cargo make --version || cargo install cargo-make
       - name: run test
+        env:
```
hero78119, line 46: `lookup_keccak` involving some deep recursion when build the circuit. I think it's a existing issue, just the new change hit the threshold unluckily. With that, we temporarily increase `RUST_MIN_STACK` to make tests pass. It doesn't break functional correctness.

Makefile.toml
```diff
@@ -12,6 +12,7 @@ args = [
   "--workspace",
 ]
 command = "cargo"
+env = { RUST_MIN_STACK = "33554432" }
```
hero78119, line 15: `lookup_keccak` involving some deep recursion when build the circuit. I think it's a existing issue, just the new change hit the threshold unluckily. With that, we temporarily increase RUST_MIN_STACK to make tests pass. It doesn't break functional correctness.

gkr_iop/src/circuit_builder.rs
```diff
@@ -1073,25 +1092,26 @@ impl<'a, E: ExtensionField> CircuitBuilder<'a, E> {
...
                     .collect_vec();
-                let cnt = rep_x.len() / 2;
-                let rhs = &rep_x[cnt * i..cnt * (i + 1)];
+                let rhs_limbs = take_til_threshold(&mut rep_x_iter, 16, &|limb| limb.0).unwrap();
 
-                assert_eq!(rhs.iter().map(|e| e.0).sum::<usize>(), 32);
+                assert_eq!(rhs_limbs.iter().map(|e| e.0).sum::<usize>(), 16);
 
-                self.require_reps_equal::<32, _, _>(
+                self.require_reps_equal::<16, _, _>(
```
hero78119, line 1110: key highlights: constrain on 16 bits limb individually instead of 32 bits (limb_16 + (1 << 16( * limb_16) because later are larger than some finite field characteristic, e.g. babybear

ceno_zkvm/src/precompiles/lookup_keccakf.rs
```diff
@@ -834,13 +835,20 @@ where
                                     .convert(sizes.clone())
                                     .values();
                             for (j, size) in sizes.iter().enumerate() {
-                                if *size != 32 {
-                                    lk_multiplicity.assert_ux_in_u16(*size, rep[j]);
+                                match *size {
+                                    32 | 1 => (),
```
spherel, line 839: Why not checking it when `size` is 1?

gkr_iop/src/circuit_builder.rs
```diff
@@ -1058,8 +1059,29 @@ impl<'a, E: ExtensionField> CircuitBuilder<'a, E> {
...
+                }
+                5 => {
+                    self.assert_ux::<_, _, 5>(|| format!("{}_{}", name().into(), i), elem.clone())?
+                }
+                1 => self.assert_bit(|| format!("{}_{}", name().into(), i), elem.clone())?,
+                _ => self.assert_ux_in_u16(
+                    || format!("{}_{}", name().into(), i),
+                    *size,
+                    elem.clone(),
+                )?,
```
spherel, line 1084: Shall we put this in a helper function?

ceno_zkvm/src/precompiles/lookup_keccakf.rs
hero78119: because size 1 means bit check, and we compiled it into `(1-expr)*expr` with zero check already. So size 1 we can skip `multiplicity += 1`

gkr_iop/src/circuit_builder.rs
hero78119: yes will do it in next PR https://github.com/scroll-tech/ceno/pull/1036 to fix precompile e2e sound

PR #1036, opened 2025-08-25
fix keccak precompile e2e

hero78119:
build on top of #1034 & #1035 & #1038

This PR fix keccak precompile e2e error, some previous left over.

- [x] rom check error => fixed ecall instrunction specs
- [x] logup sum error => The logup sum != 0 error, see https://github.com/scroll-tech/ceno/pull/1036/files#r2301146794. Need to fix ceno-recursion verifier as well
- [x] prod_r != prod_w => remove X11 read, which cause ts to bump and mismatch with circuit definition. https://github.com/scroll-tech/ceno/pull/1036/files#r2301152148


ceno_zkvm/src/scheme/verifier.rs
```diff
@@ -215,9 +215,16 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMVerifier<E, PCS>
             let input_opening_point = if circuit_vk.get_cs().is_opcode_circuit() {
                 // getting the number of dummy padding item that we used in this opcode circuit
                 let num_lks = circuit_vk.get_cs().num_lks();
-                let num_padded_instance =
```
hero78119, line 220: Need to refresh dummy logic when there got rotation

@kunxian-xia @darth-cy recursion-verifier need to fix the respective logic as well 
https://github.com/scroll-tech/ceno-recursion-verifier/blob/d7aac029fbd3acb8860af9bec3784e0790c6edc4/src/zkvm_verifier/verifier.rs#L237-L241

With that, keccak precompile should be e2e full pass

ceno_emul/src/syscalls/keccak_permute.rs
```diff
@@ -55,22 +55,12 @@ impl From<KeccakState> for [Word; KECCAK_WORDS] {
...
-    assert_eq!(vm.peek_register(Platform::reg_arg1()), 0);
-
     // Read the argument `state_ptr`.
-    let reg_ops = vec![
-        WriteOp::new_register_op(
-            Platform::reg_arg0(),
-            Change::new(state_ptr, state_ptr),
-            0, // Cycle set later in finalize().
-        ),
-        WriteOp::new_register_op(
```
hero78119, line 68: This is just to compatible with sp1 binary logic to assure x11 (a2) register are 0.
In Ceno this check seems unnecessary

ceno_zkvm/src/precompiles/lookup_keccakf.rs
```diff
@@ -772,7 +766,7 @@ where
                             c_aux64[i][j] = state64[j][i] ^ c_aux64[i][j - 1];
                             for k in 0..8 {
                                 lk_multiplicity
-                                    .lookup_xor_byte(state8[j][i][k], c_aux8[i][j - 1][k]);
```
hero78119, line 783: Previous xor operator order mismatch with circuit definition so also lead to lookup failed even functionality equivalent

gkr_iop/src/utils/lk_multiplicity.rs
```diff
@@ -179,22 +179,34 @@ impl LkMultiplicity {
...
+            return;
+        }
+        self.increment(LookupTable::Dynamic, (1 << max_bits) + v);
     }
 
     #[inline(always)]
     pub fn assert_ux_v2(&mut self, v: u64, max_bits: usize) {
+        if max_bits == 1 {
+            return;
+        }
```
kunxian-xia: why return directly if `max_bits = 1`?

ceno_zkvm/src/precompiles/lookup_keccakf.rs
```diff
@@ -48,7 +48,7 @@ use crate::{
 };
 
 pub const ROUNDS: usize = 24;
-const ROUNDS_CEIL_LOG2: usize = 5; // log_2(2^32)
+pub const ROUNDS_CEIL_LOG2: usize = 5; // log_2(2^32)
```
kunxian-xia: The comment is wrong.

gkr_iop/src/utils/lk_multiplicity.rs
hero78119: because we constrain bit via 'a * (1-a)' without need of lookup

ceno_zkvm/src/instructions/riscv/ecall/keccak.rs
```diff
@@ -71,34 +72,32 @@ impl<E: ExtensionField> Instruction<E> for KeccakInstruction<E> {
...
 
         // fetch
         cb.lk_fetch(&InsnRecord::new(
             vm_state.pc.expr(),
             InsnKind::ECALL.into(),
-            Some(E::BaseField::ZERO.expr()),
-            E::BaseField::ZERO.expr(),
-            E::BaseField::ZERO.expr(),
-            E::BaseField::ZERO.expr(),
+            None,
```
kunxian-xia, line 97: why this change? `None` is automatically translated to `RD_NULL`.

gkr_iop/src/utils/lk_multiplicity.rs
kunxian-xia: how about we panic in this case to avoid any confusion?

kunxian-xia (approved):

ceno_zkvm/src/precompiles/lookup_keccakf.rs
hero78119: Nice catch and fixed https://github.com/scroll-tech/ceno/pull/1036/commits/5fb131fb54ba44776c4aa9d304d5250e3303e1a7

PR #1038, opened 2025-08-26
mock prover full check for opcode with rotation

hero78119:
mock-prover take selector to differentiate which row is activated, a necessary feature for rotation-based precompile to work


ceno_zkvm/src/instructions.rs
```diff
@@ -54,20 +54,23 @@ pub trait Instruction<E: ExtensionField> {
...
+                (r_len + w_len..r_len + w_len + lk_len).collect_vec(),
                 // zero_record
-                (selector_type, (0..zero_len).collect_vec()),
+                (0..zero_len).collect_vec(),
             ],
             Chip::new_from_cb(cb, 0),
         );
 
+        // register selector to legacy constrain system
+        cb.cs.r_selector = Some(selector_type.clone());
```
hero78119, line 77: `cb.cs` is more like frontend constrain system. mock_prover retrieve constrain from this, so we add selector meta info for mock_prover to do more precise check

hero78119:
will be covered in #1036

PR #1043, opened 2025-08-28
Add Error enum to `multilinear_extensions`

c410-f3r:
Fixes a TODO

```
// TODO: return error.
assert!(
    return Err(crate::Error::InvalidSizeOfPartialPoint);
    partial_point.len() <= self.num_vars(),
}
```

### DIFF
--- a/multilinear_extensions/src/mle.rs
+++ b/multilinear_extensions/src/mle.rs
@@ -492,12 +492,10 @@ impl<'a, E: ExtensionField> MultilinearExtension<'a, E> {
         list
     }
 
-    pub fn fix_variables(&self, partial_point: &[E]) -> Self {
-        // TODO: return error.
-        assert!(
-            partial_point.len() <= self.num_vars(),
-            "invalid size of partial point"
-        );
+    pub fn fix_variables(&self, partial_point: &[E]) -> crate::Result<Self> {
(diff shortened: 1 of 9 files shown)


multilinear_extensions/src/mle.rs
```diff
@@ -492,12 +492,10 @@ impl<'a, E: ExtensionField> MultilinearExtension<'a, E> {
         list
     }
 
-    pub fn fix_variables(&self, partial_point: &[E]) -> Self {
-        // TODO: return error.
-        assert!(
```
hero78119, line 497: This change is good. But the original purpose is also change other functions in `multilinear_extensions`, e.g. `fix_variables_in_place`, `evaluate`, ... most of the function return the error properly instead of assert and panic

c410-f3r: I was planning to do this in a following PR if that is not a problem

hero78119: Got it and make sense to had it in following PR, as changing other function involve more change across different modules 👍

hero78119: hi just a heads up: A refactor is on the way to move `multilinear_extensions` into another repo as shared crates.
 https://github.com/scroll-tech/gkr-backend/tree/main/crates/multilinear_extensions

PR #1054, opened 2025-09-05
Ceno book.

yczhangsjtu:
This is a draft and outline of the developer document for Ceno.


docs/src/your-first-zk-program.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+This will produce an ELF file at `examples/target/riscv32im-ceno-zkvm-elf/release/fibonacci`. This is the file that is executed by the `run` command.
+
+## Running the Program
+
+As you saw in the previous chapter, you can run the program with `cargo ceno run`:
+
+```sh
+cargo ceno run --example fibonacci --hints=10 --public-io=4191
```
kunxian-xia, line 108: showcase how to pass `hints & public-io` as files

yczhangsjtu: Hints from file is done. Public IO as file is not available yet.

yczhangsjtu:
Some examples in the document failed. Summary of the current status:

- The Fibonacci example:
  - Using --hints=10 prove and verify successfully
  - Using --hints-file hints.bin the prover fails and reports the same error message as the verifier in the next case
- The example in generating-and-verifying-proofs.md:
  - Using --hints=5 the prover succeeds and claims the proof is verified successfully, but the verifier fails
  - Using --hints-file hints.bin the prover fails and reports the same error message as the verifier in the previous case

The hints file just contains the four bytes: 10 0 0 0 or 5 0 0 0

kunxian-xia:
We need a tool to generate `hints.bin` automatically from the guest program.

docs/src/generating-and-verifying-proofs.md
```diff
@@ -0,0 +1,149 @@
...
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+ceno_rt = { git = "https://github.com/scroll-tech/ceno.git" }
+rkyv = { version = "0.8", default-features = false, features = [
+  "alloc",
+  "bytecheck",
+] }
+```
```
hero78119, line 66: I think we need to add `rust-toolchain.toml`

```
[toolchain]
channel = "nightly-2025-03-25"
targets = ["riscv32im-unknown-none-elf"]
# We need the sources for build-std.
components = ["rust-src"]
```

specify as nightly build as currently we rely on some features not in stable channel. Otherwise it will suffer build failed

docs/src/generating-and-verifying-proofs.md
```diff
@@ -0,0 +1,149 @@
...
+
+This will save the keys in a `keys` directory.
+
+#### 4.3. Generate a Proof
+
+Now, run the program and generate a proof. You can provide input via the `--stdin` flag.
+
+```bash
+cargo ceno prove --hints=5 --public-io=8 --out-proof proof.bin
```
hero78119, line 128: the root cause failed reported here https://github.com/scroll-tech/ceno/pull/1054#issuecomment-3264338619 is because in `prove` internally we generate new pk/vk pair each time, so we need to append "--out-vk vk.bin" prove command to force new generated vk are serialized thus the verify can use same key. 

For why pk/vk non-deterministic is under investigating

hero78119: In "prove" we exposed few arguments which will generate and invalidate pk/vk in key gen phase, e.g. `--field` allow change different field which will break the fixed commitment happened in key-gen phase due to different field.

I suggest to simplify, we can temporarily remove keygen phase and only rely on `prove`, `verify` for documentation

PR #1055, opened 2025-09-05
generate gkr-witness on gpu

hero78119:


ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -43,18 +44,11 @@ pub struct CpuTowerProver;
 
 impl CpuTowerProver {
     pub fn create_proof<'a, E: ExtensionField, PCS: PolynomialCommitmentScheme<E>>(
-        mut out_evals: Vec<Vec<Vec<E>>>,
```
hero78119, line 46: This out_evals seems introduce in target branch recently. This design is a bit weird because usually in prover we dont need to pass evaluation. Thus the better design of this api is left it handle externally, which is what master branch currently do

ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -531,138 +522,66 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<CpuBacke
...
 {
     #[allow(clippy::type_complexity)]
-    #[tracing::instrument(
-        skip_all,
-        name = "build_main_witness",
-        fields(profiling_3),
-        level = "trace"
-    )]
-    fn build_main_witness<'a, 'b>(
+    #[tracing::instrument(skip_all, name = "table_witness", fields(profiling_3), level = "trace")]
```
hero78119, line 546: just an heads up: we should use tracing to wrap whole function for profiling.

to profile a fragment within a function, we could just wrap it e.g. 
```
let xxxx_span = entered_span!("xxxxx");
// code to profiling
exit_span!(xxxx_span)
```

ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -531,138 +522,66 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<CpuBacke
...
+    fn table_witness<'a>(
```
hero78119, line 547: table witness generation are a special case for no gpu implementation. Thus right now cpu/gpu still got same implementation.

For now, gpu code are still copy-paste from cpu, because I try to avoid introduce "self" in trait function for `TemporarilyGPUProver` to impl because ultimately we would remove `TemporarilyGPUProver`

ceno_zkvm/src/scheme/gpu/mod.rs
```diff
@@ -914,6 +955,147 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> DeviceTransporter<Gp
...
+                cuda_hal
+                    .alloc_ext_elems_on_device(1 << num_vars)
+                    .map_err(|e| format!("Failed to allocate prod GPU buffer: {:?}", e))
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .unwrap();
+
+        cuda_hal
+            .witness_infer
+            .wit_infer_by_monomial_expr(
```
hero78119, line 1063: Here invoke kernel to handle all expression

ceno_zkvm/src/scheme/utils.rs
```diff
@@ -283,6 +297,212 @@ pub(crate) fn infer_tower_product_witness<E: ExtensionField>(
     wit_layers
 }
 
+pub fn build_main_witness<
```
hero78119, line 300: This flow share between cpu and gpu. The differentiation down to `layer_witness` level

ceno_zkvm/src/scheme/cpu/mod.rs
Velaciela: Let me remove `TemporaryGpuProver` after this PR, since it's no longer needed now

PR #987, opened 2025-07-21
feat: integrate GPU prover

Velaciela:
### Performence
- fibonacci $\text{fib}(2^{20})$: runs `12M` steps (warmup + rerun)
- machine(node-19): `AMD EPYC 7702 64-Core`, `1 x RTX 3090`
```
# CPU
RUST_LOG=debug cargo run --release --package ceno_zkvm --bin e2e -- --platform=ceno --hints=20 --public-io=3130 --profiling=2   examples/target/riscv32im-ceno-zkvm-elf/release/examples/fibonacci 2>&1 | tee "fib_perf_cpu.log"
```
```
# GPU
make enable-gpu
RUST_LOG=debug cargo run --release --features gpu --package ceno_zkvm --bin e2e -- --platform=ceno --hints=20 --public-io=3130 --profiling=2 --field=goldilocks  examples/target/riscv32im-ceno-zkvm-elf/release/examples/keccak_syscall 2>&1 | tee "fib_perf_gpu.log"
make disable-gpu
```
- **phase1: basefold commit & open**

| metric | CPU time | GPU time | speedup |
|------|----------|----------|--------|
| ZKVM_create_proof | 13.5s | 8.0s | 1.68x |
| batch commit to traces | 6.07s | 1.15s | 5.27x |
| transfer MLEs to host | 0 | 550ms | N/A |
| transfer pk to device | 0 | 4.53ms | N/A |
| main_proofs (no GPU yet) | 6.04s | 6.15s | 0.98x |
| pcs_opening | 1.42s | 168ms | 8.45x |

- **(WIP) phase2: main & tower prove**

| metric | CPU time | GPU time | speedup |
|------|----------|----------|--------|
| ZKVM_create_proof | 13.6s | 6.72s | 2.02x |
| batch commit to traces | 6.07s | 1.15s | 5.27x |
| transfer MLEs to host | 0 | 550ms (next: eliminate) | N/A |
| transfer pk to device | 0 | 4.53ms | N/A |
| main & tower prove | 6.04s | 4.85s | N/A |
| -- main (todo: update)  | - | - | - |
| -- tower  | 4.12s | 2.72s (next: optimize) | 1.51x |
| pcs_opening | 1.42s | 168ms | 8.45x |

### Current Status
- [x] Add GPU Prover Hal
- [x] Manually checked that `batch_commit` and `batch_open` produce consistent results between CPU and GPU provers
    - `cargo test --release test_single_add_instance_e2e -- --nocapture`
[...]

PR #1057, opened 2025-09-05
cosmetics of PR #987

hero78119:


ceno_zkvm/src/e2e.rs
```diff
@@ -744,15 +744,6 @@ pub fn run_e2e_with_checkpoint<
     // Run proof phase
     let transcript = Transcript::new(b"riscv");
     let start = std::time::Instant::now();
-    let zkvm_proof = prover
```
hero78119, line 747: This will be performance impact because it's equivalent to generate twice proof.

In `cargo-criterion` there is time reserved for warm up so we can skip this workaround

Velaciela (approved):
LGTM! Thanks~

PR #1065, opened 2025-09-09
public io size finalize in keygen time

### LINKED ISSUE #1066: finalized public io length in keygen phase

hero78119:
To close #1066

One of non-deterministic pk/vk came from PubIoTable fixed commitment length.
Previously in key-setup time there is no public io, thus PubIoTable addr is hardcode as 16 byte -> 2 word
In prove step once setup public io, then the PubIoTable addr will be dynamic extend to cover new public io, thus the fixed commitment are not identical, cause transcript challenge got different.

To fixed the root cause, we pass public io size in key setup time, and then in "prove" phase we check public io data <= public io size.

### DIFF
--- a/ceno_cli/src/commands/common_args/ceno.rs
+++ b/ceno_cli/src/commands/common_args/ceno.rs
@@ -341,12 +341,12 @@ fn run_elf_inner<
     let public_io = options
         .read_public_io()
         .context("failed to read public io")?;
-    let pubic_io_size = options.public_io_size;
+    let public_io_size = options.public_io_size;
     assert!(
-        public_io.len() <= pubic_io_size as usize / WORD_SIZE,
+        public_io.len() <= public_io_size as usize / WORD_SIZE,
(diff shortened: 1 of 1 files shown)


ceno_cli/src/commands/common_args/ceno.rs
```diff
@@ -337,26 +341,29 @@ fn run_elf_inner<
...
+    );
 
     let platform = if compilation_options.release {
         setup_platform(
             options.platform,
             &program,
             options.stack_size(),
             options.heap_size(),
-            pub_io_size,
+            pubic_io_size,
```
yczhangsjtu: Typo? `public_io`?

hero78119: yeah it's typo and nice catch, fixed in https://github.com/scroll-tech/ceno/pull/1065/commits/d27361cd3e3bd96b8c76b8b756768e38ba4917aa

PR #998, opened 2025-08-04
memory value to limb 2

hero78119:
To close #996 

```diff
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
| opcode_name   | num_instances | lookups | reads | witnesses | writes | 0_expr_deg | 0_expr_sumcheck_deg |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+
-| LB            | 0             | 14      | 4     | 28        | 4      | [1: 4]     | [2: 5]              |
+| LB            | 0             | 12      | 4     | 28        | 4      | [1: 4]     | [2: 5]              |
-| LBU           | 0             | 13      | 4     | 27        | 4      | [1: 4]     | [2: 4]              |
+| LBU           | 0             | 11      | 4     | 27        | 4      | [1: 4]     | [2: 4]              |
-| LH            | 0             | 12      | 4     | 25        | 4      | [1: 4]     | [2: 3]              |
+| LH            | 0             | 10      | 4     | 25        | 4      | [1: 4]     | [2: 3]              |
-| LHU           | 0             | 11      | 4     | 24        | 4      | [1: 4]     | [2: 2]              |
+| LHU           | 0             | 9       | 4     | 24        | 4      | [1: 4]     | [2: 2]              |
-| LW            | 0             | 11      | 4     | 22        | 4      | [1: 4]     | []                  |
+| LW            | 0             | 9       | 4     | 22        | 4      | [1: 4]     | []                  |
-| SB            | 0             | 15      | 4     | 29        | 4      | [1: 4]     | [2: 5]              |
+| SB            | 0             | 15      | 4     | 28        | 4      | [1: 4]     | [2: 4]              |
-| SH            | 0             | 11      | 4     | 24        | 4      | [1: 4]     | [2: 2]              |
+| SH            | 0             | 11      | 4     | 23        | 4      | [1: 4]     | [2: 1]              |
+---------------+---------------+---------+-------+-----------+--------+------------+---------------------+

```

### benchmark
e2e fibonacci benchmark shows no significant change due to fibonacci not heavily use memory read/write

| Benchmark                        | Median Time (s) | Median Change (%)                           |
|----------------------------------|------------------|----------------------------------------------|
| fibonacci_max_steps_1048576      | 2.4492           | -1.07% (Change within noise threshold)       |
| fibonacci_max_steps_2097152      | 4.5536           | -1.56% (Performance has improved)            |
[...]

PR #1073, opened 2025-09-25
fix heap value limb size mismatch

hero78119:
There is a bug from previous PR https://github.com/scroll-tech/ceno/pull/998/ where the `HeapTable` are missing of update memory value. `StackTable` and `HintTable` no such problem.

After #998 actually guest program with heap allocation will failed in e2e. However we only have `fibonacci` and `keccak` in e2e so actually the bug didn't capture.

Thus, this PR also add a heap allocation in e2e ci.

### DIFF
--- a/ceno_zkvm/src/scheme/mock_prover.rs
+++ b/ceno_zkvm/src/scheme/mock_prover.rs
@@ -1147,7 +1147,7 @@ Hints:
                             .into()
                         };
 
-                    for ((w_rlc_expr, annotation), (_, w_exprs)) in (cs
+                    for ((w_rlc_expr, annotation), _) in (cs
                         .w_expressions
                         .iter()
                         .chain(cs.w_table_expressions.iter().map(|expr| &expr.expr)))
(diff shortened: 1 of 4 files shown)


ceno_zkvm/src/scheme/mock_prover.rs
```diff
@@ -1173,35 +1173,6 @@ Hints:
                         let write_rlc_records =
                             filter_mle_by_selector_mle(write_rlc_records, w_selector.clone());
 
-                        if $ram_type == RAMType::GlobalState {
```
hero78119, line 1176: here is just for better debugging message

Before fix, the debug message during read/write mismatch is 
```
ERROR ceno_zkvm::scheme::mock_prover: riscv_opcode/SW/write_memory/write_record at row 42 (pc=8001a00,ts=832) not found in Memory reads
ERROR ceno_zkvm::scheme::mock_prover: riscv_opcode/SW/write_memory/write_record at row 48 (pc=8001a00,ts=1020) not found in Memory reads
ERROR ceno_zkvm::scheme::mock_prover: riscv_opcode/SW/write_memory/write_record at row 54 (pc=8001a00,ts=1208) not found in Memory reads
```
However, the pc showed here is represent the **NEXT** pc because we generate global state message from state "write" which pc has been bump +4.

What we need there is current pc which fit the debug nature

The right way is collect this information from global state "read"

kunxian-xia (approved):

PR #1069, opened 2025-09-22
chore: upgrade toolchain to nightly-2025-08-18

lightsing:
- upgrade toolchain to nightly-2025-08-18
- upgrade edition to 2024 (some previous used nightly features are stabilized)

PR #1076, opened 2025-09-26
feat: revm precompile crypto skeleton

lightsing:
depends on #1069 (requires rust 1.88)


lightsing:
closed prefer to use https://github.com/alloy-rs/alloy/blob/d3410e6046839a07f0db1116678a4484512afa0b/crates/consensus/src/crypto.rs#L113

PR #1078, opened 2025-09-30
feat: add secp256k1 ecrecover precompile

lightsing:
patch crate: https://github.com/scroll-tech/elliptic-curves/tree/ceno/k256-13.4

test_secp256k1_ecrecover, runs 5 ecrecover:
- with precompile: 3845214 steps
- without: 65928672 steps

### DIFF
--- a/ceno_zkvm/src/scheme/gpu/mod.rs
+++ b/ceno_zkvm/src/scheme/gpu/mod.rs
@@ -45,6 +45,7 @@ use gkr_iop::gpu::gpu_prover::*;
 
 pub struct GpuTowerProver;
 
+use crate::scheme::constants::NUM_FANIN;
 use gkr_iop::gpu::{ArcMultilinearExtensionGpu, MultilinearExtensionGpu};
 
 // Extract out_evals from GPU-built tower witnesses
@@ -59,7 +60,7 @@ fn extract_out_evals_from_gpu_towers<E: ff_ext::ExtensionField>(
     let mut w_out_evals = Vec::new();
     for (i, gpu_spec) in prod_gpu.iter().enumerate() {
         let first_layer_evals: Vec<E> = gpu_spec
-            .get_final_evals(0)
+            .get_output_evals()
             .expect("Failed to extract final evals from GPU product tower");
 
         // Product tower first layer should have 2 MLEs
@@ -81,7 +82,7 @@ fn extract_out_evals_from_gpu_towers<E: ff_ext::ExtensionField>(
     let mut lk_out_evals = Vec::new();
     for gpu_spec in logup_gpu.iter() {
         let first_layer_evals: Vec<E> = gpu_spec
-            .get_final_evals(0)
+            .get_output_evals()
             .expect("Failed to extract final evals from GPU logup tower");
 
         // Logup tower first layer should have 4 MLEs
@@ -481,7 +482,7 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TowerProver<GpuBacke
         let span = entered_span!("prove_tower_relation", profiling_2 = true);
         let (point_gl, proof_gpu) = cuda_hal
             .tower
-            .create_proof(&input, basic_tr)
+            .create_proof(&cuda_hal, &input, NUM_FANIN, basic_tr)
             .expect("gpu tower create_proof failed");
         exit_span!(span);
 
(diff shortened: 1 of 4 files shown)


guest_libs/crypto/src/secp256k1.rs
```diff
@@ -1,11 +1,172 @@
...
+    let r_point = {
+        let mut buf = Aligned64([0u8; 64]);
+        buf.0[..32].copy_from_slice(&r_bytes);
+
+        // SAFETY:
+        // [x] The input array should be 64 bytes long, with the first 32 bytes containing the X coordinate in
+        //   big-endian format.
+        // [x] The caller must ensure that `point` is valid pointer to data that is aligned along a four byte
+        // boundary.
+        ceno_rt::syscalls::syscall_secp256k1_decompress(&mut buf.0, recovery_id.is_y_odd());
```
hero78119: Hi, we do not have decompress function yet, does it work if we make it natively guest program? such as comment above

```
let R = AffinePoint::decompress(&r_bytes, u8::from(recovery_id.is_y_odd()).into());
```

hero78119: As offline discussion, it will cause huge cycle different. likely
`2345586 vs 4033796`
Thus let make decompress as one of precompile

guest_libs/crypto/src/secp256k1.rs
```diff
@@ -1,11 +1,172 @@
...
+    // ProjectivePoint::lincomb(&ProjectivePoint::GENERATOR, &u1, &r_point, &u2)
+    // Equivalent to: G * u1 + R * u2
+    let pk_words = match (u1 == Scalar::ZERO, u2 == Scalar::ZERO) {
+        (true, true) => return Err(Error::new()),
+        (true, false) => secp256k1_mul(&r_point, u2),
+        (false, true) => secp256k1_mul(&AffinePoint::GENERATOR, u1),
+        (false, false) => {
+            let mut p1 = secp256k1_mul(&AffinePoint::GENERATOR, u1);
+            let p2 = secp256k1_mul(&r_point, u2);
+            syscall_secp256k1_add(&mut p1, &p2);
```
hero78119: Hi @dreamATD could you helps to review the usage here, e.g.
do we need to check p1 != p2 before calling `syscall_secp256k1_add` ? 

Besides, question from @lightsing: does it work for p1 or p2 to be infinite point? or we should handle it before syscall?

guest_libs/crypto/src/secp256k1.rs
```diff
@@ -1,11 +1,172 @@
...
+            p1
+        }
+    };
+
+    // FIXME: do we really need to verify the signature again here?
+    // Original:
+    // let vk = VerifyingKey::from_affine(pk.to_affine())?;
```
hero78119: per offline discussion, this operation seems to be the key to bind relation of (prehash, signature, recovery_id) so let retain original functionality

lightsing: resolved this.

guest_libs/crypto/src/secp256k1.rs
lightsing: ping here @dreamATD

ceno_rt/Cargo.toml
```diff
@@ -11,4 +11,5 @@ version = "0.1.0"
 
 [dependencies]
 getrandom = { version = "0.2.15", features = ["custom"], default-features = false }
```
hero78119, line 13: noticed there are 2 version `getrandom` here, are we able to remove current `getrandom` and switch to `getrandom_v3` ?

guest_libs/crypto-primitives/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
hero78119, line 1: Would be nice to add a readme to brief some primitives along with a credit of its source

e.g. [README.md](https://github.com/scroll-tech/gkr-backend/blob/main/crates/curves/README.md)

ceno_rt/Cargo.toml
lightsing: I think we should keep both to be able support them concurrently.

guest_libs/crypto-primitives/Cargo.toml
lightsing: done

PR #1090, opened 2025-10-19
Fix two TODOs

c410-f3r:
Also swaps `FIXME` with `TODO` for consistency

### DIFF
--- a/ceno_emul/src/syscalls/sha256.rs
+++ b/ceno_emul/src/syscalls/sha256.rs
@@ -38,7 +38,6 @@ pub fn sha_extend(w: &mut [u32]) {
             .wrapping_add(s0)
             .wrapping_add(w[i - 7])
             .wrapping_add(s1);
-        // TODO: why doesn't sp1 use wrapping_add?
     }
 }
 
--- a/ceno_rt/src/lib.rs
+++ b/ceno_rt/src/lib.rs
@@ -6,7 +6,7 @@ use getrandom::{Error, register_custom_getrandom};
 use core::arch::{asm, global_asm};
 use std::{
     alloc::{Layout, alloc_zeroed},
-    ptr::null,
+    ptr,
 };
 
 #[cfg(target_arch = "riscv32")]
@@ -37,7 +37,7 @@ pub extern "C" fn sys_alloc_words(nwords: usize) -> *mut u32 {
 #[unsafe(no_mangle)]
 #[linkage = "weak"]
 pub extern "C" fn sys_getenv(_name: *const u8) -> *const u8 {
-    null()
+    ptr::null()
 }
 
 /// Generates random bytes.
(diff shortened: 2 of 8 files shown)


ceno_emul/src/syscalls/sha256.rs
```diff
@@ -38,7 +38,6 @@ pub fn sha_extend(w: &mut [u32]) {
             .wrapping_add(s0)
             .wrapping_add(w[i - 7])
             .wrapping_add(s1);
-        // TODO: why doesn't sp1 use wrapping_add?
```
c410-f3r, line 41: Succint uses manual `+` operators that wrap in release mode and panic in debug. This is a hashing function so wrapping is the right intent IMO.

ceno_rt/src/lib.rs
```diff
@@ -49,26 +50,40 @@ pub extern "C" fn sys_getenv(_name: *const u8) -> *const u8 {
...
+            *recv_buf.add(idx) = a;
+            *recv_buf.add(idx.wrapping_add(1)) = b;
+            *recv_buf.add(idx.wrapping_add(2)) = c;
+            *recv_buf.add(idx.wrapping_add(3)) = d;
         }
+        idx = idx.wrapping_add(4);
     }
-    // TODO(Matthias): this is a bit inefficient,
-    // we could fill whole u32 words at a time.
-    // But it's just for testing.
```
c410-f3r, line 64: Fills 4 bytes at once. Shouldn't matter because the function is a mock but regardless 🤷

ceno_rt/src/lib.rs
```diff
@@ -49,26 +50,40 @@ pub extern "C" fn sys_getenv(_name: *const u8) -> *const u8 {
...
-    unsafe fn step() -> u32 {
-        static mut X: u32 = 0xae569764;
-        // We are stealing Borland Delphi's random number generator.
-        // The random numbers here are only good enough to make eg
-        // HashMap work.
+    // We are stealing Borland Delphi's random number generator.
+    // The random numbers here are only good enough to make eg
+    // HashMap work.
+    fn step() -> u32 {
+        static X: AtomicU32 = AtomicU32::new(0xae569764);
```
c410-f3r: > This crate provides the runtime for programs running on the Ceno VM.

IIUC it is OK to use atomics in this scenario. If so, `static mut` is unnecessary.

ceno_rt/src/lib.rs
```diff
@@ -49,26 +50,37 @@ pub extern "C" fn sys_getenv(_name: *const u8) -> *const u8 {
 #[unsafe(no_mangle)]
 #[linkage = "weak"]
 pub unsafe extern "C" fn sys_rand(recv_buf: *mut u8, words: usize) {
-    unsafe fn step() -> u32 {
```
hero78119, line 52: Do you have any insights on:
- the difference in generated assembly code?
- whether using AtomicU32 introduces any additional sys-calls?

gkr_iop/src/gkr/layer/cpu/mod.rs
```diff
@@ -302,7 +302,7 @@ pub(crate) fn prove_rotation<E: ExtensionField, PCS: PolynomialCommitmentScheme<
     )
     .collect_vec();
     exit_span!(span);
-    // TODO FIXME: we pick a random point from output point, does it sound?
+    // TODO TODO: we pick a random point from output point, does it sound?
```
hero78119: nitpick: some leftover due to replacement I guess

ceno_rt/src/lib.rs
c410-f3r: https://godbolt.org/z/TP3TsE91T shows the usage of the `lock` instruction, which is a native CPU feature (not a syscall). Optimized builds are much smaller -> https://godbolt.org/z/ojY14joaY

 The current code has a similar optimized structure -> https://godbolt.org/z/3TEoa3PbP

`AtomicU32` is more about removing one layer of `unsafe` but  I can revert back to `static mut` if you want.

gkr_iop/src/gkr/layer/cpu/mod.rs
c410-f3r: Ops... Thanks

ceno_rt/src/lib.rs
hero78119: Since we are working on riscv `zk`vm so one of the concern is the emulator are limited to support native cpu feature in particular for those which is more bind to hardware, such as `lock`, as we need to "prove" lock behaviour with a dedicated zk circuit which is hard to implemented. So I would suggest retain `static mut` behaviours as it doesn't introduce instruction outside of riv32im

c410-f3r: Sure, no problem! Thanks for the review

c410-f3r:
The code should be in good shape now.

I am trying to get more hands-on experience with ZK so can you indicate something feasible/solvable for someone new? Looking at the issues alone didn't provide much context.

ceno_rt/src/lib.rs
hero78119: thanks for the fix and clean up! previous logic is a bit confusing and now the new logic much clear :)

hero78119 (approved):
Sorry to get back lately, overall LGTM and thanks for contribution ❤️

hero78119:
> The code should be in good shape now.
> 
> I am trying to get more hands-on experience with ZK so can you indicate something feasible/solvable for someone new? Looking at the issues alone didn't provide much context.

Thanks for the interests. Yeah will schedule a house-keeping then clean up all out-of-date issues sooner, after resolving this major milestone https://github.com/scroll-tech/ceno/issues/1095 
Will categorize issues with clear github tag and welcome to pick up any issue you interests :)

PR #1102, opened 2025-10-30
Fix integration bugs

kunxian-xia:


ceno_zkvm/src/scheme/prover.rs
```diff
@@ -144,22 +155,26 @@ impl<
                     })
                 })
             {
-                num_instances.push((
-                    index,
-                    num_instance >> vk.get_cs().rotation_vars().unwrap_or(0),
-                ));
+                let inst = if vk.get_cs().has_ecc_ops() {
```
hero78119: if / else condition can be clean up and unified to

```rust
// even non ecc_ops circuit can go to same flow
let inst = {
let num_instance = num_instance >> vk.get_cs().rotation_vars().unwrap_or(0);
let num_global_reads = witnesses.num_global_reads;
let num_local_reads = num_instance - num_global_reads;
let num_local_writes = num_global_reads;
ChipInstances::new(num_local_reads, num_local_writes, num_instance)
}
```

Then in verifier, for non-ecc circuit we check num_local_reads, num_local_writes should be 0

PR #1061, opened 2025-09-08
Feat: septic curve based continuation

kunxian-xia:
### Sub tasks
- [ ] septic curve points accumulation
   - [ ] $\log N$ layer witnesses generation
   - [ ] $\log N$ layer sumchecks batched in TowerProver
- [ ] `Global` chip
- [ ] `MemoryLocal` chip

PR #1106, opened 2025-10-31
#1061 cleanup 

kunxian-xia:
#1061 Cleanup

- [x] clippy fixes
- [x] remove old unused routines for generating witness for the ecc summation tower tree. Note this approach is abandoned. We prefer to pack $log_2(N)$ layers in one layer using the relation `p[1, b] = ec_add(p[0,b], p[1,b])` (first proposed in Quark paper.

### DIFF
--- a/ceno_zkvm/src/scheme/cpu/mod.rs
+++ b/ceno_zkvm/src/scheme/cpu/mod.rs
@@ -1235,10 +1235,8 @@ mod tests {
 
         assert_eq!(quark_proof.sum, final_sum);
         let mut transcript = BasicTranscript::new(b"test");
-        let verifier = EccVerifier::new();
         assert!(
-            verifier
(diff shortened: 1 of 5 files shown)


ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -55,6 +55,7 @@ pub type TowerRelationOutput<E> = (
 
 // accumulate N=2^n EC points into one EC point using affine coordinates
 // in one layer which borrows ideas from the [Quark paper](https://eprint.iacr.org/2020/1275.pdf)
+#[derive(Default)]
```
hero78119: we can remove 
- `&self` from create_ecc_proof, 
- new() as it's do nothing currently

so the clippy will not complain about this?

PR #1094, opened 2025-10-27
ceno-cli conntinuation proofs e2e

hero78119:
Build on top of #1092 

### change scopes
- [x] verify api for continuation proofs
- [x] exclude init tables on shard > 0
- [x] ceno-cli with multiple proofs
- [x] separate fixed commitment into 2 set, one for first shard, the other for non-first shards 
- [x] multi-shards ci integration test 

### design rationales
`prover_id` and `num_provers` are exposed as CLI arguments to specify the number of physical provers in a cluster, each  mark with a prover_id. The overall trace data is divided into `shards`, which are distributed evenly among the provers. The number of shards are in general agnostic to number of provers. Each prover is assigned `n` shard where n can be even empty

Shard distribution follows a balanced allocation strategy — for example, if there are 10 shards and 3 provers, the shard counts will be distributed as 4, 3, and 3, ensuring an even workload across all provers.

PR #1107, opened 2025-10-31
api to extract global sum

hero78119:
preparation of #1094


hero78119:
Will be included in #1094

PR #1108, opened 2025-10-31
Address review comments in #1061

kunxian-xia:
Items

- [x] enforce `is_ram_register` is boolean.
- [ ] if it's global read, enforce `shard < cur_shard`.
- [ ] if it's global write, then enforce `shard = cur_shard` where `cur_shard` is fetched from public values.


ceno_zkvm/src/instructions/global.rs
```diff
@@ -228,7 +227,31 @@ impl<E: ExtensionField> GlobalConfig<E> {
...
             local_clk.expr(),
         )?;
-        // TODO: enforce shard = shard_id in the public values
+
+        // if it's global write => shard == cur_shard
+        let cur_shard = cb.query_shard_id()?;
+        cb.condition_require_zero(
+            || "global_write = true => shard = instance.shard",
+            is_global_write.expr(),
+            shard.expr() - Expression::Instance(cur_shard),
```
kunxian-xia, line 236: main sumcheck does not support public inputs now.

ceno_zkvm/src/e2e.rs
```diff
@@ -274,6 +274,11 @@ impl<'a> ShardContext<'a> {
         }
     }
 
+    #[inline(always)]
+    pub fn cur_shard(&self) -> usize {
```
hero78119, line 278: nitpick: `cur_shard_id()` or `shard_id()` better

ceno_zkvm/src/instructions/global.rs
hero78119: please try this way

```rust
let cur_shard = cb.query_shard_id();
cur_shard.expr() // this return `Expression::InstanceScalar()` which make thing work in sumcheck
```

PR #1109, opened 2025-11-01
bug fix of global chip

hero78119:
fix few issue reported in global chip integration test

### DIFF
--- a/ceno_zkvm/src/instructions/global.rs
+++ b/ceno_zkvm/src/instructions/global.rs
@@ -52,22 +52,35 @@ pub struct GlobalRecord {
     pub shard: u64,
     pub local_clk: u64,
     pub global_clk: u64,
-    pub is_write: bool,
+    pub is_to_write_set: bool,
 }
 
 impl From<(&WordAddr, &RAMRecord, bool)> for GlobalRecord {
-    fn from((vma, record, is_write): (&WordAddr, &RAMRecord, bool)) -> Self {
+    fn from((vma, record, is_to_write_set): (&WordAddr, &RAMRecord, bool)) -> Self {
         let addr = match record.ram_type {
             RAMType::Register => record.id as u32,
             RAMType::Memory => (*vma).into(),
             _ => unreachable!(),
         };
-        let value = record.prev_value.map_or(record.value, |v| v);
-        let (shard, local_clk, global_clk) = if is_write {
-            (record.shard_id, record.shard_cycle, record.cycle)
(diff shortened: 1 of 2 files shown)


ceno_zkvm/src/instructions/global.rs
```diff
@@ -52,22 +52,35 @@ pub struct GlobalRecord {
...
 
 impl From<(&WordAddr, &RAMRecord, bool)> for GlobalRecord {
-    fn from((vma, record, is_write): (&WordAddr, &RAMRecord, bool)) -> Self {
+    fn from((vma, record, is_to_write_set): (&WordAddr, &RAMRecord, bool)) -> Self {
         let addr = match record.ram_type {
             RAMType::Register => record.id as u32,
             RAMType::Memory => (*vma).into(),
             _ => unreachable!(),
         };
-        let value = record.prev_value.map_or(record.value, |v| v);
```
hero78119, line 65: for global write (i.e. local read), we need to take `value` which append to write set. In this case, it should be `record.value`, regardless of the `record.prev_value` exist or not.

In particular, `record.prev_value` determine by high level read/write operation. Here is low level which only talking about read/write set in omc

ceno_zkvm/src/instructions/global.rs
```diff
@@ -211,8 +224,8 @@ impl<E: ExtensionField> GlobalConfig<E> {
         input.extend(repeat_n(E::BaseField::ZERO.expr(), 16 - input.len()));
 
         let mut record = vec![];
-        record.push(addr.expr());
         record.push(ram_type.clone());
+        record.push(addr.expr());
```
hero78119, line 228: there is another bug, for the record order mismatch

ceno_zkvm/src/structs.rs
```diff
@@ -434,28 +434,30 @@ impl<E: ExtensionField> ZKVMWitnesses<E> {
...
         let global_input = shard_ctx
-            .read_records()
+            .write_records()
             .par_iter()
             .flat_map_iter(|records| {
+                // global write -> local reads
                 records.iter().map(|(vma, record)| {
-                    let global_read: GlobalRecord = (vma, record, false).into();
-                    let ec_point: GlobalPoint<E> = global_read.to_ec_point(&perm);
+                    let global_write: GlobalRecord = (vma, record, true).into();
```
hero78119, line 442: The assignment logic in global chip assume `local read` come before `local write`
In other words, global write need to be first, follow by global read

kunxian-xia (approved):

PR #1117, opened 2025-11-05
Fix: `ShardRamCircuit` runs too slow

kunxian-xia:
# Issue

There is recurring pattern in Poseidon2 gadget which incurs a lot of monomial terms. I.e. 
1. `state'[i] = \sum_j M[i][j] * state[j]`;
2. `state''[i] = state'[i]^3` => this will create a lot of terms due to the nature of distributive law.

## Walk around

The temporary solution to this issue is to allocate polynomials to store the value after each linear layer operation (1).

| before/after| num_polys | num monomial terms of zerocheck |
|------------|------------|------------------------------------|
| before        |  329            | 7139                     |
| after           |  374    (16*2 + 13 ⬆️)      | 957                       |

### DIFF
--- a/gkr_iop/src/gkr/layer/zerocheck_layer.rs
+++ b/gkr_iop/src/gkr/layer/zerocheck_layer.rs
@@ -165,14 +165,13 @@ impl<E: ExtensionField> ZerocheckLayer<E> for Layer<E> {
             self.n_fixed as WitnessId,
             self.n_instance,
         );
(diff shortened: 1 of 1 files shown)


hero78119 (commented):
LGTM, just 2 suggest

1. add few word in PR description, e.g. before/after change how much monimial term decrease
2. use `tracing::trace`

hero78119 (approved):
LGTM!

PR #1124, opened 2025-11-09
Recursion and aggregation of base Ceno sumcheck proofs

darth-cy:
- Binary tree aggregation of base VM Ceno proofs.
- DSL-coded ceno verifier recursion program.
- Native sumcheck operation in recursive OpenVM environment.

Recursion e2e Test Sample Command:
```
cargo run --release --package ceno_recursion --bin e2e -- --platform=ceno --min-cycle-per-shard=10 --max-cycle-per-shard=1600 examples/target/riscv32im-ceno-zkvm-elf/release/examples/keccak_syscall
```


hero78119 (approved):
👍 LGTM and let make bug fix as followup PRs

PR #1130, opened 2025-11-12
Feat: bound ec sum in ecc quark prover

kunxian-xia:
## Issue

Our current quark-based ec summation protocol only ensures the sum of `N` points encoded in 14 multivariate polynomials is stored at `(xs[1,...,1,0], ys[1,...,1,0])`. 

But we did **not** export it.

This implies the `sum` field in the `EccQuarkProof` can be manipulated arbitrarily.

## Summary
We augment the current quark-style zerocheck by requiring 
```
1. 0 = \sum_b sel_export[b] * (x[1,b] - final_x)
2. 0 = \sum_b sel_export[b] * (y[1,b] - final_y)
```

### DIFF
--- a/ceno_recursion/src/zkvm_verifier/verifier.rs
+++ b/ceno_recursion/src/zkvm_verifier/verifier.rs
@@ -1719,7 +1719,7 @@ pub fn verify_ecc_proof<C: Config>(
     let alpha_pows = gen_alpha_pows(
         builder,
         challenger,
-        Usize::from(SEPTIC_EXTENSION_DEGREE * 3 + SEPTIC_EXTENSION_DEGREE * 2),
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -253,19 +279,22 @@ impl CpuEccProver {
             let y0 = filter_bj(&ys, 0);
             let x1 = filter_bj(&xs, 1);
             let y1 = filter_bj(&ys, 1);
+            let sel_export = eq_eval(&out_rt, &lsi_on_hypercube) * eq_eval(&rt, &lsi_on_hypercube);
```
hero78119, line 282: as `sel_export` value is check here, should we also add sel_add, sel_bypass? that make thing symmetric

PR #1134, opened 2025-11-16
only transport proving key once in multi-shard

hero78119:


ceno_zkvm/src/structs.rs
```diff
@@ -687,6 +687,10 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> ZKVMProvingKey<E, PC
     pub(crate) fn set_program_entry_pc(&mut self, entry_pc: u32) {
         self.entry_pc = entry_pc;
     }
+
+    pub fn is_fixed_commitment(&self) -> bool {
```
kunxian-xia: nitpick: maybe name like `has_fixed_commitment` suits better here : )

hero78119: 👍 fixed in https://github.com/scroll-tech/ceno/pull/1134/commits/53ead5c3cc2cf63a772635f29a6bf7079537333e

PR #1135, opened 2025-11-17
shard load balance via "cell" cost per instruction/precompile 

hero78119:
### design rationale

To address GPU memory consumption per shard, this PR introduce per opcode memory overhead as `cell`, where the term are borrow from halo2 [Cell](https://github.com/zcash/halo2/blob/2308caf68c48c02468b66cfc452dad54e355e32f/halo2_proofs/src/circuit.rs#L89-L96). 

In short, a column + rotation => a cell. In Ceno currently implementation, there is rotation on most of circuit. The only exception is keccak which got rotation 32. 

Define new shard cut criteria by max-cell-per-shard. Assume we try to fill 4090 with 24GB VRAM, and one cell mean a base field element which for now is 4 bytes. Giving 16GB reserve for row-major-matrix, then then estimated max cell to fill 16GB is  `2^30 * 16 / 4 / 2` = `2147483648`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,6 +46,10 @@ jobs:
         run: |
           cargo make --version || cargo install cargo-make
       - name: run test
+        env:
+          RUSTFLAGS: "" # cleanup RUSTFLAGS and avoid target-cpu=native
         run: cargo make tests
       - name: run test + goldilocks
+        env:
+          RUSTFLAGS: "" # cleanup RUSTFLAGS and avoid target-cpu=native
         run: cargo make tests_goldilock


ceno_zkvm/src/e2e.rs
```diff
@@ -416,12 +318,11 @@ impl<'a> ShardContext<'a> {
...
 
+    /// Extract shard_id which produce this record
+    /// NOTE prev_shard_cycle_range[0] should be 0 otherwise it will panic with subtract-overflow
     #[inline(always)]
     pub fn extract_shard_id(&self, cycle: Cycle) -> usize {
-        let subcycle_per_insn = Tracer::SUBCYCLES_PER_INSN;
-        let per_shard_cycles =
-            (self.expected_inst_per_shard as u64).saturating_mul(subcycle_per_insn);
-        ((cycle.saturating_sub(subcycle_per_insn)) / per_shard_cycles) as usize
+        self.prev_shard_cycle_range.partition_point(|&t| t <= cycle) - 1
```
hero78119, line 316: binary search to found shard_id cover this cycle

ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -144,92 +144,163 @@ pub struct Rv32imConfig<E: ExtensionField> {
...
+    // record InsnKind -> cells
+    pub inst_cells_map: Vec<u64>,
+    // record opcode name -> cells
+    // serve ecall/table for no InsnKind
+    pub ecall_table_cells_map: HashMap<String, u64>,
 }
 
 impl<E: ExtensionField> Rv32imConfig<E> {
     pub fn construct_circuits(cs: &mut ZKVMConstraintSystem<E>) -> Self {
+        let mut inst_cells_map = vec![0; InsnKind::COUNT];
```
hero78119, line 160: cell of each opcode are recording during circuit registration.
This avoid per opcode/table trait interface massive change

ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -144,92 +144,163 @@ pub struct Rv32imConfig<E: ExtensionField> {
...
+        let mut ecall_table_cells_map = HashMap::new();
```
hero78119: We use slower name + HashMap to further differentiate, as all ecall use single `InsnKind::Ecall`

ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -763,3 +836,50 @@ impl<E: ExtensionField> DummyExtraConfig<E> {
         Ok(())
     }
 }
+
+impl<E: ExtensionField> StepCellExtractor for &Rv32imConfig<E> {
+    #[inline(always)]
+    fn extract_cells(&self, record: &StepRecord) -> u64 {
+        let insn_kind = record.insn.kind;
+        if !matches!(insn_kind, InsnKind::ECALL) {
```
hero78119, line 848: Opcode match will be early return with speedy vector index lookup

ceno_zkvm/src/e2e.rs
```diff
@@ -620,6 +512,107 @@ impl<'a> ShardContext<'a> {
...
+            prev_shard_cycle_range: vec![0],
+        }
+    }
+
+    /// return next shard size and shard_ctx
+    /// panic if empty match
+    pub fn position_next_shard<'a>(
+        &mut self,
+        steps: &[StepRecord],
+        step_cell_extractor: impl StepCellExtractor,
```
hero78119, line 559: `step_cell_extractor` design make rv32im logic isolated from e2e so the overall flow is very general and not bind to rv32im logic

ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -144,92 +144,189 @@ pub struct Rv32imConfig<E: ExtensionField> {
...
-            cs.register_table_circuit::<DynamicRangeTableCircuit<E, DYNAMIC_RANGE_MAX_BITS>>();
-        let double_u8_range_config = cs.register_table_circuit::<DoubleU8TableCircuit<E>>();
-        let and_table_config = cs.register_table_circuit::<AndTableCircuit<E>>();
-        let or_table_config = cs.register_table_circuit::<OrTableCircuit<E>>();
-        let xor_table_config = cs.register_table_circuit::<XorTableCircuit<E>>();
-        let ltu_config = cs.register_table_circuit::<LtuTableCircuit<E>>();
+        macro_rules! register_table_circuit {
+            ($instruction:ty, $ecall_table_cells_map:ident) => {{
+                let config = cs.register_table_circuit::<$instruction>();
+
```
kunxian-xia: No need to extract info from table circuits.

ceno_zkvm/src/instructions/riscv/rv32im.rs
```diff
@@ -763,3 +862,50 @@ impl<E: ExtensionField> DummyExtraConfig<E> {
...
+    fn extract_cells(&self, record: &StepRecord) -> u64 {
+        let insn_kind = record.insn.kind;
+        if !matches!(insn_kind, InsnKind::ECALL) {
+            // quick match for opcode and return
+            return self.inst_cells_map[insn_kind as usize];
+        }
+        // deal with ecall logic
+        match insn_kind {
+            // ecall / halt
+            InsnKind::ECALL if record.rs1().unwrap().value == Platform::ecall_halt() => *self
```
kunxian-xia: Since this match always have `insn_kind = Ecall`, we can change this to `match record.rs1().unwrap().value`.

kunxian-xia:
This should be considered as sharding strategy tailored to GPU prover, maybe we can add more custom sharding strategies in the future.

ceno_zkvm/src/instructions/riscv/rv32im.rs
hero78119: yeah I copy-paste from previous matching logic. Fixed in https://github.com/scroll-tech/ceno/pull/1135/commits/0168fe7eae1d4787adbd8f7beda26eff2452bc31

ceno_zkvm/src/instructions/riscv/rv32im.rs
hero78119: agree it make less confused. remove this https://github.com/scroll-tech/ceno/pull/1135/commits/0168fe7eae1d4787adbd8f7beda26eff2452bc31 and rename to `ecall_cells_map` without table terms

PR #1143, opened 2025-11-20
Evaluation of Additional `SelectorType` Using `SelectorContext`

darth-cy:
- Complete evaluative expressions found in ecc and gkr verification.

Cycle count example:
- `keccak_syscall`
   - shard_1: 2173782
   - shard_2: 2247447
   - shard_3: 689641


kunxian-xia (approved):
We will review it later.

PR #1144, opened 2025-11-20
fix keccak failed with 1 instance

hero78119:
Keccak (with 1 row span into rotation 32) got some legacy bug of padding. 
In short, before fixed, rmm padding logic is `next_pow2(instance * rotation)`, but the correct logic should be `next_pow2(instance) * rotation`

To reproduce the bug, set ITERATIONS = 1

https://github.com/scroll-tech/ceno/blob/81e45aad132a5938485db405e46bb0b6173e08c2/examples/examples/keccak_syscall.rs#L8

Previous bug is not always trigger because it depends on `next_pow2(instance * rotation)` == `next_pow2(instance) * rotation` or not

This PR also encaptulate rotation logic in rmm, so the prover flow got less confusion in order to deal with rotation logic

### DIFF
--- a/gkr_iop/src/selector.rs
+++ b/gkr_iop/src/selector.rs
@@ -2,6 +2,7 @@ use std::iter::repeat_n;
 
 use rayon::iter::IndexedParallelIterator;
 
+use crate::{gkr::booleanhypercube::CYCLIC_POW2_5, utils::eq_eval_less_or_equal_than};
 use ff_ext::ExtensionField;
 use multilinear_extensions::{
     Expression, WitnessId,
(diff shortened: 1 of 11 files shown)


gkr_iop/src/selector.rs
```diff
@@ -154,7 +154,10 @@ impl<E: ExtensionField> SelectorType<E> {
             }
             // compute true and false mle eq(1; b[..5]) * sel(y; b[5..]), and eq(1; b[..5]) * (eq() - sel(y; b[5..]))
             SelectorType::OrderedSparse32 { indices, .. } => {
-                assert_eq!(out_point.len(), ceil_log2(ctx.num_instances) + 5);
```
hero78119, line 157: to deal with num_instance = 1, we need to pad to 2, then get its log2 to derive num_var = 1

PR #1147, opened 2025-11-21
[log] tracing log trace records & dependency refactor

hero78119:
### change log
- [x] tracing debug of all records
- [x] refactor `ceno_crypto_primitives` and  `ceno_syscall` to https://github.com/scroll-tech/ceno-patch
   - this make https://github.com/scroll-tech/elliptic-curves not circular dependency back with `ceno` due to `ceno_crypto_primitives`
- [x] add some example

### DIFF
--- a/examples/examples/secp256k1.rs
+++ b/examples/examples/secp256k1.rs
@@ -1,15 +1,15 @@
-
 extern crate ceno_rt;
-
-use k256::{ProjectivePoint, Scalar};
-use k256::elliptic_curve::Group;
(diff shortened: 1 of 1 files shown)


hero78119:
Todo: to complete this PR
- [ ] need to update ceno dependency in https://github.com/scroll-tech/elliptic-curves/tree/ceno/k256-13.4
- [ ] then update dependency here https://github.com/scroll-tech/ceno/blob/075d900d655311f40992f4331da8f749521f4c14/guest_libs/crypto/Cargo.toml#L16

PR #1151, opened 2025-11-24
debug improvement: add phantom syscall and log pc cycle

hero78119:


ceno_zkvm/src/instructions/riscv/dummy/dummy_ecall.rs
```diff
@@ -127,6 +137,6 @@ pub struct LargeEcallConfig<E: ExtensionField> {
 
     reg_writes: Vec<(UInt<E>, WriteRD<E>)>,
 
-    start_addr: WitIn,
+    start_addr: Option<WitIn>,
```
hero78119, line 140: Some dummy ecall could have 0 mem ops. So start_addr could be optional

ceno_emul/src/syscalls/phantom/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+    const CODE: u32 = ceno_syscall::PHANTOM_LOG_PC_CYCLE;
+}
+
+pub fn log_pc_cycle(vm: &VMState) -> SyscallEffects {
+    let lable_ptr = vm.peek_register(Platform::reg_arg0());
+    let lable_len = vm.peek_register(Platform::reg_arg1());
+
+    // Read the argument `state_ptr`.
+    let reg_ops = vec![
+        WriteOp::new_register_op(
```
kunxian-xia, line 22: why not `ReadOp`?

ceno_emul/src/syscalls/phantom/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+            Platform::reg_arg0(),
+            Change::new(lable_ptr, lable_ptr),
+            0, // Cycle set later in finalize().
+        ),
+        WriteOp::new_register_op(
```
kunxian-xia, line 27: Why not `ReadOp`?

ceno_emul/src/syscalls/phantom/mod.rs
hero78119: https://github.com/scroll-tech/ceno/blob/master/ceno_emul/src/syscalls.rs#L56-L57
all type are WriteOp for the sake of convenient

PR #1157, opened 2025-11-28
integrate secpk256 scalar invert precompile

hero78119:
Related change: patch invert in https://github.com/scroll-tech/elliptic-curves/commit/fdd6c563c1fee5028465eae665082acb1589f5e2

### benchmark 

proving block `23817600` now reduce from 401M -> 317M cycles, reduce around **20.94%**

after change, overall ecrecover for 168 tx cycles
```
secp256k1_ecrecover   total=4,922,568 cycles (1.55%)
└── recover_from_prehash total=4,053,504 cycles (1.28%)
    ├── decompress total=326,256 cycles (0.10%)
    └── invert     total=47,544 cycles (0.02%)
```

### DIFF
--- a/ceno_zkvm/src/instructions/riscv/ecall/uint256.rs
+++ b/ceno_zkvm/src/instructions/riscv/ecall/uint256.rs
@@ -22,7 +22,7 @@ use rayon::{
 use sp1_curves::{
     params::NumWords,
     uint256::U256Field,
-    utils::biguint_from_words,
+    utils::{biguint_from_be_words, biguint_from_le_words},
     weierstrass::{
         WeierstrassParameters,
         secp256k1::{Secp256k1, Secp256k1BaseField},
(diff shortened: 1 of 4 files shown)


ceno_zkvm/src/precompiles/uint256.rs
```diff
@@ -381,6 +380,260 @@ impl<E: ExtensionField> ProtocolWitnessGenerator<E> for Uint256MulLayout<E> {
...
+}
+
+#[derive(Clone, Debug)]
+#[repr(C)]
+pub struct Uint256InvLayer<WitT> {
+    pub wits: Uint256InvWitCols<WitT>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Uint256InvLayout<E: ExtensionField, Spec: Uint256InvSpec> {
```
hero78119, line 410: Implementation refer from `Uint256MulLayout` within same file.
Just simplify to use in Inv case (x = 1, modulus is constant, x_inv = 1 / x) and avoid redundant witness allocation
Besides also generalize to any U256 prime field to work.

ceno_zkvm/src/instructions/riscv/ecall/uint256.rs
```diff
@@ -348,3 +361,283 @@ impl<E: ExtensionField> Instruction<E> for Uint256MulInstruction<E> {
...
+}
+
+pub fn biguint_from_be_words(be_words: &[u32]) -> BigUint {
+    let mut bytes = Vec::with_capacity(be_words.len() * 4);
+    for word in be_words {
+        bytes.extend_from_slice(&word.to_le_bytes());
+    }
+    BigUint::from_bytes_be(&bytes)
+}
+
```
spherel: Previously, I put those functions in `gkr-backend`, such as https://github.com/scroll-tech/gkr-backend/blob/9be620b18900138d0b89569759fc2b3b9451b49e/crates/curves/src/utils.rs#L51-L57. Would you mind move it there?

hero78119: Done and move its to `gkr_backend` Previously I plan to do that later because I need to rename existing api `biguint_from_words` -> `biguint_from_le_words` to differentiate those 2, so it will involve changes in other precompile files

PR #1166, opened 2025-12-03
feat: upgrade ceno-gpu

Velaciela:
The key updates for ceno-gpu are:
- scroll-tech/ceno-gpu#102: Support GPU poly as view chunks
- scroll-tech/ceno-gpu#103: Batch tower witness building, halve the temp buffer usage
- scroll-tech/ceno-gpu#105: Memory-efficient modules: configurable GPU cache levels

---

`batch_commit` now behaves differently depending on the `CENO_GPU_CACHE_LEVEL` env var. The main differences are the **peak VRAM usage** during computation and whether the output `BasefoldCommitmentWithWitness` caches the trace and codeword.

| Value | Aliases| Trace(CPU) | Trace | Codeword | API | memory usage |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| `2` | `full` (defualt) | ❌ | ✅ | ✅ | `batch_commit_cache_full` | high |
| `1` | `trace`| ❌ | ✅ | ❌ | `batch_commit_cache_trace` | medium |
| `0` | `none`| ✅ | ❌ | ❌ | `batch_commit_cache_none` | low |

The corresponding support has been added to the `mmcs` and `batch_open` modules

### DIFF
--- a/ceno_zkvm/src/scheme/prover.rs
+++ b/ceno_zkvm/src/scheme/prover.rs
@@ -76,8 +76,7 @@ impl<
         }
     }
 
-    pub fn new(pk: ZKVMProvingKey<E, PCS>, device: PD) -> Self {
-        let pk = Arc::new(pk);
+    pub fn new(pk: Arc<ZKVMProvingKey<E, PCS>>, device: PD) -> Self {
         let (device_first_shard_pk, device_non_first_shard_pk) =
             if pk.as_ref().has_fixed_commitment() {
                 (
@@ -476,6 +475,13 @@ impl<
             pi_in_evals,
             input_opening_point,
         ))
+    }
+
+    pub fn setup_init_mem(&self, hints: &[u32], public_io: &[u32]) -> crate::e2e::InitMemState {
+        let Some(ctx) = self.pk.program_ctx.as_ref() else {
+            panic!("empty program ctx")
+        };
+        ctx.setup_init_mem(hints, public_io)
     }
 }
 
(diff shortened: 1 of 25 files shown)


ceno_zkvm/src/scheme/prover.rs
```diff
@@ -190,6 +191,16 @@ impl<
         // (circuit_name, num_instances)
         let name_and_instances = witnesses.get_witnesses_name_instance();
 
+        // check memory usage
+        #[cfg(feature = "gpu")]
+        {
+            use gkr_iop::gpu::gpu_prover::*;
+            println!("[create_proof] init");
```
hero78119: How about wrap it in `tracing::debug` instead of println?

Velaciela: I plan not to bring these two commits into the codebase yet and only use them during development. 
- temp: print mem info `6c94109`
- temp: timer `4b42c6d`

I’ll introduce the memory trace after the GPU memory pool is implemented.

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -240,6 +240,7 @@ impl<
 
         let mut points = Vec::new();
         let mut evaluations = Vec::new();
+        let mut trace_idx = 0usize; // tracks the index of circuits with witness columns (num_witin > 0) in PCS data
```
hero78119, line 243: We can hide `trace_idx` and make extract_witness_mles return an iterator. It's ok we do it next

PR #1179, opened 2025-12-08
GKR Cycle Reduction for `Keccak_syscall` Example

darth-cy:
- Output performance metrics
- Reduce GKR cycles for `keccak_syscall`


Cargo.toml
```diff
@@ -89,15 +89,15 @@ cudarc = { version = "0.17.3", features = ["driver", "cuda-version-from-build-sy
 
 # ceno-recursion dependencies
 openvm = { git = "https://github.com/scroll-tech/openvm.git", branch = "feat/v1.4.1-scroll-ext", default-features = false }
-openvm-circuit = { git = "https://github.com/scroll-tech/openvm.git", branch = "feat/v1.4.1-scroll-ext", default-features = false }
+openvm-circuit = { git = "https://github.com/scroll-tech/openvm.git", branch = "feat/v1.4.1-scroll-ext", features = ["perf-metrics"] }
```
hero78119: it better we make this `perf-metrics` being control via command line `--features perf-metrics` and not enable it by default

ceno_recursion/src/aggregation/mod.rs
```diff
@@ -832,7 +835,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "need to generate proof first"]
+    // #[ignore = "need to generate proof first"]
```
hero78119: This step still can NOT run in ci because we need a app proof first. Please keep this ignored

PR #1183, opened 2025-12-09
fixed dummy sha256 specs error

hero78119:
Found dummy circuit specs id, issue found from proving L1 block `23973870` which got `sha256_extend` syscall

register count should be 1, only `state_ptr`

This error only prompt in dummy circuit. For non-dummy ecall, this register count was no used.

related ceno_syscall upgrade
- https://github.com/scroll-tech/ceno-patch/commit/0cb2c93e02e2e859e55d5d889ca8abd0a13b13be
- https://github.com/scroll-tech/elliptic-curves/commit/05a8a81ef6300ff8bbf1d810b018bbeb71ab947c

### DIFF
--- a/ceno_emul/src/syscalls/secp256k1.rs
+++ b/ceno_emul/src/syscalls/secp256k1.rs
@@ -23,7 +23,7 @@ impl SyscallSpec for Secp256k1AddSpec {
 impl SyscallSpec for Secp256k1DoubleSpec {
     const NAME: &'static str = "SECP256K1_DOUBLE";
 
-    const REG_OPS_COUNT: usize = 2;
+    const REG_OPS_COUNT: usize = 1;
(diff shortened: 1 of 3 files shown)


ceno_emul/src/syscalls/secp256k1.rs
```diff
@@ -23,7 +23,7 @@ impl SyscallSpec for Secp256k1AddSpec {
 impl SyscallSpec for Secp256k1DoubleSpec {
     const NAME: &'static str = "SECP256K1_DOUBLE";
 
-    const REG_OPS_COUNT: usize = 2;
+    const REG_OPS_COUNT: usize = 1;
```
hero78119, line 26: This is used in dummy circuit previously. Right now no use, and we should remove this field to avoid confused later

PR #1189, opened 2025-12-10
optimize risc-v emulator 

hero78119:
Related to #1191 
To optimize emulator performance, a simple attempt is refactoring `Tracer::latest_accesses`, `VMState::memory` into Vector instead of Map, so the lookup is purely index access. 

### benchmark

test with mainnet block `23587691` for 98876199 (~2^26) instructions

| Metric                     | Value                    | Reduction |
|----------------------------|--------------------------|-----------|
| peak mem                  | 55.8  GB ->  55.8  GB                 |     no change   | 
| emulator.iter_until_halt  | 18.1s → 15.4s            | 14.92%    |

### DIFF
--- a/ceno_emul/src/vm_state.rs
+++ b/ceno_emul/src/vm_state.rs
@@ -100,6 +100,13 @@ impl VMState {
         })
     }
 
+    pub fn next_step_record(&mut self) -> Result<Option<StepRecord>> {
+        if self.halted() {
+            return Ok(None);
+        }
+        self.step().map(Some)
+    }
+
     fn step(&mut self) -> Result<StepRecord> {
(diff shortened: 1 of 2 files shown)


ceno_emul/src/dense_addr_space.rs
```diff
@@ -0,0 +1,59 @@
...
+        );
+        let len_words = (end.0 - base.0) as usize;
+        Self {
+            base,
+            end,
+            cells: vec![T::default(); len_words],
+        }
+    }
+
+    pub(crate) fn read(&self, addr: WordAddr) -> T {
```
kunxian-xia, line 29: why not return `Option<T>`?

PR #1193, opened 2025-12-11
Add Root and Halo2 Prover/Verifier

darth-cy:
- Write Ceno's independent root prover.
- Use the new internal proving logic.

For halo2 proof, run test test_static_verifier
- To test root AIR permutation using trace heights, use test test_root_proof_permutation
- Before running above test, run test_aggregation to generate a root proof.
- Download KZG params (default k=23) from these links: https://github.com/scroll-tech/openvm/blob/feat/v1.4.1-scroll-ext/docs/vocs/docs/pages/specs/security/security-model.mdx


ceno_recursion/src/aggregation/root.rs
```diff
@@ -4,3 +4,151 @@
...
+pub struct CenoRootVmVerifierConfig {
+    pub leaf_fri_params: FriParameters,
+    pub internal_fri_params: FriParameters,
+    pub num_user_public_values: usize,
+    pub internal_vm_verifier_commit: [F; DIGEST_SIZE],
+    pub compiler_options: CompilerOptions,
+}
+
+#[derive(Debug)]
+pub struct CenoRootVmVerifierPvs<T> {
```
kunxian-xia, line 69: We **must** also include `init_pc`.

darth-cy: `init_pc` is added.

ceno_recursion/src/aggregation/mod.rs
```diff
@@ -548,101 +613,101 @@ pub(crate) fn chunk_ceno_leaf_proof_inputs(
...
-    // } else if vm_commit_bn254 != *expected_vm_commit {
-    // return Err(eyre::eyre!(
-    // "Invalid app vm commit: expected {:?}, got {:?}",
-    // expected_vm_commit,
-    // vm_commit_bn254
-    // ));
-    // }
-    Ok(app_commit)
-}
+// pub fn verify_e2e_stark_proof(
```
kunxian-xia: This is commented out.

darth-cy: This is the verifier for the last internal proof (in `OpenVM`, it's called `e2e_stark_proof`) and should actually be deleted. Since the aggregation now outputs the root proof, the correct verifier is generated here: https://github.com/scroll-tech/openvm/blob/main/crates/sdk/src/lib.rs#L280-L296

ceno_cli/src/sdk.rs
```diff
@@ -194,25 +191,12 @@ where
...
-
+    ) -> Proof<RootSC> {
         // TODO: cache agg_prover
-        let mut agg_prover = if let Some(agg_pk) = self.agg_pk.as_ref() {
-            let leaf_prover = new_local_prover::<BabyBearPoseidon2Engine, NativeBuilder>(
-                vb.clone(),
-                &agg_pk.leaf_vm_pk,
-                agg_pk.leaf_committed_exe.exe.clone(),
+        let mut agg_prover = if let Some(_agg_pk) = self.agg_pk.as_ref() {
+            CenoAggregationProver::from_base_vk(
```
hero78119: We need to retain original way, which is init-prover giving existing `agg_pk`. This make lightweight agg_prover initialiation and avoid heavily keygen.

ceno_recursion/src/aggregation/mod.rs
```diff
@@ -356,13 +405,35 @@ impl CenoAggregationProver {
...
+        .expect("root proof generation should pass");
+
+        // Export root proof
+        let file = File::create("root_proof.bin").expect("Create export proof file");
+        bincode::serialize_into(file, &root_proof).expect("failed to serialize internal proof");
+
+        root_proof
+    }
+
+    pub fn verify_root_proof(&self, root_proof: &Proof<RootSC>) -> Result<(), VerificationError> {
```
hero78119: This api is good for `prover` to verify, but verifier need another version because only `vk` presented


Suggest to change
```rust
pub fn verify_root_proof(&self, root_proof: &Proof<RootSC>) -> Result<(), VerificationError> {
    verify_root_proof(&self.vk.root_vm_vk, root_proof)
}
```
and define another util function `verify_root_proof` in aggregation/mod.rs which take vk as input. Verifier do NOT have `pk`, so it need another util function `verify_root_proof` to  serve verify purpose

hero78119 (changes requested):
Try new api in this PR on https://github.com/scroll-tech/ceno-reth-benchmark e2e, need to adjust further of new api design

ceno_recursion/src/aggregation/mod.rs
darth-cy: A separate verification function is added

ceno_cli/src/sdk.rs
darth-cy: Restored the original initiation

hero78119 (commented):
Reported with e2e test with block `23587691`
- e2e prove/verify pass
- final root proof size: 671K
- currently only cpu mode, as I didn't find respective gpu config for root prover. On cpu root proof take `89.78s`. We need a way to switch to gpu config

PR #1198, opened 2025-12-12
Fix(recursion): support multiple instances of a circuit

kunxian-xia:


ceno_zkvm/src/scheme/prover.rs
```diff
@@ -183,9 +183,10 @@ impl<
 
                 let circuit_idx = self.pk.circuit_name_to_index.get(name).unwrap();
                 // write (circuit_idx, num_var) to transcript
-                transcript.append_message(&circuit_idx.to_le_bytes());
+                transcript.append_field_element(&E::BaseField::from_canonical_usize(*circuit_idx));
```
kunxian-xia, line 186: The behavior of `usize.to_le_bytes()` is different on 64-bit cpu and 32-bit cpu. This kind of divergence is not desirable.

ceno_zkvm/src/scheme/prover.rs
```diff
@@ -183,9 +183,10 @@ impl<
...
                 for num_instance in num_instances {
-                    transcript.append_message(&num_instance.to_le_bytes());
+                    transcript
+                        .append_field_element(&E::BaseField::from_canonical_usize(*num_instance));
```
kunxian-xia, line 189: ditto.

ceno_recursion/src/zkvm_verifier/verifier.rs
```diff
@@ -242,143 +244,148 @@ pub fn verify_zkvm_proof<C: Config<F = F>>(
         .range(0, chip_indices.len())
         .for_each(|idx_vec, builder| {
             let i = idx_vec[0];
```
hero78119, line 246: TODO soundness: check omc init table validity https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/scheme/verifier.rs#L250-L265

PR #1206, opened 2025-12-24
Add fptower precompile and tests

dreamATD:
# Description
Add fptower precompiles.
Run the following commands to test:
```
cargo run --release --package ceno_zkvm --bin e2e -- --platform=ceno examples/target/riscv32im-ceno-zkvm-elf/release/examples/bn254_fptower_syscalls
```

### DIFF
--- a/ceno_zkvm/src/precompiles/fptower/fp2_addsub.rs
+++ b/ceno_zkvm/src/precompiles/fptower/fp2_addsub.rs
@@ -327,23 +327,23 @@ mod tests {
     use super::*;
 
     use ff_ext::{BabyBearExt4, SmallField};
-    use gkr_iop::circuit_builder::{CircuitBuilder, ConstraintSystem};
-    use itertools::Itertools;
-    use std::sync::Arc;
-    use num::BigUint;
-    use rand::RngCore;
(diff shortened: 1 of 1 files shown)


ceno_zkvm/src/precompiles/fptower/fp.rs
```diff
@@ -0,0 +1,529 @@
...
+    pub n_structural_witin: usize,
+    pub n_challenges: usize,
+}
+
+impl<E: ExtensionField, P: FpOpField> FpOpLayout<E, P> {
+    fn new(cb: &mut CircuitBuilder<E>) -> Self {
+        let wits = FpOpWitCols {
+            is_add: cb.create_witin(|| "fp_op_is_add"),
+            is_sub: cb.create_witin(|| "fp_op_is_sub"),
+            is_mul: cb.create_witin(|| "fp_op_is_mul"),
```
hero78119, line 119: These op flags are for reused same chip for different ecall opcode but currently we still build them into different chip, so this boolean columns along with respective constraints are redundant overhead.

Its ok for now, later we tweak design to implement different opcode share same chip. Its equivalent to batch main sumcheck and I believe it will bring more performance gain in particular in GPU: less kernel invoke, less cpu/gpu execution interleaving

PR #1213, opened 2025-12-31
feat: add sha extend precompile.

dreamATD:
# Description
## Sha extend circuit
This PR adds sha extend precompile, computing the parameter `w[0..64]` defined as follows:
```rust
// w[0..16] are copied from the input
for i in 16..64 {
    let s0 = w[i - 15].rotate_right(7) ^ w[i - 15].rotate_right(18) ^ (w[i - 15] >> 3);
    let s1 = w[i - 2].rotate_right(17) ^ w[i - 2].rotate_right(19) ^ (w[i - 2] >> 10);
    w[i] = w[i - 16]
        .wrapping_add(s0)
        .wrapping_add(w[i - 7])
        .wrapping_add(s1);
}
```

The sha extend circuit represents a single round computation in sha extend. It fetches the instruction at the same pc as the first cycle, reads the same register, and defines the global state with increasing cycles. I each instance, it reads w[i - 2], w[i - 7], w[i - 15], w[i - 16] from the memory and computes w[i] and writes it back.

## Changes in `ceno_patch`
The function is replaced from `syscall_sha256_extend(w_i: &mut [u32; 64]);` to `syscall_sha256_extend(w_i: &mut u32)`:
```
/// Executes one SHA256 extend round in place.
#[allow(unused_variables)]
pub fn syscall_sha256_extend(w_i: &mut u32) {
    #[cfg(target_os = "zkvm")]
    {
        let w_i = w_i as *mut u32;
        unsafe {
            asm!(
            "ecall",
            in("t0") SHA_EXTEND,
            in("a0") w_i,
            );
        }
    }

    #[cfg(not(target_os = "zkvm"))]
    unreachable!()
}
```
And the functionality in `ceno_emul` is changed accordingly.

## Tiny change of `SelectTypeLayout`
The original definition is as follows:
```rust
#[derive(Clone, Debug)]
pub struct SelectorTypeLayout<E: ExtensionField> {
    pub sel_mem_read: SelectorType<E>,
    pub sel_mem_write: SelectorType<E>,
    pub sel_lookup: SelectorType<E>,
    pub sel_zero: SelectorType<E>,
}
```
while they are sometimes duplicated and in sha extend circuit, it needs to set `round_i = 16` in the first round and update `pc` during the last round. Therefore,  the definition is changed as follows:
```
#[derive(Clone, Debug)]
pub struct SelectorTypeLayout<E: ExtensionField> {
    pub sel_first: Option<SelectorType<E>>,
    pub sel_last: Option<SelectorType<E>>,
    pub sel_all: SelectorType<E>,
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3002,7 +3002,7 @@ dependencies = [
 [[package]]
 name = "k256"
 version = "0.13.4"
-source = "git+https://github.com/scroll-tech/elliptic-curves?branch=ceno%2Fk256-13.4#e7c4aaa5f44b20ff948292510ed28c4e0a567211"
+source = "git+https://github.com/scroll-tech/elliptic-curves?branch=ceno%2Fk256-13.4#17adc274db2fb10510449026ec785ae4fc234540"
 dependencies = [
  "ceno_crypto_primitives",
  "ceno_syscall",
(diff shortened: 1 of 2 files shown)


hero78119 (commented):
First quick pass and LGTM!

Just one add-on: as this PR introduce OrderSparse change, we also need to modify recursion verifier, and I will submit another PR for it 👍

PR #1225, opened 2026-01-14
optimize preflight executor

hero78119:
### Design rationales
Refactor shard cycle logic into pre-flight tracer in low overhead way. With that, in preflight tracer we only update `next_access` with cross-shard memory access. This optimization bring much gain on preflight executor performance

### benchmark
test on `23817600`

| Stage                               | Before                | After                 |
|-------------------------------------|-----------------------|-----------------------|
|e2e| 195s | 161s |
| emulator.preflight-execute (total)  | 39.5 s · 20.10%       |     10.4 s · 6.44%                  |
| emulator.new-preflight-tracer       |                       | 3.01 ms · 0.00%       |
| emulator.init_mem                   |                       | 9.94 ms · 0.01%       |

### DIFF
--- a/ceno_emul/src/tracer.rs
+++ b/ceno_emul/src/tracer.rs
@@ -95,11 +95,11 @@ pub trait Tracer {
     const SUBCYCLE_MEM: Cycle = 3;
     const SUBCYCLES_PER_INSN: Cycle = 4;
 
-    fn new(platform: &Platform, config: &Self::Config) -> Self;
+    fn new(platform: &Platform, config: Self::Config) -> Self;
 
     fn with_next_accesses(
         platform: &Platform,
-        config: &Self::Config,
+        config: Self::Config,
         next_accesses: Option<Arc<NextCycleAccess>>,
     ) -> Self
     where
@@ -887,10 +887,9 @@ pub struct PreflightTracer {
     latest_accesses: LatestAccesses,
     next_accesses: NextCycleAccess,
     register_reads_tracked: u8,
-    record_next_accesses: bool,
     planner: Option<ShardPlanBuilder>,
     current_shard_start_cycle: Cycle,
-    step_cell_extractor: Option<Arc<dyn StepCellExtractor>>,
+    config: PreflightTracerConfig,
 }
 
 #[derive(Clone)]
@@ -912,10 +911,9 @@ impl fmt::Debug for PreflightTracer {
             .field("latest_accesses", &self.latest_accesses)
             .field("next_accesses", &self.next_accesses)
             .field("register_reads_tracked", &self.register_reads_tracked)
-            .field("record_next_accesses", &self.record_next_accesses)
             .field("planner", &self.planner)
             .field("current_shard_start_cycle", &self.current_shard_start_cycle)
-            .field("step_cell_extractor", &self.step_cell_extractor.is_some())
+            .field("config", &self.config)
             .finish()
     }
 }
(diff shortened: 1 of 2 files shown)


ceno_emul/src/tracer.rs
```diff
@@ -235,6 +248,116 @@ impl LatestAccesses {
...
+    pub fn new(max_cell_per_shard: u64, max_cycle_per_shard: Cycle) -> Self {
+        let initial_cycle = FullTracer::SUBCYCLES_PER_INSN;
+        ShardPlanBuilder {
+            shard_cycle_boundaries: vec![initial_cycle],
+            max_cell_per_shard,
+            target_cell_first_shard: if max_cell_per_shard == u64::MAX {
+                u64::MAX
+            } else {
+                max_cell_per_shard
+            },
```
kunxian-xia: this condition is redundant as this expression always evaluates to `max_cell_per_shard`.

ceno_emul/src/tracer.rs
```diff
@@ -235,6 +248,116 @@ impl LatestAccesses {
...
+            self.max_cell_per_shard
+        };
+
+        // always include step in current shard to simplify overall logic
+        self.cur_cells = self.cur_cells.saturating_add(step_cells);
+        self.cur_cycle_in_shard = self
+            .cur_cycle_in_shard
+            .saturating_add(FullTracer::SUBCYCLES_PER_INSN);
+
+        let cycle_limit_hit = self.max_cycle_per_shard < Cycle::MAX
```
kunxian-xia: `self.max_cycle_per_shard` should always less than `Cycle::MAX`. Therefore we don't need to check this condition.

ceno_emul/src/tracer.rs
```diff
@@ -760,14 +883,104 @@ impl FullTracer {
...
 pub struct PreflightTracer {
     cycle: Cycle,
     pc: Change<ByteAddr>,
+    last_kind: InsnKind,
+    last_rs1: Option<Word>,
     mmio_min_max_access: Option<BTreeMap<WordAddr, (WordAddr, WordAddr, WordAddr, WordAddr)>>,
     latest_accesses: LatestAccesses,
     next_accesses: NextCycleAccess,
     register_reads_tracked: u8,
+    record_next_accesses: bool,
```
kunxian-xia: why not just include `config: PreflightTracerConfig` since the fields `record_next_access` and `step_cell_extractor` are already there?

ceno_emul/src/tracer.rs
```diff
@@ -39,9 +39,17 @@ pub struct StepRecord {
     syscall: Option<SyscallWitness>,
 }
 
+pub trait StepCellExtractor {
+    fn cells_for_kind(&self, kind: InsnKind, rs1_value: Option<Word>) -> u64;
```
kunxian-xia, line 43: why not just pass `Instruction` to this function?

ceno_emul/src/tracer.rs
hero78119: We didnt take the ownership of XXXTracerConfig so just extract field from config. Yeah I think take ownership make overall design more clean and done in https://github.com/scroll-tech/ceno/pull/1225/commits/396e6aa0e24283c6d2d1c6beb908fb6a302eabfc

ceno_emul/src/tracer.rs
hero78119: 👍 clean up in https://github.com/scroll-tech/ceno/pull/1225/commits/396e6aa0e24283c6d2d1c6beb908fb6a302eabfc

ceno_emul/src/tracer.rs
hero78119: 👍 clean up in https://github.com/scroll-tech/ceno/pull/1225/commits/396e6aa0e24283c6d2d1c6beb908fb6a302eabfc

ceno_emul/src/tracer.rs
hero78119: Instrunction not work for this purpose. Instrunction provide rs1 **index** but we need rs1_value register value.

kunxian-xia (approved):
LGTM!

PR #1235, opened 2026-01-20
ecc quark gpu prover

### LINKED ISSUE #1232: EC quark prover in GPU

hero78119:
To close #1232 

Move ecc quark major mles of septic curve and sumcheck to GPU. selector remain in cpu as its overhead are negligible, due to the num rows in ecc quark are relatively small, .e.g ~2^21.

### benchmark on larger block `24169100 `

> `23817600` block are relative `small` where only 14 shards, thus we can only see ~2s e2e different

24169100 with 24 shards and each have shard proofs

| Stage / Component                       | Before Time | After Time |
|----------------------------------------|-------------|------------|
| reth-block (block 24169100)             | 249 s       | 240 s      |
| app.prove                              | 169 s       | 160 s      |
| prove_ec_sum_quark (shard id 0)   | 640 ms      | 49.2 ms    |
| prove_ec_sum_quark (shard id 1)   | 682 ms      | 49.6 ms    |
| prove_ec_sum_quark (shard id 2)   | 650 ms      | 50.6 ms    |
| prove_ec_sum_quark (shard id 3)   | 355 ms      | 28.3 ms    |
...

### DIFF
--- a/ceno_zkvm/src/scheme/gpu/mod.rs
+++ b/ceno_zkvm/src/scheme/gpu/mod.rs
@@ -10,19 +10,37 @@ use crate::{
     },
     structs::{ComposedConstrainSystem, PointAndEval, TowerProofs},
 };
+use ceno_gpu::bb31::{CudaHalBB31, GpuPolynomial};
+use either::Either;
 use ff_ext::ExtensionField;
 use gkr_iop::{
-    gkr::{self, Evaluation, GKRProof, GKRProverOutput, layer::LayerWitness},
-    gpu::{GpuBackend, GpuProver},
+    gkr::{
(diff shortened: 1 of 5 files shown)


ceno_zkvm/src/scheme/gpu/mod.rs
```diff
@@ -320,46 +352,43 @@ fn build_tower_witness_gpu<'buf, E: ExtensionField>(
                 last_layer
             })
             .collect::<Vec<_>>()
+    } else if lk_denominator_last_layer.is_empty() {
```
hero78119, line 350: No logic change, just make clippy happy

Issue #1233, opened 2026-01-20
sumcheck prover optimization


kunxian-xia:
- https://eprint.iacr.org/2025/1117
- [Some Improvements for the PIOP for ZeroCheck](https://eprint.iacr.org/2024/108.pdf) from polygon zero

PR #1247, opened 2026-01-28
sumcheck skip send p0

hero78119:
Related to #1233 
refer https://eprint.iacr.org/2024/108.pdf "sec 3.1"

### benchmark 

send one less interpolation show significant improvement on `prove_tower_relation_gpu`. For `prove_main_constraints`, `pcs_opening` there is no significant different and most likely noise

| Operation                     | master (s) | degree-1 (s) | Improve (master→degree-1) |
|------------------------------|------------|---------------|----------------------------|
| prove_tower_relation_gpu     | 32.689     | 26.446        | -19.14%                     |
| prove_main_constraints       | 9.932      | 10.146        | +2.15%                     |
| pcs_opening                  | 19.081     | 19.386        | +1.60%                     |

| Layer     | master (s) | degree-1 (s) | Improve (master→degree-1) |
|-----------|-------------|--------------|----------------------------|
| app_prove | 87.000      | 82.100       | 5.63%                      |

raw data
- master https://github.com/scroll-tech/ceno-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/feat/latest/mainnet23817600-20260126-165027_summary.md
- degree -1 https://github.com/scroll-tech/ceno-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/feat/latest/mainnet23817600-20260128-161951_summary.md

> after degree - 1 now trigger more gap, we need to optimize per shard step records generation + witness assignment

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2235,7 +2235,7 @@ dependencies = [
 [[package]]
 name = "ff_ext"
 version = "0.1.0"
-source = "git+https://github.com/scroll-tech/gkr-backend.git?rev=8ce1577d61af241606d30aefe9ac896e7502809a#8ce1577d61af241606d30aefe9ac896e7502809a"
+source = "git+https://github.com/scroll-tech/gkr-backend.git?tag=v1.0.0-alpha.21#77b22021fcbd863ce16fe697258ddb0101418f62"
 dependencies = [
  "once_cell",
  "p3",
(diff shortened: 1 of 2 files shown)


ceno_zkvm/src/scheme/cpu/mod.rs
```diff
@@ -295,7 +295,6 @@ impl CpuEccProver {
             }
         }
         let final_sum = SepticPoint::from_affine(final_sum_x, final_sum_y);
-        assert_eq!(zerocheck_proof.extract_sum(), E::ZERO);
```
hero78119, line 298: this util function `extract_sum` no longer valid as now uni-variate lack of p0. We can do the zero check in sanity stage

PR #1253, opened 2026-02-09
fork transcript to support chip concurrent prove

hero78119:


ceno_zkvm/src/scheme/prover.rs
```diff
@@ -352,6 +369,15 @@ impl<
             drop(witness_iter);
             exit_span!(main_proofs_span);
 
+            // merge forked transcript into transcript
+            let forked_sampling = forked_transcripts
+                .into_iter()
+                .map(|mut fork_transcript| fork_transcript.sample_vec(1)[0])
```
kunxian-xia, line 375: why only sample 1 ext?

PR #1254, opened 2026-02-11
perf: Concurrent chip proving

Velaciela:
better merge after #1250: to observe perf gain after the overlap gap in witness is reduced

`CENO_SCHEDULER_WORKERS_NUM` sets the number of concurrent streams, defaulting to `64`.
`RUST_MIN_STACK=16777216` multi-thread execution needs this

for dev & ci: `CENO_GPU_MEM_TRACKING=1 CENO_CONCURRENT_CHIP_PROVING=0` to verify GPU memory estimation

---

## Summary

This PR introduces a **memory-aware parallel scheduler** for chip proof generation on the GPU, enabling multiple chip proofs to run concurrently while respecting GPU VRAM limits. The key idea is a greedy backfilling algorithm that maximizes GPU utilization and eliminates long-tail latency from large chips blocking the pipeline.

## Motivation

Previously, chip proofs were generated **sequentially** — one circuit at a time. This left significant GPU idle time, especially when small chip proofs could run in parallel with large ones. By scheduling chip proofs concurrently with preemptive memory reservation, we can significantly improve overall proving throughput.

### 1. New Scheduler Module (`scheduler.rs`)
- Implements a **greedy backfilling algorithm**: tasks are sorted by memory requirement (descending, "big rocks first"), and the scheduler tries to fit the largest task first; if it doesn't fit, it backfills with smaller tasks.
- Supports both **sequential** and **concurrent** execution modes, controlled by `CENO_CONCURRENT_CHIP_PROVING` env var (default: concurrent on GPU).
- Uses `std::thread::scope` with a fixed worker pool (8 workers) and MPSC channels for task dispatch and completion notification.
- Integrates with a **CUDA memory pool** (`CudaMemPool`) for preemptive VRAM booking/unbooking, preventing OOM during parallel execution.
- Handles worker panics (catches unwind, converts to `ZKVMError`) to avoid scheduler deadlocks.

### 2. GPU Memory Estimation (`gpu/memory.rs`)
- New module providing **pre-execution GPU memory estimation** for each chip proof stage:
  - `estimate_trace_bytes` — witness polynomial extraction + structural MLEs
  - `estimate_main_witness_bytes` — read/write/logup record polynomials
  - `estimate_tower_bytes` — tower build + prove stages
  - `estimate_ecc_quark_bytes_from_num_vars` — ECC quark sumcheck
  - `estimate_main_constraints_bytes` — GKR zerocheck + rotation sumcheck
- `estimate_chip_proof_memory` computes the **peak memory** as: `resident + max(stage temporaries) + safety margin`
- **Dev-mode validation** (`CENO_GPU_MEM_TRACKING=1 CENO_CONCURRENT_CHIP_PROVING=0`): `check_gpu_mem_estimation` compares estimated vs. actual GPU memory usage at each stage, with configurable tolerance (1MB under-estimate, 5MB over-estimate).

### 3. Standalone `_impl` Functions for Thread Safety
- Extracted `prove_tower_relation_impl`, `prove_main_constraints_impl`, `prove_ec_sum_quark_impl` as **standalone functions** (no `&self`) from the original trait methods.
- This eliminates `Send`/`Sync` requirements on `GpuProver`/`ZKVMProver`, enabling parallel execution without capturing `&self`.
- The original trait methods now delegate to these standalone functions.
- Added `create_chip_proof_gpu_impl` as a standalone GPU-specific entry point for the concurrent scheduler.

### 4. Deferred Witness Extraction (GPU)
- Introduces a **`ChipInputPreparer` trait** with GPU/CPU implementations.
- GPU path: witness MLEs and structural witnesses are extracted **just-in-time** per task (deferred extraction), rather than all-at-once upfront. This reduces peak VRAM usage.
- CPU path: no-op (input is eagerly populated during task building, as before).
- New standalone functions: `extract_witness_mles_for_trace`, `transport_structural_witness_to_gpu`.

### 5. Per-Thread CUDA Stream Support
- GPU operations now use **per-thread CUDA streams** (`get_thread_stream()`, `bind_thread_stream()`), enabling concurrent kernel execution across worker threads.
- All GPU allocation/kernel calls (tower build, sumcheck, filter, etc.) are updated to accept an optional stream parameter.

### 6. Prover Refactoring: 3-Phase Architecture
The prover's `create_proof_of_shard` is restructured into three clean phases:
- **Phase 1** (`build_chip_tasks`): Build `ChipTask` structs from witness data, with `#[cfg]` for eager (CPU) vs deferred (GPU) extraction.
- **Phase 2** (`run_chip_proofs`): Execute tasks via the scheduler. Concurrent mode uses standalone `_impl` functions; sequential mode uses `self.create_chip_proof`.
- **Phase 3** (`collect_chip_results`): Aggregate results into `BTreeMap<circuit_idx, Vec<ZKVMChipProof>>`, points, evaluations, and PI updates.

### 7. `Rc` → `Arc` Migration
- `CpuBackend` and `GpuBackend` references changed from `Rc` to `Arc` to support cross-thread sharing required by the concurrent scheduler.

### 8. Transcript Forking
- Transcript forking is now handled internally by the scheduler: each worker clones the parent transcript, appends the `task_id`, and samples one extension-field element after proving. This avoids sending non-`Send` transcript objects across threads.

---

## Benchmark
    - APP Prove - baseline(master + witgen-perf#1250): **74.500s**
    - APP Prove - concurrent chip proving: **60.800s** 
      - (1.225x speedup @ 4090 24GB block#600)
---

### DIFF
--- a/ceno_zkvm/src/scheme/gpu/memory.rs
+++ b/ceno_zkvm/src/scheme/gpu/memory.rs
@@ -11,36 +11,21 @@ use ceno_gpu::{
 use ff_ext::ExtensionField;
 use gkr_iop::gpu::{
     BB31Base, GpuBackend,
-    gpu_prover::{BB31Ext, CudaHalBB31, MemTracker},
+    gpu_prover::{
+        BB31Ext, CacheLevel, CudaHalBB31, MemTracker, get_gpu_cache_level, get_mem_tracking_mode,
+    },
 };
 use mpcs::PolynomialCommitmentScheme;
 use std::sync::OnceLock;
 
 use crate::scheme::scheduler::{ChipProvingMode, get_chip_proving_mode};
 
-static MEM_TRACKING_MODE: OnceLock<bool> = OnceLock::new();
-
-static TRACE_CACHE_LEVEL: OnceLock<bool> = OnceLock::new();
-
-fn get_trace_cached_stats() -> bool {
-    *TRACE_CACHE_LEVEL.get_or_init(|| {
-        let cache_level =
-            std::env::var("CENO_GPU_CACHE_LEVEL").unwrap_or_else(|_| "full".to_string());
-        matches!(cache_level.as_str(), "2" | "full" | "1" | "trace")
-    })
-}
-
-pub fn get_mem_tracking_mode() -> bool {
-    *MEM_TRACKING_MODE
-        .get_or_init(|| matches!(std::env::var("CENO_GPU_MEM_TRACKING").as_deref(), Ok("1")))
-}
-
 pub fn init_gpu_mem_tracker<'a>(
     cuda_hal: &'a CudaHalBB31,
     label: &'static str,
 ) -> Option<MemTracker<'a>> {
     let is_sequential = get_chip_proving_mode() == ChipProvingMode::Sequential;
-    let is_mem_tracking = get_mem_tracking_mode();
+    let is_mem_tracking = get_mem_tracking_mode() == true;
     if is_sequential && is_mem_tracking {
         Some(cuda_hal.inner.mem_tracker(label))
     } else {
@@ -337,16 +322,18 @@ pub(crate) fn estimate_tower_bytes<E: ExtensionField, PCS: PolynomialCommitmentS
 /// - `resident`: poly copies that remain as witness MLEs after extraction
 /// - `temporary`: temp_buffer allocation (2x), freed after extraction
 ///
-/// Returns `(0, 0)` when trace is cached (default), because get_trace creates views without allocation.
+/// Returns `(0, 0)` when trace is cached (`CacheLevel::Trace` or `CacheLevel::Full`),
+/// When cache is disabled (`CacheLevel::None`, the default), estimates actual allocation costs.
 pub(crate) fn estimate_trace_extraction_bytes(num_witin: usize, num_vars: usize) -> (usize, usize) {
-    if get_trace_cached_stats() {
-        (0, 0)
-    } else {
+    if matches!(get_gpu_cache_level(), CacheLevel::None) {
+        // Default cache level is None
         let base_elem_size = std::mem::size_of::<BB31Base>();
         let mle_len = 1usize << num_vars;
         let poly_bytes = num_witin * mle_len * base_elem_size;
         // get_trace allocates poly copies (resident) + temp_buffer (2x, freed after)
         (poly_bytes, 2 * poly_bytes)
+    } else {
+        (0, 0)
     }
 }
 
--- a/gkr_iop/src/gpu/mod.rs
+++ b/gkr_iop/src/gpu/mod.rs
@@ -26,9 +26,10 @@ pub mod gpu_prover {
             CudaHalBB31, GpuDigestLayer, GpuFieldType, GpuMatrix, GpuPolynomial, GpuPolynomialExt,
         },
         common::{
+            CacheLevel,
             basefold::utils::convert_ceno_to_gpu_basefold_commitment,
             buffer::BufferImpl,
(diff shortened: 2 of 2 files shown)


ceno_zkvm/src/scheme/gpu/mod.rs
```diff
@@ -243,6 +705,88 @@ impl<E: ExtensionField, PCS: PolynomialCommitmentScheme<E>> TraceCommitter<GpuBa
...
+        .map(|mle| Arc::new(MultilinearExtensionGpu::from_ceno(&cuda_hal, mle)))
+        .collect();
+
+    let estimated_bytes = estimate_structural_mle_bytes(num_structural_witin, num_vars);
+    check_gpu_mem_estimation(gpu_mem_tracker, estimated_bytes);
+
+    result
+}
+
+#[allow(clippy::too_many_arguments)]
```
hero78119: misc: duplicate definition

ceno_zkvm/src/scheme/scheduler.rs
```diff
@@ -0,0 +1,445 @@
...
+                    tasks_inflight += 1;
+                    task_tx.send(task).unwrap();
+                } else if tasks_inflight > 0 {
+                    // C: Block wait for completion when nothing fits
+                    tracing::info!(
+                        "[scheduler] Pool full, waiting for task completion... pool_booked={:.2}MB, inflight={}",
+                        pool.get_booked_total() as f64 / (1024.0 * 1024.0),
+                        tasks_inflight
+                    );
+                    if let Ok(msg) = done_rx.recv() {
```
hero78119: ignore error without break will probably lock main-thread here even some worker are panic.
it should be fixed in #1256

Velaciela: fixed in [cad3a64](https://github.com/scroll-tech/ceno/pull/1254/commits/cad3a64b03512d4abcc70b8d337730a471f48e40)

gkr_iop/src/gkr/layer/gpu/mod.rs
```diff
@@ -455,10 +426,17 @@ pub(crate) fn prove_rotation_gpu<E: ExtensionField, PCS: PolynomialCommitmentSch
     let span = entered_span!("rotation derived left/right eval", profiling_3 = true);
     let bh = BooleanHypercube::new(rotation_cyclic_group_log2);
     let (left_point, right_point) = bh.get_rotation_points(&row_challenges_e);
+    // Capture parent thread's CUDA stream so Rayon workers can reuse it.
+    // TODO: GPU batch evaluation and its batch version
```
hero78119, line 430: TODO: fix me notes as it make prove_rotation to be sequential as they shared same parent stream id, which is ok since prove rotation cost are low

ceno_zkvm/src/scheme/scheduler.rs
```diff
@@ -0,0 +1,505 @@
...
+            for _worker_id in 0..num_workers {
+                let rx = Arc::clone(&task_rx);
+                let tx = done_tx.clone();
+                let execute_fn = &execute_task;
+                let tr = &transcript_ref;
+
+                s.spawn(move || {
+                    loop {
+                        let task = {
+                            let lock = rx.lock().unwrap();
```
hero78119, line 333: A note: we prefer worker reuse instead of round-robin, but it depends on rayon thread schedule. If we require stream id to be maximal re-used, we can comback and think about how to tune this part

This only affect readibility, not affect performance

hero78119 (approved):
This bring a huge milestone for Ceno GPU utilization in base layer! 
Learn a lot during the reviewing. Awesome work & LGTM 💯

PR #1259, opened 2026-03-03
feat: GPU witness generation (RV32IM + Keccak + ShardRam)

Velaciela:
related: scroll-tech/ceno#1265

# GPU Witness Generation

Accelerate witness generation by offloading computation from CPU to GPU.
This module (`ceno_zkvm/src/instructions/gpu/`) contains all GPU-side dispatch,
caching, and utility code for the witness generation pipeline.

The CUDA backend lives in the sibling repo `ceno-gpu/` (`cuda_hal/src/common/witgen/`).

### Module Layout

```
gpu/
├── dispatch.rs         — GPU dispatch entry point (try_gpu_assign_instances, gpu_fill_witness)
├── config.rs           — Environment variable config (3 env vars), kind tags
├── cache.rs            — Thread-local device buffer caching, shared EC/addr buffers
├── chips/              — Per-chip column map extractors + chip-specific GPU dispatch
│   ├── add.rs ... sw.rs  (24 RV32IM column map extractors)
│   ├── keccak.rs         (column map + keccak GPU dispatch: gpu_assign_keccak_instances)
│   └── shard_ram.rs      (column map + batch EC computation: gpu_batch_continuation_ec)
├── utils/
│   ├── column_map.rs   — Shared column map extraction helpers (extract_rs1, extract_rd, ...)
│   ├── d2h.rs          — Device-to-host: witness transpose, LK counter decode, compact EC D2H
│   ├── debug_compare.rs— GPU vs CPU comparison (activated by CENO_GPU_DEBUG_COMPARE_WITGEN)
│   ├── lk_ops.rs       — LkOp enum, SendEvent struct
│   ├── sink.rs         — LkShardramSink trait, CpuLkShardramSink
│   ├── emit.rs         — Emit helper functions (emit_u16_limbs, emit_logic_u8_ops, ...)
│   ├── fallback.rs     — CPU fallback: cpu_assign_instances, cpu_collect_lk_and_shardram
│   └── test_helpers.rs — Test utilities: assert_witness_colmajor_eq, assert_full_gpu_pipeline
└── mod.rs              — Module declarations + lk_shardram integration tests (19 tests)
```

### Data Flow

```
                    Pass 1: PreflightTracer
                    ┌──────────────────────┐
                    │  ShardPlanBuilder     │ → shard boundaries
                    │  addr_future_accesses │ → next-access HashMap (GPU cache reads and sorts before H2D)
                    └──────────┬───────────┘
                               │
                    Pass 2: FullTracer (per shard)
                    ┌──────────▼───────────┐
                    │  Vec<StepRecord>      │ 136 bytes/step, #[repr(C)]
                    └──────────┬───────────┘
                               │ H2D (cached per shard in cache.rs)
                    ┌──────────▼───────────────────────────────────┐
                    │              GPU Per-Instruction              │
                    │  ┌─────────────┬──────────────┬────────────┐ │
                    │  │ F-1 Witness │ F-2 LK Count │ F-3 EC/Addr│ │
                    │  │ (col-major) │  (atomics)   │ (shared buf)│ │
                    │  └──────┬──────┴──────┬───────┴─────┬──────┘ │
                    └─────────┼─────────────┼─────────────┼────────┘
                              │             │             │
                      GPU transpose    D2H counters   flush at shard end
                              │             │             │
                    ┌─────────▼─────────────▼─────────────▼────────┐
                    │                 CPU Merge                     │
                    │  RowMajorMatrix  LkMultiplicity  ShardContext │
                    └──────────────────────┬───────────────────────┘
                                           │
                    ┌──────────────────────▼───────────────────────┐
                    │           ShardRamCircuit (GPU)               │
                    │  Phase 1: per-row Poseidon2 (344 cols)       │
        …11562 chars truncated…zkvm/src/instructions/gpu/chips/sw.rs

### CODE REVIEW

File: ceno_emul/src/tracer.rs [hunk 1]
```diff
@@ -972,6 +1108,7 @@ pub struct PreflightTracer {
     mmio_min_max_access: Option<BTreeMap<WordAddr, (WordAddr, WordAddr, WordAddr, WordAddr)>>,
     latest_accesses: LatestAccesses,
     next_accesses: NextCycleAccess,
+    next_accesses_vec: Vec<PackedNextAccessEntry>,
```
hero78119: We can re-build `SortedNextAccesses` from `self.next_accesses` map so we can avoid introduce this new vector field and `PackedNextAccessEntry`

hero78119 (commented):
A quick review regarding to tracer & SortedNextAccesses field

ceno_emul/src/tracer.rs
Velaciela: > re-build

slower than this method (roughly 50ms vs 200ms)
`CENO_NEXT_ACCESS_SOURCE=preflight` vs `CENO_NEXT_ACCESS_SOURCE=hashmap`

hero78119: I believe the slow come from the sequential traverse of hashmap
how about par_extend() ?
```rust
                info_span!("next_access_from_hashmap").in_scope(|| {
                    let total_entries: usize =
                        addr_future_accesses.values().map(|pairs| pairs.len()).sum();
                    let mut entries = Vec::with_capacity(total_entries);
                    entries.par_extend(addr_future_accesses.par_iter().flat_map_iter(
                        |(cycle, pairs)| {
                            pairs
                                .iter()
                                .map(move |&(addr, next_cycle)| {
                                    PackedNextAccessEntry::new(*cycle, addr.0, next_cycle)
                                })
                        },
                    ));
                    entries
                })
```

If this work, we can remove vector version from tracer and only rebuild here and gated by gpu feature

Velaciela:
```
# using preflight-appended vec (9321311 entries)
┝━ [ceno] preflight-execute [ 7.85s | 6.11% ]
┝━ next_access_presort [ 52.6ms | 0.00% / 0.04% ]
│  ┝━ ｉ [info]: [next-access presort] 
│  ┝━ next_access_par_sort [ 52.5ms | 0.04% ] n: 9321311

# converting from HashMap - Serial
┝━ [ceno] preflight-execute [ 7.85s | 6.17% ]
┝━ next_access_presort [ 110ms | 0.01% / 0.09% ]
│  ┝━ next_access_from_hashmap [ 68.4ms | 0.05% ]
│  ┝━ next_access_par_sort [ 33.1ms | 0.03% ] n: 9321311

# converting from HashMap - Parallel
┝━ [ceno] preflight-execute [ 7.76s | 4.49% ]
┝━ next_access_presort [ 174ms | 0.00% / 0.10% ]
│  ┝━ next_access_from_hashmap [ 138ms | 0.08% ]
│  ┝━ next_access_par_sort [ 28.5ms | 0.02% ] n: 9321311
```

1. Each element's work (PackedNextAccessEntry::new) is a few assignments and bit-shifts — trivial for the parallelism gains to outweigh Rayon's scheduling/synchronization overhead.
2. The cost is mainly memory allocation and HashMap random access, which are memory-bandwidth-bound and only get worse with multiple threads contending for cache.

Velaciela: Perhaps we could switch to the rebuild-from-hashmap approach first, without introducing interface changes, and address performance concerns later.

Velaciela: - [remove: next_accesses_vec](https://github.com/scroll-tech/ceno/pull/1259/commits/1fe56b81965c0b97dac4fd96cf02e8735c8da4e2)
- [remove: sorted_next_accesses from ShardContextBuilder and ShardContext](https://github.com/scroll-tech/ceno/pull/1259/commits/411a62b658d101b48ac545fa61252a977174c66e)
- [move: PackedNextAccessEntry to GPU codebase](https://github.com/scroll-tech/ceno/pull/1259/commits/2513a2feff152687fc06f90310b40066fd341f3a)
- [remove: gpu_ec_records from ShardContext](https://github.com/scroll-tech/ceno/pull/1259/commits/204e2903bc69074b4dc5516400f391c371acf6f4)

And ShardRamCircuit cleanup:
- [remove: gpu_ec_records_to_shard_ram_inputs from structs.rs](https://github.com/scroll-tech/ceno/pull/1259/commits/327b0edc74e3b4ca7dace985d32a6385ebe94e55)
- [refactor: move GPU shard_ram impl to gpu/chips/shard_ram.rs](https://github.com/scroll-tech/ceno/pull/1259/commits/50b44113794839e35bccc8467c54a9b5afcebcdf)

Velaciela:
GPU witgen now [runs successfully](https://github.com/scroll-tech/ceno-reth-benchmark/actions/runs/24176035636) on the reth-benchmark machine. Notes:                                                                                           
- GPU witgen still requires D2H copies back to CPU and runs serially with GPU proving (not yet perf- or production-ready)
- [`7090881`](https://github.com/scroll-tech/ceno/pull/1259/commits/7090881030ed8ee947fa92c05eafd97a94436fa2) refactor witgen cache to release all GPU memory between shards (to avoid occupying VRAM needed by the proving stage)
- The witgen module increases CUDA runtime memory overhead; requires bumping [RESERVED_FOR_RUNTIME to 900 MB](https://github.com/scroll-tech/ceno-gpu/pull/142/changes/2d8eb44da74b533d57ad73f7e7163ba313ee4866) to fit.

hero78119:
This follow-up range extends the original GPU witgen work into a full `WITGEN=1`, `CACHE_LEVEL=none` proving flow.

**Flow Overview**

ASCII overview of the current flow:

```text
Shard execution
  |
  v
GPU witgen per chip
  - build witness directly on GPU
  - collect shard-side raw state on GPU
    * step indices
    * shard metadata
    * keccak/shard-ram raw buffers
    * compact EC / address access side data
  |
  \--> CACHE_LEVEL = none
         do NOT retain all witness/device-backed traces
         keep only:
           - replay metadata per chip
           - shard-resident raw GPU state shared by the shard
         so peak VRAM is lower after commit
```

```text
Later, during chip proof / opening:

ChipTask
  |
  +--> has gpu_replay_plan?
          |
          +--> no: use normal committed witness path
          |
          \--> yes:
                 replay_witness()
                   |
                   +--> use shard-resident raw GPU state
                   +--> use per-chip replay metadata
                   +--> reconstruct witness again on GPU
                   +--> attach col-major device backing
                   +--> feed witness directly into chip proving / PCS flow
```

```text
Why this exists:

without replay:
  all chip witnesses/device-backed traces stay alive across commit/prove/open
  -> simpler
  -> much higher steady-state VRAM

with WITGEN=1 + CACHE=none:
  witness is produced early, dropped after commit, then rebuilt on demand
  -> lower retained VRAM
  -> more complex replay/device-backing invariants
```

**Main Changes In This Range**

- completes the deferred/replayable GPU witness path so `CENO_GPU_ENABLE_WITGEN=1` can run with `CENO_GPU_CACHE_LEVEL=0`
- adds the replay plumbing needed for `ShardRam` and structural witness handling
- reworks memory estimation / booking / release so large-payload runs are less likely to OOM
- aligns `gkr_iop` / `ceno_zkvm` GPU interfaces around device-backed witness extraction and deferred reuse
- includes follow-up fixes for build, clippy, and e2e/verification regressions discovered after the initial GPU witgen landing

**Why `CENO_GPU_LARGE_TASK_BOOKING_MARGIN_MB` Is Needed**

- scheduling is based on estimated peak memory, but the actual peak is not just the final witness size
- large tasks temporarily need extra headroom for:
  - transpose/materialization buffers
  - tower build / prove scratch
  - PCS commit / opening scratch
  - temporary coexistence of replayed witness + structural witness + codeword/opening buffers
- if scheduling books only the nominal estimate, two "safe" tasks can still overlap and exceed real VRAM at transient peaks
- the extra booking margin is therefore a scheduler safety buffer, not a logical requirement of the protocol

**Important Remaining Pressure Point**

- batch open still restores all required device backing up front before entering the actual batch-open work
- that means replayed traces / codewords can all become live together at the boundary before openings are consumed
- so even with `CACHE=none`, peak VRAM can still spike during PCS opening because restoration is eager rather than streamed / windowed
- this is the main remaining architectural source of high VRAM pressure in the deferred path
- the likely long-term direction is to optimize this path and keep `WITGEN=1` as the only GPU mode, with `CACHE_LEVEL=none` as the retained policy instead of supporting multiple GPU caching modes

**Residual Risks / Gaps**

- replay correctness depends on preserving the "device-backed col-major witness is immutable after attachment" invariant
- scheduler success still depends on estimation quality plus booking headroom, especially for keccak-heavy or large multi-chip shards
- PCS opening is still the place where deferred restoration can bunch up and create a second VRAM peak even after witness replay reduced earlier pressure

**Brief Summary**

- `#1259` introduced GPU witness generation
- this follow-up range makes that path practically usable for bigger workloads by adding replayable cache-none proving, `ShardRam` replay support, and stronger VRAM scheduling / lifecycle management
- the main remaining optimization opportunity is to make batch-open restoration more incremental so restored device backing does not pile up all at once

hero78119 (approved):

PR #1275, opened 2026-03-24
Trace Gen for Tower and Proof Shape

darth-cy:


ceno_recursion_v2/src/continuation/prover/inner/mod.rs
```diff
@@ -163,11 +163,13 @@ where
         }
 
         let engine = E::new(self.pk.params.clone());
-        #[cfg(debug_assertions)]
-        debug_constraints(&self.circuit, &ctx, &engine);
+        /* debug block: Step 1 placeholder - skip strict debug constraint checks */
```
hero78119: we can bypass `#[cfg(debug_assertions)]` check via release build
This is my sample testing command,
`RUST_LOG=debug RUST_MIN_STACK=33554432 RUST_BACKTRACE=1 cargo test --release -p ceno_recursion_v2 leaf_app_proof_round_trip_placeholder -- --nocapture`

PR #1277, opened 2026-03-25
refactor(zkvm): canonicalize PI transcript/opening flow via instance_value + PublicValues

hero78119:
This PR refactors how public inputs (PI) are represented and consumed across zkVM proving/verification so that PI handling is circuit-driven and consistent end-to-end.
Related to #1263.

### Design rationale
Previously, PI transcript absorption and some verifier checks depended on broader/raw PI containers, which made ordering and indexing assumptions harder to audit and keep consistent across:
- prover
- rust verifier
- recursion verifier.

This PR moves PI consumption to a canonical, circuit-defined path:
- `instance_values` defines which scalar PI values a circuit observes (and in what order).
- `instance_openings` defines which PI MLE columns are opened.
- PI openings are derived from `PublicValues` MLEs, rather than ad-hoc/raw vectors.

This reduces ambiguity in PI ordering/indexing and aligns transcript-visible PI handling with circuit semantics.

### pi update
Previous `pi_update` mechanism will trigger verifier failed if a identical public value used > 1 chip because they will be in different opening point, thus evaluation value will be different. This PR fixed the issue, and refactor public value check back to `verify_chip_phase`

### soundness fix
This PR also fixed a soundness issue by removing `wit_in_evals` and `fixed_in_evals` from the proof payload. Instead of trusting prover-supplied values for these evaluations, the verifier now deterministically derives them from the sumcheck evaluations during verification. This prevents a malicious prover from introducing mismatches between claimed eval vectors and the actual sumcheck-consistent values.

### Follow up
#1280


kunxian-xia (approved):
LGTM

PR #1292, opened 2026-04-03
recursion-v2 preparation: isolate base-layer chip transcripts

hero78119:
Move base-layer chip metadata binding out of the main transcript and into each per-chip fork transcript.

Design rationale: per-chip proofs should own the challenge + fork id + chip metadata observations they depend on. This keeps native proving/verifying and the recursion verifier aligned on transcript order while preserving independent chip scheduling.


hero78119:
Done in #1363

PR #1298, opened 2026-04-07
misc: remove dead n_challenges field

dreamATD:
# Description
`n_challenges` on `Layer`, `Chip`, and `GKRCircuit` was always 0 — every
layout struct initialized it to 0 and never assigned any other value.
The design intended per-layer challenge sampling, but it was never
actually used.

This PR removes the field and all its plumbing:
- `Layer.n_challenges` field and the `update_challenges` method it drove
- `n_challenges` parameter from `Layer::new`, `Layer::from_circuit_builder`, `LayerConstraintSystem::into_layer`, and `LayerConstraintSystem::into_layer_with_lookup_eval_iter`
- `Chip.n_challenges` and the corresponding `Chip::new_from_cb` parameter
- `GKRCircuit.n_challenges`
- `ProtocolBuilder::n_challenges()` trait method
- `generate_layer_challenges` in the recursion verifier (replaced by passing `challenges` directly)

No behavior change — `sample_and_append_challenge_pows(0, ...)` was a no-op, and `update_challenges` with n=0 neither sampled nor wrote anything meaningful.

### DIFF
--- a/ceno_recursion/src/zkvm_verifier/verifier.rs
+++ b/ceno_recursion/src/zkvm_verifier/verifier.rs
@@ -812,13 +812,11 @@ pub fn verify_gkr_circuit<C: Config>(
 
     for (i, layer) in gkr_circuit.layers.iter().enumerate() {
         let layer_proof = builder.get(&gkr_proof.layer_proofs, i);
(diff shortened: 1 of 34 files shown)


hero78119 (approved):
LGTM 👍

PR #1302, opened 2026-04-10
refactor(zkvm,gkr): simplify chip finalize flow

dreamATD:
## Summary

This PR simplifies the chip `finalize` flow by moving layer construction into each chip/layout's `finalize` implementation, so callers no longer need to manually compute `out_evals`, build a `Layer`, and call `chip.add_layer(...)`.

## What changed

- Added a shared `default_out_eval_groups(...)` helper in `gkr_iop`.
- Updated `finalize(...)` implementations to directly assemble and return a fully built `Chip`.
- Removed duplicated caller-side boilerplate in precompile, instruction, and table builders.

## Why

This makes the finalize contract simpler and more consistent: callers ask for a finalized chip and receive a ready-to-use one, instead of partially constructing it in multiple places.

## Validation

- `cargo check`
- `cargo check --all-targets`

### DIFF
--- a/gkr_iop/src/chip.rs
+++ b/gkr_iop/src/chip.rs
@@ -29,7 +29,7 @@ pub struct Chip<E: ExtensionField> {
 }
 
 impl<E: ExtensionField> Chip<E> {
-    pub fn new_from_cb(cb: &CircuitBuilder<E>) -> Chip<E> {
+    pub fn new_from_cb(cb: &CircuitBuilder<E>, _n_challenges: usize) -> Chip<E> {
         Self {
             n_fixed: cb.cs.num_fixed,
             n_committed: cb.cs.num_witin as usize,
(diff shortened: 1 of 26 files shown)


gkr_iop/src/chip.rs
```diff
@@ -29,7 +29,7 @@ pub struct Chip<E: ExtensionField> {
 }
 
 impl<E: ExtensionField> Chip<E> {
-    pub fn new_from_cb(cb: &CircuitBuilder<E>) -> Chip<E> {
+    pub fn new_from_cb(cb: &CircuitBuilder<E>, _n_challenges: usize) -> Chip<E> {
```
hero78119: does `_n_challenges`  some left-over?

hero78119 (changes requested):
👍 apart from one comment, other LGTM!

gkr_iop/src/chip.rs
dreamATD: Sorry, yes. I've fixed this in the latest commit.

hero78119 (approved):
LGTM!

PR #1321, opened 2026-04-22
feat(verifier): replace panics with error returns

kunxian-xia:
## Problem

A malformed proof could cause the verifier process to crash — any `unwrap` / `expect` / unchecked indexing / `assert!` on proof-derived data is a liveness / DoS risk (a crafted proof kills the process instead of being cleanly rejected).

## Changes

- **Panic cleanup.** Convert every `assert!` / `assert_eq!` / `unwrap` / `expect` / unchecked indexing on proof-derived data to `ZKVMError` returns across `verify_proofs_halt`, `verify_proof_validity`, `verify_chip_proof`, `TowerVerify::verify`, and `EccVerifier::verify_ecc_proof`. A malformed proof is now rejected cleanly in all paths.
- **Document the verifier's semantic contract.** New sections in `CLAUDE.md` and `docs/src/technical-overview.md` ("What the verifier guarantees") state the two program-level facts a valid Ceno proof attests to: **execution starts at `vk.entry_pc`** and **the terminal shard invokes the halt ecall**. The exit code is deliberately *not* a verifier guarantee — `public_values.exit_code` is bound by the halt-ecall chip to register `a0`, but the guest program defines its own exit-code semantics, so a non-zero value may be a legitimate application signal. Callers that want "exited successfully" compare `exit_code == 0` themselves.
- `CLAUDE.md` additionally flags prefix proofs (`expect-halt = false`) as a dev/bench affordance, not a production surface. This caveat is contributor-facing and is kept out of the user-facing mdbook.

## Test plan

- [x] `cargo check --workspace --all-targets`
- [x] `cargo make clippy` (workspace, `-D warnings`)
- [x] `cargo test -p ceno_zkvm --lib` (152 passed)
- [x] `cargo test -p ceno_zkvm --lib scheme::` (18 passed)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,6 +77,24 @@ Before declaring a change done: at minimum run `cargo make clippy` and
 the tests for the crate you touched. For anything in prover/verifier
 code paths, also run `cargo make tests` end-to-end.
 
+## Verifier semantic contract
+
+A valid Ceno proof attests to exactly two program-level facts:
+**execution starts at the verifying key's entry PC**, and
+**execution exits with code zero**. Everything else in the verifier
+(transcript, sumcheck, PCS, tower/GKR, shard EC sum) is plumbing that
+makes those two statements meaningful. The caller-supplied "expect
+halt" flag picks the terminal mode (full run vs. prefix run); the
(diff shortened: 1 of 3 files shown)


hero78119 (approved):
LGTM!

PR #1322, opened 2026-04-22
feat(verifier): commit terminal mode on verifier instance

kunxian-xia:
Follow-up to #1321.

Moves the full-trace / prefix-run / single-shard-debug distinction from a per-call `expect_halt` flag onto the verifier instance as a `VerifierMode` field. Fixes the e2e callers that were deriving `expect_halt` from the proof itself, which made the halt-consistency check tautological.

- `VerifierMode { FullRun, PrefixRun, DebugSegment }`; `ZKVMVerifier::new(vk)` defaults to `FullRun`.
- `verify_full_trace_proofs_halt` + `verify_single_shard_segment_halt` collapsed into one mode-dispatched `verify_proofs`.
- Callers pick mode at construction from out-of-band knowledge (`exit_code`, `target_shard_id`), never from the proof.
- `CLAUDE.md` + `docs/src/technical-overview.md` rewritten around the three modes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,11 +82,12 @@ code paths, also run `cargo make tests` end-to-end.
 What a valid Ceno proof attests to depends on the **verifier mode**,
 committed at verifier construction and not derived from the proof:
 
-- **FullRun** — trace from `vk.entry_pc` to halt. Production default.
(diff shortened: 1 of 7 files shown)


kunxian-xia:
Per discussion offline, we think this change is not necessary any more.

PR #1223, opened 2026-01-12
#1084 follow up: verifier check on dynamic init table

hero78119:
## Problem

Dynamic heap/hint init tables need verifier-side checks so shard continuation and dynamic-length constraints cannot be bypassed. This PR also needs to stay compatible with current master after the linker/public-io cleanup and the later single-shard debug verification support.

## Design Rationale

Keep the memory-state verifier ISA-extensible by carrying RV32-specific heap/hint bounds in RV32imMemStateConfig, while enforcing continuation and dynamic-init checks in the native and recursion verifier paths. Merge on top of master instead of reverting master-only changes.

## Change Highlights

- ceno_zkvm
  - restore verifier checks for dynamic heap/hint init tables
  - enforce heap/hint continuation and proof-size checks across shards
  - keep ZKVMVerifier / ZKVMVerifyingKey extensible with a mem-state verifier generic
  - merge master single-shard e2e verification flow and fix its halt expectation for debug shard verification
- ceno_recursion
  - restore heap/hint bound checks in aggregation leaf verification
  - merge shard-id plumbing for single-shard e2e_aggregate
- ceno_emul / ceno_rt
  - reconcile this PR's memory layout swap with master's removal of the PUBLIC I/O linker term
  - fix emulator dense-memory bounds for the merged layout

## Benchmark / Performance Impact

No intended performance change beyond verifier-side checks. Previous measurements on this work showed negligible overhead; no new benchmark run was needed for the merge-only follow-ups.

Benchmark command(s): not rerun for the merge-only follow-ups.

Environment (CPU/GPU, core count, rust toolchain, commit hash): validated on local dev environment at head 7da2e884f7e860b35ecae49fdb322ce02a5f27b8.

raw data:
- master: n/a
- this PR: n/a

## Testing

- cargo make clippy
- cargo check --config net.git-fetch-with-cli=true -p ceno_zkvm -p ceno_recursion --bins --release
- cargo run --config net.git-fetch-with-cli=true --release --package ceno_zkvm --bin e2e -- --platform=ceno --max-cycle-per-shard=1600 examples/target/riscv32im-ceno-zkvm-elf/release/examples/keccak_syscall
- cargo run --config net.git-fetch-with-cli=true --features gpu --release --package ceno_zkvm --bin e2e -- --platform=ceno --max-cycle-per-shard=1600 examples/target/riscv32im-ceno-zkvm-elf/release/examples/keccak_syscall
[...]

PR #1323, opened 2026-04-23
[docs] mem continuation check

hero78119:
Add docs follows #1223 


docs/src/memory-continuation-checks.md
```diff
@@ -0,0 +1,87 @@
+# Memory Continuation Checks
```
kunxian-xia, line 1: @copilot does this doc implies the main intuition/idea that heap/init does not need to initialized in the first shard? We can load it until it's first accessed. Then we have to require that not a single address are being initialized twice.

kunxian-xia (approved):

Issue #1340, opened 2026-05-19
finish the TODOs in `shard_ram.rs`


kunxian-xia:
The TODOs in [`shard_ram.rs`](https://github.com/scroll-tech/ceno/blob/master/ceno_zkvm/src/tables/shard_ram.rs#L280) has big impact on our soundness guarantee of cross shard executions. E.g. you can refer to #1338 for an example of breach / exploit. Therefore we **MUST** add proper constraints to complete these TODOs.

PR #1343, opened 2026-05-25
fix(soundness): bind ShardRam y-sign to is_global_write

### LINKED ISSUE #1338: ShardRam read/write direction is not bound to the public `shard_rw_sum`
## Summary

This issue was reproduced against `master` at commit `298ea12198463b3d820b1c68fc3f180f42ec0825` (`2026-05-04`).

The scope of the reproduction is the official example guest `ceno_rt_alloc` and the standard proving and verification flow:

- program setup
- emulation
- witness generation
- proof creation
- full-trace base verification
- recursion verifier program execution

For the same guest execution, Ceno accepts two different proof batches with different public `shard_rw_sum` values:

- an honest proof batch generated from the normal `ShardRam` witness
- a malicious proof batch generated after negating the post-witgen `ShardRam` points and updating the public `shard_rw_sum` accordingly

The prover and both verifiers are otherwise unchanged. The malicious proof batch is accepted by the normal full-trace verifier and by the recursion verifier program.

The attachment bundle contains:

- the reproduction harness: [attachments/reproduce/reproduce_shard_ram_direction.rs](attachments/reproduce/reproduce_shard_ram_direction.rs)
- the official example guest: [attachments/reproduce/ceno_rt_alloc.rs](attachments/reproduce/ceno_rt_alloc.rs)
- a rerun script: [attachments/reproduce/run.sh](attachments/reproduce/run.sh)
- the proof artifacts and public-value diff: [attachments/artifacts/](attachments/artifacts/)
- a sanitized execution log: [attachments/results/run.log](attachments/results/run.log)

This issue appears related to the earlier public ShardRam soundness review in `#1097`, but it does not appear to be the same fixed issue as `#1098`. The change in `#1098` addressed record typing in the read/write bus path, while this report concerns a different gap: the `ShardRam` circuit still does not bind the EC point sign to read/write direction. The attached reproduction also shows current-`master` acceptance of a malicious proof batch through both full-trace verification and the recursion verifier program.

## Root cause analysis

The intended design is documented clearly: the `x` coordinate is derived from the cross-shard RAM record, and the sign of `y` is used to encode read vs write. The relevant documentation excerpt is included in [attachments/reproduce/architecture_overview_excerpt.md](attachments/reproduce/architecture_overview_excerpt.md).

The host-side witness code follows that design. In `ShardRamRecord::to_ec_point`, it flips the sign of `y` so that reads and writes land in different halves of the field; see [attachments/reproduce/shard_ram_root_cause_excerpt.rs](attachments/reproduce/shard_ram_root_cause_excerpt.rs).

The circuit, however, does not enforce that convention. It constrains:

- the record fields
- the Poseidon2-derived `x` coordinate
- the EC accumulation against the public `shard_rw_sum`

but it does not constrain the `y` sign to match `is_global_write`. The same source file already notes this as a missing condition.

As a result, both `(x, y)` and `(x, -y)` satisfy the same record binding for a given witness row. Once the `ShardRam` witness is generated, an attacker can negate the real leaf points, recompute the internal EC tree, update the public `shard_rw_sum`, and still produce a proof that verifies successfully.

In the attached reproduction, the mutation is applied consistently across all shards, so the final cross-shard infinity check still passes even though the per-shard public sums change.

## Reproduce steps

1. Check out Ceno at commit `298ea12198463b3d820b1c68fc3f180f42ec0825`.
2. Run:

```bash
./attachments/reproduce/run.sh /path/to/ceno
```

3. Inspect [attachments/results/run.log](attachments/results/run.log). The expected output includes:
   - `execution split into 17 shards`
   - `honest base verifier accepted`
   - `malicious base verifier accepted`
   - `honest recursion verifier program accepted`
   - `malicious recursion verifier program accepted`

4. Compare the per-shard public values in [attachments/artifacts/shard-public-values.txt](attachments/artifacts/shard-public-values.txt).

The attached proof artifacts are from one successful run:

- [attachments/artifacts/honest.proof.bin](attachments/artifacts/honest.proof.bin)
- [attachments/artifacts/malicious.proof.bin](attachments/artifacts/malicious.proof.bin)
- [attachments/artifacts/vk.bin](attachments/artifacts/vk.bin)

## Impact

This is a statement-binding failure on a documented public value.

For the same execution trace, the verifier accepts two different public cross-shard RAM summaries. In the attached run:

- the exit code is unchanged
- the control-flow public values are unchanged
- the heap and hint range public values are unchanged
- the public I/O digest is unchanged
- all 17 shard-level `shard_rw_sum` values change

That makes `shard_rw_sum` unreliable as a public summary of cross-shard RAM flow. Any downstream component that treats this value as a canonical commitment to read/write direction is exposed to inconsistent but accepted proofs.

The issue also survives recursion. The malicious proof batch is accepted not only by the base verifier, but also by the recursion verifier program.

## Recommended fix

1. Enforce the `y`-sign convention in-circuit.
   - If `is_global_write = 1`, require the `y` encoding selected for writes.
   - If `is_global_write = 0`, require the `y` encoding selected for reads.

2. Bind direction directly into the committed point representation if that is simpler operationally.
   - For example, include the direction bit in the point derivation, rather than relying on an unconstrained sign choice.

3. Add an end-to-end regression test.
   - Start from an honest multi-shard execution.
   - Negate all `ShardRam` witness points, recompute the internal tree, and update the public sum.
   - Proof creation or verification should fail.

**Attachments**
https://drive.google.com/file/d/1zn6RzDofVdlv2nCd4TPgJfwtDddQMCVj/view?usp=sharing

dreamATD:
## Problem

Issue #1338 demonstrates a working soundness exploit on `master`. For the
same RISC-V execution trace, the base verifier *and* the recursion verifier
both accept two different proof batches whose per-shard public
`shard_rw_sum` values differ on **all 17** shards. The attacker takes an
honest witness, replaces every cross-shard EC accumulator leaf `(x, y)`
with its inverse `(x, -y)`, updates `shard_rw_sum`, and reproves.

Root cause: `ceno_zkvm/src/tables/shard_ram.rs:276-281` was a TODO. The
host code in `ShardRamRecord::to_ec_point` uses the sign of `y[6]` to
encode read vs write, but the circuit only constrained the curve equation
and the EC sum — never tying `y[6]`'s half-of-field to `is_global_write`.
Both `(x, y)` and `(x, -y)` satisfied every existing check, so the
public summary of cross-shard RAM flow was unbound.

The bug also survived recursion (per the reporter's reproduction).

## Design Rationale

Approach mirrors SP1's `crates/core/machine/src/operations/global_interaction.rs:210-236`
**idea** but uses Ceno's existing gadgets, not SP1's column layout.
Three pieces:

1. **Offset by +1.** Express `y[6]` in terms of a fresh witness `y6_lo`
   such that `y[6] = 0` is never valid in either branch (it is the
   2-torsion fixed point where both signs collapse).
2. **Safe band + prover retry.** Restrict `y6_lo` to `[0, (p-1)/2)`. For
   the rare exception `y[6] = 0` (probability `~1/p ≈ 2^-31` per record)
   the host rejects and retries with a new `nonce`.
3. **Byte-decomposition range check.** `y6_lo` decomposed into four byte
   limbs `b0..b3` (`assert_byte` for `b0..b2`, `lookup_ltu_byte(b3, 60, 1)`
   for `b3`). For BabyBear, `(p-1)/2 = 60·2^24` exactly, so `b3 < 60`
   gives the tightest no-overlap band.

In-circuit branch equality enforced via `condition_require_equal`:

- read  (`is_global_write = 0`): `y[6] = y6_lo + 1`  ⇒ `y[6] ∈ [1, (p-1)/2]`
- write (`is_global_write = 1`): `y[6] = p - 1 - y6_lo` ⇒ `y[6] ∈ [(p+1)/2, p-1]`

Union covers `[1, p-1]` with no overlap; `y[6] = 0` is the unique
exception, blocked in both branches.

**Why not a single `AssertLtConfig(y6_lo, (p-1)/2, max_bits=30)`?**
On BabyBear (`p = 0x78000001`, 31-bit) a naive AssertLt fails to enforce
canonical `<` because the diff can wrap mod p. A malicious `y6_lo ∈
[0x74000001, p-1]` (≈ 2^26 values) produces a wrapped diff that still
fits in 30 bits, so the constraint accepts upper-half values and the
exploit survives. Pre-bounding `y6_lo` via byte decomposition kills the
wrap. Ceno's `DynamicRangeTableCircuit<E, 18>` also doesn't carry the
30-bit entries that a single `assert_const_range(_, 30)` would request.

**Why M = 60 (vs SP1's 63).** SP1 targets KoalaBear; its `(p-1)/2 =
0x3f800000`, so 63 leaves a small safety band. For BabyBear,
`(p-1)/2 = 60·2^24` exactly — using 63 here would let `y[6]` straddle
`p/2`, reintroducing the ambiguity.

Also corrects the stale comment on `shard_ram.rs:277-278`, which had
the convention reversed (it claimed write ⇒ lower half but the host
code does the opposite).

## Change Highlights

- `ceno_zkvm/src/tables/shard_ram.rs`
  - `ShardRamRecord::to_ec_point`: reject `y6 == 0` and try the next
    `nonce`. Same negation policy otherwise.
  - `ShardRamConfig`: new field `y6_lo_bytes: [WitIn; 4]`.
  - `ShardRamConfig::configure`: replace the TODO with the byte
    decomposition, byte-range / LTU lookups, and the
    `condition_require_equal` branch equality. Updates the stale comment.
  - `ShardRamCircuit::assign_instance`: compute `y6_lo` from `y[6]` and
    `is_to_write_set`, assign byte limbs, register byte and LTU
    multiplicities.
  - New regression test `test_shard_ram_y_sign_attack_breaks_byte_bound`
    that (1) verifies honest points land in the expected half and
    (2) shows that the #1338 attack would force `b3 ≥ 60`, making the
    in-circuit LTU lookup unsatisfiable.

Per ShardRam row this PR adds **4 byte WitIn columns** plus 3
byte-range and 1 LTU lookup multiplicity. ShardRam rows scale with
cross-shard RAM events, not with cycles, so the absolute cost is
sub-percent on the prover. No full prover bench was rerun (no
hot-loop arithmetic changed).

Existing `test_shard_ram_circuit` (170k reads + 1420 writes, full
chip proof) runtime is unchanged within noise on this machine:

```text
master   : ~5.0 s
this PR  : ~5.0 s
```

## Testing

```sh
cargo fmt --all --check
cargo check --workspace --all-targets --release
cargo make clippy
RUST_MIN_STACK=33554432 cargo test -p ceno_zkvm tables::shard_ram::tests --lib --release
```

All pass locally on BabyBear. Both `test_shard_ram_circuit` and the
new `test_shard_ram_y_sign_attack_breaks_byte_bound` are green.

`cargo make tests` / `cargo make tests_goldilock` should be re-run by
CI; the change is feature-gated to BabyBear via the `debug_assert_eq!`
on `MODULUS_U64` and goldilocks does not exercise shard_ram (per
`integration.yml` commented-out lines and CLAUDE.md).

## Risks and Rollout

- **Soundness.** Closes #1338. The new constraint is hot-loop adjacent
  (per shard_ram row) but only adds local byte arithmetic and existing
  lookups — no change to transcript, sumcheck, PCS, or EC accumulation.
- **GPU.** The CUDA witness generator
  (`instructions/gpu/chips/shard_ram.rs` → `witgen_shard_ram_per_row`)
  lives in the separate `ceno-gpu-mock` repo and does not yet populate
  the new `y6_lo_bytes` columns. CPU runs are correct; GPU runs (behind
  the `gpu` feature, not in default CI) will fail at proof time until
  the kernel is updated. **Follow-up required before re-enabling GPU
  shard_ram.**
- **Recursion.** The recursive verifier consumes the same constraint
  system, so no separate recursion code change is needed; once the
  base proof passes, the recursive one will too. The reporter's PoC
  going through the recursion verifier should now be rejected.
- **Field support.** Hardcodes the BabyBear constant `M = 60`. A
  `debug_assert_eq!(MODULUS_U64, 0x78000001, ...)` guards against
  accidental use on a different field; shard_ram is BabyBear-only
  today per CLAUDE.md.

## Follow-ups

- Port the y6_lo_bytes population to the `ceno-gpu-mock` CUDA witgen
  kernel and re-enable GPU shard_ram.
- The remaining #1340 TODOs (`local read ⇄ global write` pairing on
  `shard_ram.rs:235-236`, `shard == shard_id` binding on line 244)
  are intentionally out of scope here. Both should land as separate
  reviewable PRs.

Fixes #1338.
Partially addresses #1340.

### DIFF
--- a/ceno_zkvm/src/tables/shard_ram.rs
+++ b/ceno_zkvm/src/tables/shard_ram.rs
@@ -133,24 +133,29 @@ impl ShardRamRecord {
                 hasher.permute(input.clone())[0..SEPTIC_EXTENSION_DEGREE].into();
             if let Some(p) = SepticPoint::from_x(x) {
                 let y6 = (p.y.0)[SEPTIC_EXTENSION_DEGREE - 1].to_canonical_u64();
-                let is_y_in_2nd_half = y6 >= (prime / 2);
+                // y6 == 0 is the 2-torsion exception: both signs collapse, so
+                // the in-circuit y-sign binding cannot distinguish read/write.
+                // Reject and try a new nonce. Probability per record ~ 1/p ≈ 2^-31.
+                if y6 != 0 {
+                    let is_y_in_2nd_half = y6 >= (prime / 2);
 
-                // we negate y if needed
-                // to ensure read => y in [0, p/2) and write => y in [p/2, p)
-                let negate = match (self.is_to_write_set, is_y_in_2nd_half) {
-                    (true, false) => true, // write, y in [0, p/2)
-                    (false, true) => true, // read, y in [p/2, p)
-                    _ => false,
-                };
+                    // negate y to enforce host convention:
+                    //   read  (is_to_write_set = 0) => y6 in [1, (p-1)/2]
+                    //   write (is_to_write_set = 1) => y6 in [(p+1)/2, p-1]
+                    let negate = match (self.is_to_write_set, is_y_in_2nd_half) {
+                        (true, false) => true, // write, y in [1, p/2), flip to upper half
+                        (false, true) => true, // read, y in [p/2, p), flip to lower half
+                        _ => false,
+                    };
 
-                let point = if negate { -p } else { p };
+                    let point = if negate { -p } else { p };
 
-                return ECPoint { nonce, point };
-            } else {
-                // try again with different nonce
-                nonce += 1;
-                input[6] = E::BaseField::from_canonical_u32(nonce);
+                    return ECPoint { nonce, point };
+                }
             }
+            // try again with different nonce
+            nonce += 1;
+            input[6] = E::BaseField::from_canonical_u32(nonce);
         }
     }
 }
@@ -180,6 +185,9 @@ pub struct ShardRamConfig<E: ExtensionField> {
     pub(crate) x: Vec<WitIn>,
     pub(crate) y: Vec<WitIn>,
     pub(crate) slope: Vec<WitIn>,
+    // 4-byte decomposition of `y6_lo`, the helper used to bind the sign of
+    // `y[SEPTIC_EXTENSION_DEGREE - 1]` to `is_global_write`. See `configure`.
+    pub(crate) y6_lo_bytes: [WitIn; 4],
     pub(crate) perm_config: Poseidon2Config<E, 16, 7, 1, 4, 13>,
 }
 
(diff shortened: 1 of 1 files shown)


dreamATD:
Closing — opened from the wrong GitHub account. Will reopen under `sphere` shortly. Branch `fix/issue-1338-shard-ram-y-sign` is unchanged.

PR #1344, opened 2026-05-25
fix(soundness): bind ShardRam y-sign to is_global_write

### LINKED ISSUE #1338: ShardRam read/write direction is not bound to the public `shard_rw_sum`

### LINKED ISSUE #1340: finish the TODOs in `shard_ram.rs`

spherel:
## Problem

Issue #1338 reproduces a soundness break on `master`. For the same RISC-V
execution, the base verifier *and* the recursion verifier both accept two
distinct proof batches whose public per-shard `shard_rw_sum` values differ
on all 17 shards. The attacker takes an honest witness, replaces every
cross-shard EC accumulator leaf `(x, y)` with its inverse `(x, -y)`,
updates `shard_rw_sum`, and reproves.

Root cause: `ceno_zkvm/src/tables/shard_ram.rs:276-281` was a TODO. The
host code in `ShardRamRecord::to_ec_point` encodes read vs write in the
sign of `y[6]`, but the circuit only constrained the curve equation and
the EC sum — never tying `y[6]`'s half-of-field to `is_global_write`.
Both `(x, y)` and `(x, -y)` satisfied every existing check, so the public
summary of cross-shard RAM flow was unbound.

The defect survives recursion (the reporter's PoC verifies through the
recursion verifier program).

## Design Rationale

Approach borrows the **idea** from SP1's
`crates/core/machine/src/operations/global_interaction.rs:210-236`,
not its column layout. Three pieces:

1. **Offset by +1.** Express `y[6]` in terms of a fresh witness `y6_lo`
   so `y[6] = 0` is never valid in either branch (it is invariant under the negate operation, thus make it impossible to  distinguish read and write).
2. **Safe band + prover retry.** Restrict `y6_lo` to `[0, (p-1)/2)`. For
   the rare exception `y[6] = 0` (probability `~1/p ≈ 2^-31` per record)
   the host rejects and retries with a new `nonce`.
3. **Byte-decomposition range check.** `y6_lo` decomposed into four byte
   limbs `b0..b3` (`assert_byte` for `b0..b2`, `lookup_ltu_byte(b3, 60, 1)`
   for `b3`). For BabyBear, `(p-1)/2 = 60·2^24` exactly, so `b3 < 60`
   gives the tightest no-overlap band.

In-circuit branch equality via `condition_require_equal`:

- read  (`is_global_write = 0`): `y[6] = y6_lo + 1`  ⇒ `y[6] ∈ [1, (p-1)/2]`
- write (`is_global_write = 1`): `y[6] = p - 1 - y6_lo` ⇒ `y[6] ∈ [(p+1)/2, p-1]`

Union covers `[1, p-1]` with no overlap; `y[6] = 0` is excluded.

**Why not a single `AssertLtConfig(y6_lo, (p-1)/2, max_bits=30)`?**
On BabyBear (`p = 0x78000001`, 31-bit) the AssertLt gadget only
constrains `lhs - rhs ≡ diff - 2^max_bits (mod p)` with `diff ∈ [0, 2^30)`
— it does not pre-bound `lhs` to be canonical-small. A malicious
`y6_lo ∈ [0x74000001, p-1]` (≈ 2^26 values) produces a *field-wrap* diff
that still fits in 30 bits, so the constraint accepts upper-half values
and the exploit survives. Byte-decomposing first kills the wrap. Ceno's
`DynamicRangeTableCircuit<E, 18>` also does not carry 30-bit lookup
entries, so a direct `assert_const_range(_, 30)` is not available
anyway.

**Why M = 60 (vs SP1's 63).** SP1 targets KoalaBear; its `(p-1)/2 =
0x3f800000`, so 63 leaves a small safety band. For BabyBear,
`(p-1)/2 = 60·2^24` exactly — 63 would let `y[6]` straddle `p/2` and
reintroduce the ambiguity.

Also corrects the stale comment that previously had the convention
reversed (claimed write ⇒ lower half, opposite of what the host code
does).

### `ceno_zkvm/src/tables/shard_ram.rs` — chip-level y-sign binding

- `ShardRamRecord::to_ec_point`: reject `y6 == 0` and try the next
  `nonce`. Classify with strict `y6 > prime / 2` so the boundary
  `(p-1)/2` correctly stays in the read region (a previous draft used
  `>=` which misclassified that single boundary value and would have
  produced an out-of-range `y6_lo` for both branches).
- `ShardRamConfig`: new field `y6_lo_bytes: [WitIn; 4]`.
- `ShardRamConfig::configure`: replace the TODO with the byte
  decomposition, byte-range / LTU lookups, and the
  `condition_require_equal` branch equality.
- `ShardRamCircuit::assign_instance`: compute `y6_lo` from `y[6]` and
  `is_to_write_set` via a small `y6_lo_value` helper, assign byte
  limbs, register byte and LTU multiplicities.
- New test `test_shard_ram_y_sign_circuit_rejects_negation` drives
  `assign_instances_with_lk_multiplicities` + `MockProver` over one
  honest row and one sign-flipped row, asserting `lookup_Ltu` rejects
  the tampered witness. A concrete challenge is supplied so the
  no-challenge `run` path doesn't drop `structural_witin`.

### Lookup-multiplicity plumbing for ShardRam

ShardRam's per-row y6_lo byte / LTU lookups must reach
`combined_lk_mlt` so the U8 / LTU table `mlt` columns balance.
ShardRam runs after opcode + dummy circuits, before
`finalize_lk_multiplicities`. To surface mlt without burdening every
other table circuit:

- `ceno_zkvm/src/tables/mod.rs`: `TableCircuit` trait gains a second
  default-unimplemented method
  `assign_instances_with_lk_multiplicities` alongside the existing
  `assign_instances`. ShardRam overrides the former; every other
  table keeps overriding the latter.
- `ceno_zkvm/src/structs.rs`: `ZKVMWitnesses::assign_shared_circuit`
  threads a `LkMultiplicity::default()` through ShardRam's
  parallel-chunk witgen and inserts
  `lk_multiplicity.into_finalize_result()` into
  `lk_mlts["ShardRamCircuit"]` before finalize. Asserts swap from
  `combined_lk_mlt.is_some()` to `is_none()` to lock the ordering.
  `assign_table_circuit` tolerates `combined_lk_mlt = None` by
  passing an empty multiplicity slice, so `LocalFinalCircuit` (which
  ignores the argument anyway) can also run before finalize.
- `ceno_zkvm/src/e2e.rs`: move
  `MmuConfig::assign_continuation_circuit` (LocalFinal + ShardRam) to
  just before `finalize_lk_multiplicities`. Mirror the move inside
  the GPU debug-compare block so `combined_lk_mlt` diff stays
  meaningful.
- `ceno_zkvm/src/instructions/riscv/rv32im/mmu.rs`: docstring updated
  to describe the new ordering invariant.

### Device-resident GPU shortcut for ShardRam (mlt mirror)

`ZKVMWitnesses::try_assign_shared_circuit_gpu` dispatches into
`instructions::gpu::chips::shard_ram::try_gpu_assign_shared_circuit`
to keep the continuation EC computation device-resident
(`gpu_batch_continuation_ec_on_device` + `merge_and_partition_records`)
when `is_gpu_witgen_enabled()`. The GPU kernels never enter the CPU
`assign_instance` per-row push, so the y6_lo lookup multiplicity is
derived host-side:

- After step 6 of `try_gpu_assign_shared_circuit` (merge+partition),
  D2H `partitioned_buf` once to `Vec<u32>` and walk it with stride
  `record_u32s = 26` (`GpuShardRamRecord` `#[repr(C)]` layout).
  Per record extract `is_to_write_set` (u32 offset 10) and
  `point_y[6]` (u32 offset 25), compute `y6_lo`, push the same
  4 lookup queries the CPU path emits per row, then
  `into_finalize_result()` and return alongside the chunked
  `Vec<ChipInput<E>>`. `debug_assert_eq!(record_u32s, 26)` guards
  against `ceno_gpu` layout drift.
- `try_assign_shared_circuit_gpu` inserts both `ChipInput` and the
  derived multiplicity into `self.witnesses` /
  `self.lk_mlts["ShardRamCircuit"]` so finalize folds the GPU-path
  contribution into `combined_lk_mlt` the same way the CPU shortcut
  does.

### Verifier: account for `has_ecc_ops` row doubling

`ShardRamCircuit::has_ecc_ops()` adds an extra hypercube variable;
the chip matrix has `2 * next_pow2(num_instance)` rows where the
back half is EC-tree internal nodes with `selector_zero = 0`. Before
this fix the chip had `num_lks = 0`, so the verifier's
`dummy_table_item_multiplicity` correction never had to consider it.
With the new byte/LTU queries the correction under-counted dummy
lookups by a factor of 2 and shard verification failed with
`logup_sum != 0`.

- `ceno_zkvm/src/scheme/verifier.rs`: multiply `next_pow2_instance`
  by 2 when `circuit_vk.get_cs().has_ecc_ops()`.
- `ceno_recursion/src/zkvm_verifier/verifier.rs`: mirror the same
  adjustment in the recursive verifier (lockstep per CLAUDE.md).

Per ShardRam row this PR adds **4 byte WitIn columns** plus 3 byte-range
and 1 LTU lookup multiplicities. ShardRam rows scale with cross-shard
RAM events, not with cycles, so the absolute cost is sub-percent on the
prover. No full prover bench was rerun (no hot-loop arithmetic changed).

Existing `test_shard_ram_circuit` (170k reads + 1420 writes, full chip
proof) runtime is unchanged within noise:

```text
master   : ~5.0 s
this PR  : ~5.0 s
```

## Testing

```sh
cargo fmt --all --check
cargo check --workspace --all-targets --release
cargo make clippy
cargo clippy --workspace --all-targets --release -- -D warnings
RUST_MIN_STACK=33554432 cargo test --workspace --lib --release
cargo run --release --package ceno_zkvm --features sanity-check --bin e2e -- \
  --platform=ceno --max-cycle-per-shard=20000 --hints=10 --public-io=4191 \
  examples/target/riscv32im-ceno-zkvm-elf/release/examples/fibonacci
```

All pass locally on BabyBear. `test_shard_ram_circuit` and
`test_shard_ram_y_sign_circuit_rejects_negation` are green. End-to-end
multi-shard fibonacci verifies `ShardRamCircuit` and `LocalRAMTableFinal`
on every shard with `exit code 0. Success.`

`cargo make tests` / `cargo make tests_goldilock` should be re-run by
CI; the change is gated to BabyBear via a `debug_assert_eq!` on
`MODULUS_U64` and goldilocks does not exercise shard_ram (per
`integration.yml` commented-out lines and CLAUDE.md).

## Risks and Rollout

- **Soundness.** Closes #1338. The new constraint only adds local byte
  arithmetic and existing lookups — no change to transcript, sumcheck,
  PCS, or EC accumulation. Recursive and native verifiers move in
  lockstep (the `has_ecc_ops` row-factor fix lands in both).
- **GPU.** The device-resident GPU shortcut now derives the y6_lo
  lookup multiplicity host-side from the merged partitioned device
  buffer (single D2H of ~26 u32 × records). Layout assumption is
  guarded by `debug_assert_eq!(record_u32s, 26)` against
  `ceno_gpu::GpuShardRamRecord`. CPU + GPU paths converge on the same
  `combined_lk_mlt` contribution; runtime verification with
  `CENO_GPU_ENABLE_WITGEN=1 --features gpu` on a CUDA host is
  recommended before tag.
- **Recursion.** The recursive verifier mirrors the native verifier's
  `has_ecc_ops` × 2 row adjustment; no separate constraint-system
  change is needed for the y-sign binding itself.
- **Field support.** Hardcodes the BabyBear constant `M = 60`. A
  `debug_assert_eq!(MODULUS_U64, 0x78000001, ...)` guards against
  accidental use on a different field; shard_ram is BabyBear-only
  today per CLAUDE.md.

## Follow-ups

- The remaining #1340 TODOs (`local read ⇄ global write` pairing on
  `shard_ram.rs:235-236`, `shard == shard_id` binding on line 244) are
  intentionally out of scope here.

Fixes #1338.
Partially addresses #1340.

### DIFF
--- a/ceno_zkvm/src/tables/shard_ram.rs
+++ b/ceno_zkvm/src/tables/shard_ram.rs
@@ -132,9 +132,8 @@ impl ShardRamRecord {
                 hasher.permute(input.clone())[0..SEPTIC_EXTENSION_DEGREE].into();
             if let Some(p) = SepticPoint::from_x(x) {
                 let y6 = (p.y.0)[SEPTIC_EXTENSION_DEGREE - 1].to_canonical_u64();
-                // Skip the 2-torsion case where (x, y) == (x, -y); the y-sign
-                // binding in the circuit cannot distinguish read from write
-                // when y6 = 0.
+                // Reject cases where y6 = 0 because then the y-sign
+                // binding in the circuit cannot distinguish read from write.
                 if y6 != 0 {
                     // Strict `>`: `prime / 2 = (p-1)/2` belongs to the lower
                     // half (read region `[1, (p-1)/2]`). Using `>=` would
--- a/ceno_zkvm/src/tables/shard_ram.rs
+++ b/ceno_zkvm/src/tables/shard_ram.rs
@@ -41,6 +41,8 @@ use witness::{InstancePaddingStrategy, next_pow2_instance_padding, set_val};
 
 use crate::{instructions::riscv::constants::UInt, scheme::constants::SEPTIC_EXTENSION_DEGREE};
 
+pub(crate) const Y6_LO_TOP_BYTE_LT_BOUND: u64 = 60;
+
 /// A record for a read/write into the shard RAM
 #[derive(Debug, Clone)]
 pub struct ShardRamRecord {
@@ -305,7 +307,7 @@ impl<E: ExtensionField> ShardRamConfig<E> {
         // `lookup_ltu_byte(a, b, 1)` asserts `a, b` are bytes and `a < b`.
         cb.lookup_ltu_byte(
             y6_lo_bytes[3].expr(),
-            E::BaseField::from_canonical_u64(60).expr(),
+            E::BaseField::from_canonical_u64(Y6_LO_TOP_BYTE_LT_BOUND).expr(),
             Expression::ONE,
         )?;
         let y6_lo = y6_lo_bytes[0].expr()
@@ -423,7 +425,7 @@ impl<E: ExtensionField> ShardRamCircuit<E> {
             let b = (y6_lo_u64 >> (8 * i)) & 0xff;
             lk_multiplicity.assert_const_range(b, 8);
         }
-        lk_multiplicity.lookup_ltu_byte((y6_lo_u64 >> 24) & 0xff, 60);
+        lk_multiplicity.lookup_ltu_byte((y6_lo_u64 >> 24) & 0xff, Y6_LO_TOP_BYTE_LT_BOUND);
 
         let ram_type = E::BaseField::from_canonical_u32(record.ram_type as u32);
         let mut input = [E::BaseField::ZERO; 16];
--- a/ceno_zkvm/src/tables/shard_ram.rs
+++ b/ceno_zkvm/src/tables/shard_ram.rs
@@ -1,4 +1,3 @@
-use rustc_hash::FxHashMap;
 use std::{iter::repeat_n, marker::PhantomData};
 
 use crate::{
@@ -133,24 +132,32 @@ impl ShardRamRecord {
                 hasher.permute(input.clone())[0..SEPTIC_EXTENSION_DEGREE].into();
             if let Some(p) = SepticPoint::from_x(x) {
                 let y6 = (p.y.0)[SEPTIC_EXTENSION_DEGREE - 1].to_canonical_u64();
-                let is_y_in_2nd_half = y6 >= (prime / 2);
+                // Skip the 2-torsion case where (x, y) == (x, -y); the y-sign
+                // binding in the circuit cannot distinguish read from write
+                // when y6 = 0.
+                if y6 != 0 {
+                    // Strict `>`: `prime / 2 = (p-1)/2` belongs to the lower
+                    // half (read region `[1, (p-1)/2]`). Using `>=` would
+                    // misclassify it and produce `y6_lo = (p-1)/2` whose top
+                    // byte b3 = 60 fails `lookup_ltu_byte(b3, 60, 1)`.
+                    let is_y_in_2nd_half = y6 > (prime / 2);
 
-                // we negate y if needed
-                // to ensure read => y in [0, p/2) and write => y in [p/2, p)
-                let negate = match (self.is_to_write_set, is_y_in_2nd_half) {
-                    (true, false) => true, // write, y in [0, p/2)
-                    (false, true) => true, // read, y in [p/2, p)
-                    _ => false,
-                };
+                    // Enforce convention:
+                    //   is_to_write_set = 0 (read)  => y6 in [1, (p-1)/2]
+                    //   is_to_write_set = 1 (write) => y6 in [(p+1)/2, p-1]
+                    let negate = matches!(
+                        (self.is_to_write_set, is_y_in_2nd_half),
+                        (true, false) | (false, true)
+                    );
 
-                let point = if negate { -p } else { p };
+                    let point = if negate { -p } else { p };
 
-                return ECPoint { nonce, point };
-            } else {
-                // try again with different nonce
-                nonce += 1;
-                input[6] = E::BaseField::from_canonical_u32(nonce);
+                    return ECPoint { nonce, point };
+                }
             }
+            // try again with different nonce
+            nonce += 1;
+            input[6] = E::BaseField::from_canonical_u32(nonce);
         }
     }
 }
@@ -180,6 +187,9 @@ pub struct ShardRamConfig<E: ExtensionField> {
     pub(crate) x: Vec<WitIn>,
     pub(crate) y: Vec<WitIn>,
     pub(crate) slope: Vec<WitIn>,
+    // Byte limbs of `y6_lo`, the helper that binds `y[SEPTIC_EXTENSION_DEGREE - 1]`
+    // to `is_global_write` in `configure`.
+    pub(crate) y6_lo_bytes: [WitIn; 4],
     pub(crate) perm_config: Poseidon2Config<E, 16, 7, 1, 4, 13>,
 }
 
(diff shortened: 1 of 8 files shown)


ceno_zkvm/src/tables/shard_ram.rs
```diff
@@ -133,24 +133,28 @@ impl ShardRamRecord {
...
-
-                return ECPoint { nonce, point };
-            } else {
-                // try again with different nonce
-                nonce += 1;
-                input[6] = E::BaseField::from_canonical_u32(nonce);
+                // Skip the 2-torsion case where (x, y) == (x, -y); the y-sign
+                // binding in the circuit cannot distinguish read from write
+                // when y6 = 0.
+                if y6 != 0 {
```
kunxian-xia, line 139: why do we need this condition?

kunxian-xia: i think we do need to handle the 2-torsion case where $y = \vec{\mathbf{0}}$ in which we **cannot** distinguish read from write.

spherel: If y6 == 0, we will re-generate the point.

kunxian-xia: in fact, the curve that we use does not have a point with order 2. All point's order is a prime. Please refer to [claude](https://claude.ai/chat/4cfaa431-d110-43dd-934b-c5e7e6b3b520) for a self-contained explanation.

ceno_zkvm/src/tables/shard_ram.rs
```diff
@@ -273,12 +282,45 @@ impl<E: ExtensionField> ShardRamConfig<E> {
...
+        );
+        let y6_lo_bytes: [WitIn; 4] =
+            std::array::from_fn(|i| cb.create_witin(|| format!("y6_lo_b{i}")));
+        for (i, w) in y6_lo_bytes.iter().enumerate().take(3) {
+            cb.assert_byte(|| format!("y6_lo_b{i} byte"), w.expr())?;
+        }
+        // `lookup_ltu_byte(a, b, 1)` asserts `a, b` are bytes and `a < b`.
+        cb.lookup_ltu_byte(
+            y6_lo_bytes[3].expr(),
+            E::BaseField::from_canonical_u64(60).expr(),
```
kunxian-xia: @copilot please make this `60` as a constant s.t. we can refers to it here and in witness generation.

kunxian-xia (approved):

PR #1360, opened 2026-06-16
fix(soundness): bound untrusted num_instances in native + recursion verifiers

### LINKED ISSUE #1178: structural witness max length check under field arithmetics
1. respective check in rust verifier not found in recursion verifier 
https://github.com/scroll-tech/ceno/blob/ff341ef5960ee2031e90d4d510ab013c83bcc350/ceno_zkvm/src/scheme/verifier.rs#L484-L507

@darth-cy am I correct?

2. This check seems invalid under prime(ext) field because of wrapping logic

spherel:
## Summary

`num_instances` (per-shard, per-circuit instance counts) is read straight from
the untrusted proof and then flows into size/shift/field arithmetic in **both**
verifiers. Neither side bounded it, so a crafted proof could push these
quantities out of range — a verifier-side DoS and a field-order decoupling, i.e.
a soundness concern. This PR adds the matching upper bound on both the native
Rust verifier and the recursion verifier so they stay in lockstep.

## Problem

In the native verifier, `proof.num_instances` feeds:
- `proof.num_instances.iter().sum()` — can wrap (or panic in debug);
- `next_pow2_instance_padding` and the `1 << log2_num_instances` `usize` shifts —
  can overflow / panic on attacker input (DoS);
- the `from_canonical_usize` transcript casts — can exceed the base-field modulus
  and wrap, decoupling the transcript-bound value from the real count.

The recursion verifier (`verify_chip_proof_pre_main`) is worse: it does not even
derive `log2_num_instances`/`sum_num_instances` — it reads them as **untrusted
hints**. An out-of-range `log2_num_instances` drives `pow_2(log2_num_instances)`
past the BabyBear field size (~2³¹), wrapping the `next_pow2`/`next_pow2 - sum`
offset used by the eq selectors, which would let the recursive proof verify a
different statement than the native verifier.

## Fix

A valid proof's instance count is always bounded by the PCS setup size
(`2^MAX_NUM_VARIABLES`, with the witness committed at `≤ 2^MAX_NUM_VARIABLES`), so
rejecting anything larger only drops malformed proofs earlier and does not change
the statement verified for any valid proof.

- **Native** (`ceno_zkvm/src/scheme/verifier.rs`): reject any per-shard
  `num_instances` entry (and their sum) `> 2^MAX_NUM_VARIABLES`, both where the
  count is first summed and (authoritatively) in `verify_chip_proof_pre_main`,
  right before the shift/cast it protects.
- **Recursion** (`ceno_recursion/src/zkvm_verifier/verifier.rs`): in
  `verify_chip_proof_pre_main`, before the hints feed `pow_2(log2_num_instances)`
  and the offset, assert `log2_num_instances ≤ MAX_NUM_VARIABLES` and
  `sum_num_instances` / each `num_instances[i] ≤ 2^MAX_NUM_VARIABLES`. The bound
  lives in the verifier function (not the hint `read()`, which is not part of the
  verified circuit).

`2^MAX_NUM_VARIABLES = 2^24` is far below the BabyBear modulus, so the field-order
wrap is closed on both sides.

## Test plan

- [x] `cargo check` / `cargo clippy` clean for both `ceno_zkvm` and `ceno_recursion`
- [x] Native: integration e2e subset (fibonacci + ceno_rt_alloc + keccak_syscall,
      debug MOCK + release prove/verify, single + multi-shard) — 9/9 pass; every
      release run exercises the new bound in `verify_chip_proof_pre_main`
- [x] Recursion: `e2e_aggregate` over multi-shard `keccak_syscall` verifies —
      both leaf proofs run `verify_chip_proof_pre_main` (and thus the new asserts)
      for every chip across 2 shards, then internal aggregation completes (final
      height 1), confirming valid proofs are not rejected

## Follow-up (out of scope)

This closes the **bound** (DoS / field-wrap). Fully *binding* the recursion's
`log2_num_instances` / bit-decomposition hints to be exactly consistent with
`num_instances` (rather than only bounding them) — which the native side gets for
free by deriving `log2` — remains a separate hardening item.

Closes #1178, #1240.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hero78119:
Commit 5c797531 and 7f482b97 covers:

- Merged master and kept the old ceno_recursion deletion.
- Added shared num_instances bounds: MAX_NUM_VARIABLES = 29, per-entry num_instances < 2^28.
- Native verifier now rejects oversized proof-controlled num_instances entries before summing.
- recursion-v2 ProofShapeAir enforces the same per-entry bound via the existing u8 range table.
- Added recursion proof-shape coverage for accepting 2^28 - 1 and rejecting 2^28.

Validated with fmt, ceno_recursion_v2 proof_shape tests, ceno_zkvm scheme tests, and git diff --check.

hero78119:
Update for `7f482b9..372fec9`:

- Keeps PCS capacity at `MAX_NUM_VARIABLES = 24`, but changes raw `num_instances` entries to the exclusive bound `< 2^29`.
- Removes the recursion proof-shape cap on `height_1 + height_2`; only `height_1` and `height_2` are range-checked as raw entries.
- Adds Jagged reshape-height capacity checks: concrete GPU Jagged paths assert setup/prover metadata fits `<= MAX_NUM_VARIABLES`, and recursion `VmPvsAir` constrains the proof-supplied witness commitment reshape height.
- Fixed VK commitments in recursion are asserted during trace generation; only witness commitment metadata is AIR-constrained.

Verified with `cargo fmt --all --check`, `cargo check -p ceno_zkvm --bin e2e`, `cargo test -p ceno_zkvm scheme`, `cargo test -p ceno_recursion_v2 proof_shape`, `cargo check -p ceno_recursion_v2 --lib`, and the integration workflow aggregation command for `e2e_aggregate`.

